PR #7, opened 2026-06-30
Added Tetris to Games

Yesanith:
Added Tetris to the Games section. It doesn't have an icon like the other games yet, and I'm currently working on implementing the correct scoring and level progression systems. I may also add keyboard support to make the game smoother to play, but I haven't figured out how to implement that yet.


src/Aetherphone/Apps/Games/Tetris/TetrisApp.cs
```diff
@@ -0,0 +1,207 @@
...
+
+            pendingSubmit = false;
+        }
+
+        if (!board.GameOver)
+        {
+            board.Update(deltaSeconds);
+        }
+
+        if (board.ClearedLinesThisFrame > 0)
```
XeldarAlz: Hard-drop line clears never trigger the shake/flash feedback.

This `ClearedLinesThisFrame > 0` check runs near the top of `Draw`, but the `D`/HardDrop button is handled about 70 lines below, so `HardDrop()` sets `ClearedLinesThisFrame` after this check has already passed for the frame. On the next frame, `board.Update()` resets it back to `0` (the first line of `TetrisBoard.Update`) before this check runs again.

Since gravity is slow (0.72s at level 1) and there is no soft-drop button, hard drop is the main way the player clears lines, so `fx.AddTrauma`/`fx.Flash` here ends up unreachable for the common case, including the 4-line tetris. Only gravity-locked clears, which set the field inside `Update` within the same frame, currently get the effect.

Suggested fix: handle the control buttons before this check, or read `ClearedLinesThisFrame` after the input is processed.

src/Aetherphone/Apps/Games/GamesApp.cs
```diff
@@ -61,6 +62,7 @@ public GamesApp(GameStatsStore stats)
             new SweeperApp(),
             new PairsApp(),
             new GemSwapApp(),
+            new TetrisApp(),
```
XeldarAlz, line 65: `TetrisApp` tracks and persists a best score via `Stats.SubmitScore("tetris", ...)`, but `StatValue` (further down, around line 236) has no `"tetris"` case, so it returns `string.Empty` and the launcher card's best-score chip is always suppressed. Every other score-based game (`2048`, `breakout`, `snake`, and so on) shows it.

Adding a `case "tetris":` that returns `GameNumber.Label(stats.Get("tetris").BestScore)` would let the tracked best appear on the card, consistent with the others.

src/Aetherphone/Apps/Games/Tetris/TetrisRenderer.cs
```diff
@@ -0,0 +1,173 @@
...
+        {
+            var y = grid.Origin.Y + row * grid.Pitch;
+            drawList.AddLine(new Vector2(grid.Origin.X, y), new Vector2(grid.Origin.X + grid.Width, y), lineColor, 1f * scale);
+        }
+    }
+
+    private static void DrawActivePiece(ImDrawListPtr drawList, TetrisBoard board, GameGrid grid, int boardY, float alpha, float scale)
+    {
+        var tint = GamePalette.Lighten(PieceColorOf((int)board.ActiveKind), 0.12f);
+        var cells = new (int X, int Y)[]
```
XeldarAlz: Small allocation in the draw loop: this builds a 4-element tuple array, and `DrawActivePiece` is called twice per frame (ghost and active), so it allocates twice every frame at 60 fps.

The intermediate array isn't required. Looping `index` from 0 to 3 and passing `board.GetCell(index)` straight into the `DrawCell` call produces the same output without the per-frame allocation.

src/Aetherphone/Apps/Games/Tetris/TetrisBoard.cs
```diff
@@ -0,0 +1,435 @@
...
+
+    public bool Rotate(int direction)
+    {
+        if (GameOver || !HasActivePiece)
+        {
+            return false;
+        }
+
+        var nextRotation = (activeRotation + (direction >= 0 ? 1 : 3)) & 3;
+        var kicks = new (int X, int Y)[] { (0, 0), (-1, 0), (1, 0), (-2, 0), (2, 0), (0, -1) };
```
XeldarAlz: The wall-kick table is rebuilt on every rotation. Since it is a fixed table, promoting it to a `private static readonly (int X, int Y)[]` field (like the `Shapes` table above) avoids the per-rotation allocation. Lower frequency than the renderer case since it is input-driven rather than per-frame, but a quick win.

XeldarAlz (commented):
Thanks for adding Tetris. The code is clean and follows the IMiniGame pattern nicely (7-bag randomizer, hold/swap, ghost piece, back-to-back and combo scoring). All referenced framework APIs and localization keys resolve, and the locale entries are complete across de, en, fr, and tr. I left a few inline notes below. The hard-drop feedback ordering in TetrisApp is the one worth addressing before merge; the others are a small completeness gap and two allocation cleanups. Nice work overall.

Yesanith:
Addressed the review feedback. Also implemented a detailed scoring system based on the classic Tetris scoring rules, and added a preview of the next piece for the player.

PR #35, opened 2026-07-21
Fix audio playback failing with BadDeviceId on virtual audio drivers (SteelSeries Sonar)

YozoraCho:
## What

Routes all NAudio playback (voice notes, calls, songs, radio, custom sound files) through a new `AudioOutputFactory` that uses WASAPI shared mode (`WasapiOut`) instead of legacy winmm (`WaveOutEvent`), with a waveOut fallback if WASAPI init fails.

## Why

With virtual audio drivers such as SteelSeries Sonar as the default output device, the game process can see zero winmm waveOut devices, so every `WaveOutEvent` fails with `BadDeviceId calling waveOutOpen` — voice notes, calls, songs, and radio produce no audio at all. WASAPI still sees the endpoints (mic capture and the game-engine ringtones already worked), so switching the output API fixes playback. No new dependencies: `NAudio.Wasapi` was already referenced.

Example log output before the fix:

```
WRN | [Aetherphone] Voice note playback failed: BadDeviceId calling waveOutOpen
WRN | [Aetherphone] Voice output failed to start: BadDeviceId calling waveOutOpen
WRN | [Aetherphone] Song playback failed: BadDeviceId calling waveOutOpen
```

## How to test

1. Set a SteelSeries Sonar virtual device (or any output that reproduces the waveOut issue) as the Windows default output. On a normal device, playback should behave exactly as before.
2. `/phone` → Messenger → play a voice note.
3. Make a call and confirm you hear the other party.
4. Play a song and the radio.
5. Before this change, all of these log `BadDeviceId calling waveOutOpen` in `/xllog`; after, they play with no warnings.

## Checklist

- [ ] If this changes user-visible behavior, README is updated (no user-facing changes beyond audio now working)

### DIFF
--- a/src/Aetherphone/Core/Telephony/Audio/VoiceMixer.cs
+++ b/src/Aetherphone/Core/Telephony/Audio/VoiceMixer.cs
@@ -30,7 +30,7 @@ internal sealed class VoiceMixer : ISampleProvider, IDisposable
     private readonly short[] decodeBuffer = new short[OpusAudio.FrameSamples * 6];
     private readonly byte[] pcmBytes = new byte[OpusAudio.FrameSamples * 6 * sizeof(short)];
     private float[] mixScratch = Array.Empty<float>();
-    private WaveOutEvent? output;
+    private IWavePlayer? output;
     private VolumeSampleProvider? volumeProvider;
     private float volume = 0.85f;
     public WaveFormat WaveFormat => format;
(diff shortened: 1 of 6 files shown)


src/Aetherphone/Core/Telephony/Audio/VoiceMixer.cs
```diff
@@ -55,8 +55,9 @@ public void Start(int deviceNumber, float startVolume)
         volume = Math.Clamp(startVolume, 0f, 1f);
         try
         {
+            _ = deviceNumber; // WASAPI plays on the system default output; winmm device indices no longer apply.
```
YozoraCho, line 58: The winmm output-device plumbing could be removed as a follow-up. Was left for comment more than the discard.

XeldarAlz (approved):
Thanks for tracking this down and fixing it so cleanly, @YozoraCho! Nailed the root cause: virtual drivers (Sonar/VoiceMeeter) expose WASAPI endpoints but zero winmm waveOut devices, so `WaveOutEvent` dies with `BadDeviceId`. The fix is well scoped:

- All 5 playback paths routed through `AudioOutputFactory` (voice notes, sound effects, radio, songs, call mixer); capture (`WaveInEvent`) correctly untouched.
- The `_ = deviceNumber;` in `VoiceMixer` is a no-op in practice since `ResolveOutput` already returns -1 (default device), so no behavior change there.
- Gain stays in-graph via `VolumeSampleProvider`, so it doesn't touch device volume (avoids the FFXIV mixer-ducking issue).
- Nice safety net with the waveOut fallback + `waveOutGetNumDevs()` in the log.

LGTM. 🙏

PR #38, opened 2026-07-22
fix(device): use synchronous Ping.Send to avoid native fault under Wine

Kiro-XVI:
## Problem

`DeviceStatus.RunAsync` polls the signal-bar readout via `Ping.SendPingAsync`. Under Wine the async ICMP path faults at the native level rather than throwing, so the `try/catch` in `SampleNetworkAsync` never sees it and the game process dies.

Two crash signatures on the same machine within five minutes:

- native access violation (`C0000005`) in `ffxiv_dx11.exe` on the FrameworkUpdate thread
- CLR `ExecutionEngineException` (`0x80131506`) on a `.NET TP Worker` inside a task continuation — the shape `SendPingAsync` produces

Neither left an entry in `dalamud.log`, consistent with a native fault rather than a swallowed managed exception.

## Ruled out

- **Kernel ICMP permissions** — `net.ipv4.ping_group_range = 0 2147483647`, fully permissive
- **ICMP under Wine generally** — `wine ping 8.8.8.8` against a fresh prefix on the *same* wine-xiv 10.8 build: 4/4 replies, ~48ms, clean exit

Synchronous ICMP is healthy. The fault is specific to the async completion path.

## Causation

An ablation branch removing the probe entirely ran clean, confirming `DeviceStatus` as the source. Verified Dalamud was loading the local build rather than the installed plugin.

## Change

One line. `Task.Run` keeps the loop async without blocking a framework thread, while routing through the synchronous entry point.

## Testing

CachyOS, XIVLauncher.Core 1.4.0, wine-xiv-staging-fsync-git 10.8, Dalamud 15.0.2.3, .NET 10, RX 9070 XT (RADV).

Ran 01:17:19  with no crash. Signal bars, latency and packet loss all reporting normally. Both original crashes occurred within three minutes of login.

## Notes

This is the minimal fix. The more robust approach is the IpHlpApi route PingPlugin uses — `GetExtendedTcpTable` filtered to the game PID plus `GetRTTAndHopCount` — which would also retire the hardcoded data-center host list in `SyncTarget`/`TryDataCenterHost`. Happy to do that as a follow-up.

Happy to keep testing on Linux since you can't reproduce locally.

### DIFF
--- a/src/Aetherphone/Core/Device/DeviceStatus.cs
+++ b/src/Aetherphone/Core/Device/DeviceStatus.cs
@@ -86,7 +86,7 @@ internal sealed class DeviceStatus : IDisposable
             try
             {
                 SampleBattery();
-                await SampleNetworkAsync(ping).ConfigureAwait(false);
+                SampleNetwork(ping);
                 await Task.Delay(SampleIntervalMilliseconds, token).ConfigureAwait(false);
             }
             catch (OperationCanceledException)
@@ -124,14 +124,14 @@ internal sealed class DeviceStatus : IDisposable
         charging = false;
     }
 
-    private async Task SampleNetworkAsync(Ping ping)
+    private void SampleNetwork(Ping ping)
     {
         var endpoint = target;
         var succeeded = false;
         long roundtrip = 0;
         try
         {
-            var reply = await Task.Run(() => ping.Send(endpoint, PingTimeoutMilliseconds)).ConfigureAwait(false);
+            var reply = ping.Send(endpoint, PingTimeoutMilliseconds);
             if (reply.Status == IPStatus.Success)
             {
                 succeeded = true;
--- a/src/Aetherphone/Core/Device/DeviceStatus.cs
+++ b/src/Aetherphone/Core/Device/DeviceStatus.cs
@@ -131,7 +131,7 @@ internal sealed class DeviceStatus : IDisposable
         long roundtrip = 0;
         try
         {
-            var reply = await ping.SendPingAsync(endpoint, PingTimeoutMilliseconds).ConfigureAwait(false);
+            var reply = await Task.Run(() => ping.Send(endpoint, PingTimeoutMilliseconds)).ConfigureAwait(false);
             if (reply.Status == IPStatus.Success)
             {
                 succeeded = true;


src/Aetherphone/Core/Device/DeviceStatus.cs
```diff
@@ -131,7 +131,7 @@ private async Task SampleNetworkAsync(Ping ping)
         long roundtrip = 0;
         try
         {
-            var reply = await ping.SendPingAsync(endpoint, PingTimeoutMilliseconds).ConfigureAwait(false);
+            var reply = await Task.Run(() => ping.Send(endpoint, PingTimeoutMilliseconds)).ConfigureAwait(false);
```
XeldarAlz: The `Task.Run` wrapper is redundant here. `RunAsync` is already dispatched to the thread pool at line 52, so this loop never touches a framework thread. Net pool pressure is identical either way: with `Task.Run` the outer thread is released while an inner pool thread blocks for up to 1.5s; without it, the same one thread blocks. You just pay an extra closure + delegate + `Task` per sample for it.

Please drop the wrapper and make the method synchronous:

```csharp
private void SampleNetwork(Ping ping)
{
    var endpoint = target;
    var succeeded = false;
    long roundtrip = 0;
    try
    {
        var reply = ping.Send(endpoint, PingTimeoutMilliseconds);
```

and call `SampleNetwork(ping);` in `RunAsync`. That also removes the async state machine, and the `await Task.Delay` structure stays as-is.

XeldarAlz: Minor, and it goes away with the change above: the cancellation token isn't passed to `Task.Run`. `Send` can't be cancelled mid-flight anyway, so on `Dispose` a pool thread stays parked for up to 1.5s regardless.

XeldarAlz (commented):
Diagnosis looks right, and the evidence (ablation branch, clean `wine ping` baseline, permissive `ping_group_range`, no managed exception in `dalamud.log`) points squarely at the async completion path. Happy to take this.

One change requested on the implementation, plus a note.

On the follow-up: yes please. `GetExtendedTcpTable` filtered to the game PID + `GetRTTAndHopCount` is the right direction, and retiring the hardcoded data-center list in `TryDataCenterHost` is worth it on its own. Open it as a separate PR.

And yes, please keep testing on Linux, I can't reproduce any of this locally.

PR #39, opened 2026-07-22
Favorite Radio Station feature

Hubkaw:
## What
A simple Favorite radio feature that adds a small star next to the radio station that pins them to the home screen.

image

## Why

I was annoyed to always search for my favorite Gensokyo Radio so had to do it.  
  
This is my first ever pr to someone else's repo please be patient with me x)  
  
## How to test

1) open the phone
2) navigate to Music App
3) click one of the radio categories to display station list
4) hover over any to see the star appear
5) press the star 
6) navigate to Home screen of Music app
7) Favorite Stations section should appear and you should be able to play the stations there.
8) Navigate to NowPlaying screen of a radio
9) check the star in top right corner - should work the same way as the one on the list

## Checklist

- [x] If this changes user-visible behavior, README is updated (I dont think README.md updates are needed here?)
- [x] Matches existing style: uses the shared `Windows/Components/` widgets, no `what` comments (I think so?)


src/Aetherphone/Core/Localization/L.cs
```diff
@@ -1253,6 +1253,7 @@ internal static class Music
         public static readonly LocString RadioSearchSub = new("music.radioSearchSub", "Search by name, genre, or country");
         public static readonly LocString YourPlaylists = new("music.yourPlaylists", "Your playlists");
         public static readonly LocString AddToPlaylist = new("music.addToPlaylist", "Add to playlist");
+        public static readonly LocString FavoriteStations = new("music.favoriteStation", "Favorite stations");
```
XeldarAlz: Key mismatch: `music.favoriteStation` here, `music.favoriteStations` in `en.json`.

`Loc.T` looks up by `entry.Key` (`Loc.cs:34`), so the JSON entry is dead and this always falls back to the inline English. It works today, but the string can never be translated. Pick one spelling for both.

src/Aetherphone/Localization/en.json
```diff
@@ -996,6 +996,7 @@
   "music.radioSearchSub": "Search by name, genre, or country",
   "music.yourPlaylists": "Your playlists",
   "music.addToPlaylist": "Add to playlist",
+  "music.favoriteStations": "Favorite stations",
```
XeldarAlz: Two things:

1. Key doesn't match `L.cs` (see comment there), so this entry is currently unreachable.
2. The repo ships nine locales; only `en.json` was updated, so the other eight silently fall back to English.

If you'd rather not write eight translations, say so and I'll fill them in on merge.

src/Aetherphone/Apps/Music/MusicApp.Home.cs
```diff
@@ -431,20 +432,52 @@ private void DrawStationRow(float scale, RadioStation station, int index)
                 clock, ui.Accent, 1f, playback.IsPlaying);
         }
 
+        var favoriteStar = false;
+        var isFavoriteStation = IsFavoriteStation(station.StreamUrl);
+
+        if (isFavoriteStation || hovered)
+        {
+            var starX = current ? max.X - 36f * scale : max.X - 18f * scale;
```
XeldarAlz, line 442: The star overlaps the station name.

`trailing` (line 421) is still `current ? 32f : 10f`, so text runs to `max.X - 10f`. The star covers roughly `max.X - 32` to `max.X - 4`, so long station names render underneath it. Same on current rows, where the equalizer and star together need more than 32f.

Widen `trailing` to reserve room, and do it unconditionally: if the width changed on hover, `FitText` would re-truncate and the label would jump.

Nit: the `Plus` button in `MusicApp.NowPlaying.cs` passes a tooltip; both new stars are unlabelled.

src/Aetherphone/Apps/Music/MusicApp.Home.cs
```diff
@@ -431,20 +432,52 @@ private void DrawStationRow(float scale, RadioStation station, int index)
...
+
         ImGui.SetCursorScreenPos(origin);
         ImGui.Dummy(new Vector2(width, rowHeight));
         if (!hovered || !ImGui.IsMouseClicked(ImGuiMouseButton.Left))
         {
             return;
         }
 
-        if (current)
+        if (current && !favoriteStar)
```
XeldarAlz: The guard is repeated. It folds into the early return above:

```csharp
if (favoriteStar || !hovered || !ImGui.IsMouseClicked(ImGuiMouseButton.Left))
{
    return;
}
```

Then the original `if (current) / else` body stays as it was.

src/Aetherphone/Apps/Music/MusicApp.cs
```diff
@@ -648,6 +658,34 @@ private bool IsCurrentStation(RadioStation station)
         return playback.RadioActive && playback.Radio.CurrentStation == station.Name;
     }
 
+    private bool IsFavoriteStation(string streamUrl)
+    {
+        return !string.IsNullOrEmpty(streamUrl) &&
+               configuration.RadioFavorites.Exists(favorite => favorite.StreamUrl == streamUrl);
```
XeldarAlz: This lambda captures `streamUrl`, so it allocates a closure and a delegate per call, once per station row per frame.

`VenuesApp.IsFavorite` is a plain `List.Contains` with no closure. Same fix as `GetFavoriteRadioStations` below.

src/Aetherphone/Apps/Music/MusicApp.cs
```diff
@@ -648,6 +658,34 @@ private bool IsCurrentStation(RadioStation station)
...
+        {
+            configuration.RadioFavorites.Add(RadioStationRecord.From(station));
+        }
+
+        configuration.Save();
+    }
+
+    private RadioStation[] GetFavoriteRadioStations()
+    {
+        return configuration.RadioFavorites.Select(c => c.ToStation()).ToArray();
```
XeldarAlz: Main thing I'd like changed: this allocates an array through LINQ, and `DrawFavoriteRadioStationsSection` calls it every frame.

The draw path is performance-critical here. `MergeStations` (~line 500, same file) does equivalent work with explicit `for` loops and a `HashSet`, and `MapsApp` keeps favourites in a field synced on toggle rather than deriving them on read.

Suggested: keep a `RadioStation[] favoriteStations` field, rebuild it only in `ToggleFavoriteStation` and once at construction, and read it from both the draw code and `IsFavoriteStation`. That removes the per-frame array, the LINQ, and the closure above. It also removes the second `GetFavoriteRadioStations()` call in `PlayStationFromFavorites`, which currently indexes a freshly derived array with an index from a different instance.

Nit: `c` here and `f` in `ToggleFavoriteStation` are abbreviations (you got this right with `index` in the loops), and there's a double blank line below.

src/Aetherphone/Core/Radio/RadioPlayer.cs
```diff
@@ -39,6 +40,7 @@ public RadioPlayer()
 
     public RadioPlaybackState State => state;
     public string CurrentStation => currentStation;
+    public RadioStation CurrentStationInfo => currentStationInfo;
```
XeldarAlz: Subtle one. `state` and `currentStation` are `volatile` on purpose: written from the streaming worker, read from the draw thread, single word so the write is atomic.

`currentStationInfo` is a six-field struct, and it's written under `lock (gate)` in `StartStation`, outside any lock in `Stop()`, and read here with no lock. A torn read could mix `Name` from one station with `StreamUrl` from another, so the star could evaluate against the wrong URL or persist a stitched-together record.

Fix consistent with the file: return it under `lock (gate)` and move the `Stop()` assignment inside the lock.

XeldarAlz (commented):
Thanks for this, and welcome! Good first PR: well scoped, and you clearly read the surrounding code first. The config + `Save()` toggle, the `RadioStationRecord` shape, reusing `AppSkin.IconButton`, resolving `LocString` at draw time, and storing the full station instead of just the uuid are all exactly right.

I left inline comments. Before merge:

**Bugs**
- Loc key mismatch between `L.cs` and `en.json`, so the JSON entry is unreachable.
- The star overlaps long station names: `trailing` was not widened to reserve room for it.
- `RadioPlayer.CurrentStationInfo` is a multi-field struct read lock-free from the draw thread.

**Conventions**
- LINQ and per-frame allocations in the draw path (`GetFavoriteRadioStations`, `IsFavoriteStation`). Caching a field fixes both.
- Abbreviated lambda params (`f`, `c`), two double blank lines.

**Polish**
- Only `en.json` was updated; the repo ships nine locales. Easy to miss, since English renders fine either way.

Not a rejection, just one round of cleanup. Happy to help with any of it.

src/Aetherphone/Core/Localization/L.cs
Hubkaw: fixed

src/Aetherphone/Localization/en.json
Hubkaw: added to all locales, used online translator

src/Aetherphone/Apps/Music/MusicApp.Home.cs
Hubkaw: fixed, i checked with multiple phone/text sizes including max and min and it looked good. Also added tooltips

src/Aetherphone/Apps/Music/MusicApp.Home.cs
Hubkaw: fixed

src/Aetherphone/Apps/Music/MusicApp.cs
Hubkaw: Done. IDE was throwing warnings at me for using Contains with a struct so i also implemented IEquatable to it

src/Aetherphone/Apps/Music/MusicApp.cs
Hubkaw: Done, and changed abbreviations to "stationRecord"

src/Aetherphone/Core/Radio/RadioPlayer.cs
Hubkaw: fixed

XeldarAlz:
Verified all seven fixes against `ad603ec`:

- **Loc key + locales** - `L.cs` now `music.favoriteStations`, matches `en.json`; all 9 locales carry the 3 new keys, no em dashes, terminology consistent with existing entries.
- **Star overlap** - `trailing` widened to 50f/33f and kept hover-independent so `FitText` won't re-truncate; tooltips added to both stars.
- **Click guard** - folded into the early return as suggested.
- **Per-frame LINQ/closure** — replaced with cached `favoriteRadioStations` field rebuilt on open/toggle; `IEquatable<RadioStation>` + `GetHashCode` make `Contains` allocation-free; the cross-instance index bug in `PlayStationFromFavorites` is gone.
- **Torn struct read** - `CurrentStationInfo` getter and `Stop()` assignment now both under `lock (gate)`.

Nothing blocking. Nice touch adding `IEquatable`/`GetHashCode`. Merging now.

Please tell me your discord name so I can give you a Developer role.

Hubkaw:
@XeldarAlz my discord is @Hubkaw

PR #40, opened 2026-07-22
Fix Linux/Wine music playback: managed Opus/webm decode, fix thumbnail decode

BluntEXE:
## Summary

Made for You and search playback stuck permanently in "Buffering" on Linux/Wine, and thumbnails failed to load. The two bugs are unrelated but both come from native Windows decode paths that don't work under Wine.

### Song playback

`SelectAudioStream` now prefers YouTube's Opus/webm audio stream over mp4/AAC. `OpusWebmSampleProvider` decodes it through Concentus (already a dependency, used for Telephony) and a new minimal WebM demuxer (NEbml), both pure managed with no native dependency. mp4/AAC through `MediaFoundationReader` stays as a fallback for the rare video with no Opus stream.

`ForwardSeekableStream` works around a bug in YoutubeExplode 6.6.0: `MediaStream.Read()` hangs forever on a zero-byte read request, which NEbml's EBML parser issues constantly and correctly (any 1-byte-length VInt needs zero extra bytes). This was the actual cause of the permanent "Buffering" hang. I isolated the exact call and reproduced the hang against a live stream to confirm it before writing the fix.

Two more demuxer bugs turned up during that live-network testing:
- Real WebM files lead with an EBML header element before `Segment`. The synthetic unit tests I wrote first didn't cover this, so they passed anyway.
- Packet reads across cluster boundaries reported end-of-stream after the first cluster instead of continuing into the next one.

The disk cache key is now namespaced by codec, so any mp4 bytes cached before this change can't get misread as Opus.

### Thumbnails

`MediaCache` now decodes with SixLabors.ImageSharp (already a dependency) and hands Dalamud already-decoded pixels through `CreateFromRawAsync`, instead of raw JPEG bytes through `CreateFromImageAsync`. The latter goes through a native WIC/COM path that failed 100% of the time for these thumbnails under Wine.

## Testing

- 44 existing tests and 2 new ones pass.
- Confirmed header parsing, 3000 sequential packet reads across cluster boundaries, and thumbnail decode against live YouTube data for the videos from the original bug report.
- The reporter confirmed it working in-game on their Linux/Wine setup.

**Not tested:** the mp4/AAC fallback path (no video in testing lacked an Opus stream), and seek/resume on a webm/Opus track mid-song (Update: seek/resume working as intended using Linux. Still worth a full pass on Windows to ensure fix is not OS dependant)

### DIFF
--- a/src/Aetherphone/Aetherphone.csproj
+++ b/src/Aetherphone/Aetherphone.csproj
@@ -52,6 +52,7 @@
 
   <ItemGroup>
     <PackageReference Include="Concentus" Version="2.2.2" />
+    <PackageReference Include="NEbml" Version="0.11.0" />
     <PackageReference Include="NAudio.WinMM" Version="2.3.0" />
     <PackageReference Include="NAudio.Wasapi" Version="2.3.0" />
     <PackageReference Include="NetStone" Version="1.4.1" />
--- a/src/Aetherphone/Core/Net/MediaCache.cs
+++ b/src/Aetherphone/Core/Net/MediaCache.cs
@@ -1,10 +1,7 @@
 using System.Collections.Concurrent;
 using Aetherphone.Core.Media;
-using Dalamud.Interface.Textures;
 using Dalamud.Interface.Textures.TextureWraps;
 using Dalamud.Plugin.Services;
-using SixLabors.ImageSharp;
-using SixLabors.ImageSharp.PixelFormats;
 
 namespace Aetherphone.Core.Net;
 
@@ -86,7 +83,7 @@ internal sealed class MediaCache : IDisposable
                 return;
             }
 
-            var wrap = await DecodeToTextureAsync(bytes, key, token).ConfigureAwait(false);
+            var wrap = await ImageProcessor.DecodeToTextureAsync(textures, bytes, key, token).ConfigureAwait(false);
             if (!ready.TryAdd(key, wrap))
             {
                 wrap.Dispose();
@@ -110,29 +107,6 @@ internal sealed class MediaCache : IDisposable
         {
             inFlight.TryRemove(key, out _);
         }
-    }
-
-    /// <summary>
-    /// Decodes with ImageSharp (pure managed, cross-platform) and hands Dalamud already-decoded
-    /// pixels via CreateFromRawAsync, instead of raw JPEG bytes via CreateFromImageAsync. The
-    /// latter goes through a native WIC/COM decode path that has been observed to fail 100% of
-    /// the time for these thumbnails under Wine, in a way that (unlike the search/network code)
-    /// is not itself the cause of any exception this method's catch block was designed for -
-    /// it just never produces a usable texture. Decoding is CPU-bound, so run it off whatever
-    /// thread is awaiting this rather than blocking it.
-    /// </summary>
-    private async Task<IDalamudTextureWrap> DecodeToTextureAsync(byte[] bytes, string key, CancellationToken token)
-    {
-        var (pixels, width, height) = await Task.Run(() =>
-        {
-            using var image = Image.Load<Rgba32>(bytes);
-            var buffer = new byte[image.Width * image.Height * 4];
-            image.CopyPixelDataTo(buffer);
-            return (buffer, image.Width, image.Height);
-        }, token).ConfigureAwait(false);
-
-        return await textures.CreateFromRawAsync(RawImageSpecification.Rgba32(width, height), pixels, key, token)
-            .ConfigureAwait(false);
     }
 
     public void Dispose()
--- a/src/Aetherphone/Core/Net/MediaCache.cs
+++ b/src/Aetherphone/Core/Net/MediaCache.cs
@@ -1,7 +1,10 @@
 using System.Collections.Concurrent;
 using Aetherphone.Core.Media;
+using Dalamud.Interface.Textures;
 using Dalamud.Interface.Textures.TextureWraps;
 using Dalamud.Plugin.Services;
+using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.PixelFormats;
 
 namespace Aetherphone.Core.Net;
 
@@ -83,7 +86,7 @@ internal sealed class MediaCache : IDisposable
                 return;
             }
 
-            var wrap = await textures.CreateFromImageAsync(bytes, key, token).ConfigureAwait(false);
+            var wrap = await DecodeToTextureAsync(bytes, key, token).ConfigureAwait(false);
             if (!ready.TryAdd(key, wrap))
             {
                 wrap.Dispose();
@@ -107,6 +110,29 @@ internal sealed class MediaCache : IDisposable
         {
             inFlight.TryRemove(key, out _);
         }
+    }
+
+    /// <summary>
+    /// Decodes with ImageSharp (pure managed, cross-platform) and hands Dalamud already-decoded
+    /// pixels via CreateFromRawAsync, instead of raw JPEG bytes via CreateFromImageAsync. The
+    /// latter goes through a native WIC/COM decode path that has been observed to fail 100% of
+    /// the time for these thumbnails under Wine, in a way that (unlike the search/network code)
+    /// is not itself the cause of any exception this method's catch block was designed for -
+    /// it just never produces a usable texture. Decoding is CPU-bound, so run it off whatever
+    /// thread is awaiting this rather than blocking it.
+    /// </summary>
+    private async Task<IDalamudTextureWrap> DecodeToTextureAsync(byte[] bytes, string key, CancellationToken token)
+    {
+        var (pixels, width, height) = await Task.Run(() =>
+        {
+            using var image = Image.Load<Rgba32>(bytes);
+            var buffer = new byte[image.Width * image.Height * 4];
+            image.CopyPixelDataTo(buffer);
+            return (buffer, image.Width, image.Height);
+        }, token).ConfigureAwait(false);
+
+        return await textures.CreateFromRawAsync(RawImageSpecification.Rgba32(width, height), pixels, key, token)
+            .ConfigureAwait(false);
     }
 
     public void Dispose()
--- a/src/Aetherphone/Core/Songs/ForwardSeekableStream.cs
+++ b/src/Aetherphone/Core/Songs/ForwardSeekableStream.cs
@@ -19,8 +19,8 @@ namespace Aetherphone.Core.Songs;
 internal sealed class ForwardSeekableStream : Stream
 {
     private readonly Stream inner;
+    private readonly byte[] discardBuffer = new byte[8192];
     private long position;
-    private byte[] discardBuffer = Array.Empty<byte>();
 
     public ForwardSeekableStream(Stream inner) => this.inner = inner;
 
@@ -64,11 +64,6 @@ internal sealed class ForwardSeekableStream : Stream
         var remaining = offset;
         while (remaining > 0)
         {
-            if (discardBuffer.Length == 0)
-            {
-                discardBuffer = new byte[8192];
-            }
-
             var chunk = Read(discardBuffer, 0, (int)Math.Min(remaining, discardBuffer.Length));
             if (chunk <= 0)
             {
--- a/src/Aetherphone/Core/Songs/ForwardSeekableStream.cs
+++ b/src/Aetherphone/Core/Songs/ForwardSeekableStream.cs
@@ -1,0 +1,97 @@
+namespace Aetherphone.Core.Songs;
+
+/// <summary>
+/// Wraps a forward-only network stream so NEbml's EbmlReader can skip past elements without
+/// tearing down and re-opening the underlying HTTP connection.
+///
+/// NEbml's ReadNext() calls Skip() on every call, and Skip() always tries Stream.Seek() first
+/// with no fallback if it throws. YoutubeExplode's MediaStream reports CanSeek = true, but
+/// implements seeking by discarding the current HTTP connection and issuing a brand new
+/// GetStreamAsync request for "current position to end of file" on the next Read() - meaning
+/// every skipped WebM element (most TrackEntry sub-fields, SeekHead, etc.) was triggering a
+/// fresh HTTP connection. A single TrackEntry has enough of those to open 5-10+ connections
+/// just parsing the header, which is what was hanging (or getting throttled by YouTube's CDN)
+/// instead of a normal single continuous download.
+///
+/// This wrapper reports CanSeek = true (so NEbml's fast path succeeds) but implements forward
+/// Seek by reading-and-discarding from the same already-open connection.
+/// </summary>
+internal sealed class ForwardSeekableStream : Stream
+{
+    private readonly Stream inner;
+    private long position;
+    private byte[] discardBuffer = Array.Empty<byte>();
+
+    public ForwardSeekableStream(Stream inner) => this.inner = inner;
+
+    public override bool CanRead => true;
+    public override bool CanSeek => true;
+    public override bool CanWrite => false;
+    public override long Length => inner.Length;
+
+    public override long Position
+    {
+        get => position;
+        set => throw new NotSupportedException("Only Seek(offset, SeekOrigin.Current) with a non-negative offset is supported.");
+    }
+
+    public override int Read(byte[] buffer, int offset, int count)
+    {
+        // A zero-byte read must always return 0 with no I/O per the Stream contract - and NEbml
+        // does this constantly (any 1-byte-length EBML VInt needs zero extra bytes). YoutubeExplode
+        // 6.6.0's MediaStream.ReadAsync doesn't handle count=0 correctly: it treats "read 0 bytes"
+        // as "connection needs resetting" and loops forever re-establishing the HTTP connection,
+        // since a 0-byte request can only ever return 0. Short-circuiting here avoids ever handing
+        // it a request it can't complete.
+        if (count == 0)
+        {
+            return 0;
+        }
+
+        var read = inner.Read(buffer, offset, count);
+        position += read;
+        return read;
+    }
+
+    public override long Seek(long offset, SeekOrigin origin)
+    {
+        if (origin != SeekOrigin.Current || offset < 0)
+        {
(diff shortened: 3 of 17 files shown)


src/Aetherphone/Core/Songs/OpusWebmSampleProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private void SeekTo(double targetSeconds)
+    {
+        if (targetSeconds < positionSeconds)
+        {
+            // Can't seek backwards without re-opening the stream from the start; the caller
+            // (SongPlayer) only seeks forward in practice (scrubbing ahead, resuming after a
+            // retry), so this is a documented limitation rather than a bug fix target here.
+            throw new NotSupportedException("Backward seeking is not supported by OpusWebmSampleProvider.");
```
XeldarAlz: **Blocking.** `MusicApp.NowPlaying.cs:289` is a draggable progress slider that seeks in both directions, so the assumption above does not hold.

`PlayOnce` sets `reader.CurrentTime` with no try/catch, so this throws into `Run`'s generic catch, which retries from the *old* position. Dragging back restarts the track; doing it twice lands in `Failed`. Opus is now preferred on every platform, so this regresses Windows too.

Also: `positionSeconds` is the decode position, ahead of audible by the NAudio buffer, so even a forward-looking scrub can land behind it and throw.

Fix: reopen the source and forward-seek from zero.

src/Aetherphone/Core/Songs/OpusWebmSampleProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        this.source = source;
+        this.channels = channels;
+        demuxer = new WebmOpusDemuxer(source);
+        demuxer.ReadHeader();
+#pragma warning disable CS0618 // Obsolete in favor of OpusCodecFactory, which P/Invokes a native
+                              // libopus when present. That's the exact native/Wine-fragile
+                              // dependency this class exists to avoid, so the plain managed
+                              // constructor is the deliberate choice here, not an oversight.
+        decoder = new OpusDecoder(sampleRate, channels);
```
XeldarAlz: `Core/Telephony/Audio/OpusAudio.cs:16` already sets `OpusCodecFactory.AttemptToUseNativeLibrary = false`, so `CreateDecoder` gives you the same managed guarantee without the obsolete API or the pragma. A `CreateDecoder(int channels)` overload there keeps Concentus config in one place.

src/Aetherphone/Core/Songs/OpusWebmSampleProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+            return false;
+        }
+
+        var samplesNeeded = OpusFrameSizeSamples * channels;
+        if (pending.Length < samplesNeeded)
+        {
+            pending = new float[samplesNeeded];
+        }
+
+        var decodedPerChannel = decoder.Decode(packet.AsSpan(), pending.AsSpan(0, samplesNeeded), OpusFrameSizeSamples);
```
XeldarAlz: 960 is 20ms; Opus packets go up to 120ms (5760 samples). If one arrives, Concentus throws `OPUS_BUFFER_TOO_SMALL` on the render thread, where `Run`'s retry cannot see it. Sizing `pending` for 5760 (~46 KB, once) removes the assumption for free.

src/Aetherphone/Core/Songs/OpusWebmSampleProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+
+    public TimeSpan TotalTime { get; }
+
+    public TimeSpan CurrentTime
+    {
+        get => TimeSpan.FromSeconds(positionSeconds);
+        set => SeekTo(value.TotalSeconds);
+    }
+
+    public OpusWebmSampleProvider(Stream source, int channels = 2, int sampleRate = 48_000)
```
XeldarAlz: `channels = 2` is fixed, but the demuxer already walks `TrackEntry` - reading `Audio`/`Channels` (0xE1/0x9F) is a few lines. `CodecPrivate` pre-skip is also ignored, so tracks start with a few ms of warm-up artefacts.

src/Aetherphone/Core/Songs/SongPlayer.cs
```diff
@@ -401,33 +414,48 @@ private void BeginCacheFill(string videoId, CancellationToken token)
...
         }
 
+        var isOpus = IsOpus(best);
         using var source = youtube.Videos.Streams.GetAsync(best, token).GetAwaiter().GetResult();
         using var memory = new MemoryStream();
         source.CopyToAsync(memory, token).GetAwaiter().GetResult();
         var bytes = memory.ToArray();
-        cache.Set(videoId, bytes);
-        return bytes;
+        cache.Set(isOpus ? OpusCacheKey(videoId) : videoId, bytes);
```
XeldarAlz, line 432: Line 259 only reads `OpusCacheKey(videoId)`, so this mp4 write is never read back: those videos re-download every play and `BeginCacheFill` writes dead bytes. Try the Opus key first, then fall back to the legacy key with `bytesAreOpus = false` - that also keeps existing users' cached songs.

src/Aetherphone/Core/Net/MediaCache.cs
```diff
@@ -109,6 +112,29 @@ private async Task LoadAsync(string key, Func<CancellationToken, Task<byte[]?>>
...
+    /// <summary>
+    /// Decodes with ImageSharp (pure managed, cross-platform) and hands Dalamud already-decoded
+    /// pixels via CreateFromRawAsync, instead of raw JPEG bytes via CreateFromImageAsync. The
+    /// latter goes through a native WIC/COM decode path that has been observed to fail 100% of
+    /// the time for these thumbnails under Wine, in a way that (unlike the search/network code)
+    /// is not itself the cause of any exception this method's catch block was designed for -
+    /// it just never produces a usable texture. Decoding is CPU-bound, so run it off whatever
+    /// thread is awaiting this rather than blocking it.
+    /// </summary>
+    private async Task<IDalamudTextureWrap> DecodeToTextureAsync(byte[] bytes, string key, CancellationToken token)
```
XeldarAlz: Should be shared rather than private here - six other call sites have the identical bug (see summary). `Core/Media/ImageProcessor.cs` is the existing home for ImageSharp work.

While moving it: this allocates a `W*H*4` managed buffer per image on top of ImageSharp's own (~8 MB on the LOH for a 1080p photo). Fine for thumbnails, worth pooling once Photos and wallpapers share the path.

src/Aetherphone/Core/Net/MediaCache.cs
```diff
@@ -109,6 +112,29 @@ private async Task LoadAsync(string key, Func<CancellationToken, Task<byte[]?>>
         }
     }
 
+    /// <summary>
```
XeldarAlz: I would cut this block. Beyond the repo's comment style, the "in a way that (unlike the search/network code)..." sentence is hard to parse, and none of it is needed to read the method. It is already in the commit message.

src/Aetherphone/Core/Songs/WebmOpusDemuxer.cs
```diff
@@ -0,0 +1,313 @@
...
+
+        reader.LeaveContainer();
+    }
+
+    /// <summary>
+    /// Reads the next raw Opus packet belonging to the audio track, skipping SimpleBlocks/Blocks
+    /// for any other track (there shouldn't be any in an audio-only stream, but don't assume).
+    /// Returns null once the container is exhausted.
+    /// </summary>
+    private bool insideCluster;
```
XeldarAlz: This field sits between the `/// <summary>` and `ReadNextPacket`, so the doc is attached to the field instead of the method. Same at line 90. Both should move up next to `timestampScale` / `opusTrackNumber`.

src/Aetherphone/Core/Songs/WebmOpusDemuxer.cs
```diff
@@ -0,0 +1,313 @@
...
+    /// producing corrupt audio, so this gap is visible if it turns out to matter in practice.
+    /// </summary>
+    private byte[]? ReadBlockPayload()
+    {
+        var size = (int)reader.ElementSize;
+        var buffer = new byte[size];
+        var read = 0;
+        while (read < size)
+        {
+            var n = reader.ReadBinary(buffer, read, size - read);
```
XeldarAlz: House rule: no abbreviations, including loop indexes. `var n` here, `i` at line 305 and in the test helper. `Core/Net/DiskCache.cs` uses `index` throughout.

src/Aetherphone/Core/Songs/ForwardSeekableStream.cs
```diff
@@ -0,0 +1,97 @@
...
+/// instead of a normal single continuous download.
+///
+/// This wrapper reports CanSeek = true (so NEbml's fast path succeeds) but implements forward
+/// Seek by reading-and-discarding from the same already-open connection.
+/// </summary>
+internal sealed class ForwardSeekableStream : Stream
+{
+    private readonly Stream inner;
+    private long position;
+    private byte[] discardBuffer = Array.Empty<byte>();
```
XeldarAlz: `private readonly byte[] discardBuffer = new byte[8192];` drops the length check from the `Seek` loop and matches the repo's `readonly` preference.

The rest of this class is my favourite part of the PR.

src/Aetherphone/Aetherphone.csproj
```diff
@@ -52,6 +52,7 @@
 
   <ItemGroup>
     <PackageReference Include="Concentus" Version="2.2.2" />
+    <PackageReference Include="NEbml" Version="0.11.0" />
```
XeldarAlz, line 55: `NEbml.Core.dll` ships in the Release output, so `THIRD-PARTY-NOTICES.md` needs a matching row in the MIT table (lines 89-99).

XeldarAlz (changes requested):
Great debugging here. Isolating the zero-byte `MediaStream.Read()` hang against a live stream is not easy, and `ForwardSeekableStream` is a clean answer to it.

Verified locally: `build -c Release` clean, `dotnet test` green at 44/44, `NEbml.Core.dll` ships in the output. (Minor: master has 42 tests, so it's 42 + 2, not 44 + 2.)

A few changes before merge. Details are inline.

**Blocking**

1. Backward seek throws and takes playback down with it. Affects Windows too, not just Wine. `OpusWebmSampleProvider.cs:110`
2. Use the existing `OpusAudio` factory instead of the obsolete `OpusDecoder` ctor + pragma. `OpusWebmSampleProvider.cs:50`
3. The thumbnail fix covers 1 of 7 call sites and should be shared. `MediaCache.cs:124`
4. The mp4 cache entry is now written but never read. `SongPlayer.cs:422`

**The other 6 `CreateFromImageAsync` sites** have the identical Wine bug and are still broken after this PR: `RemoteImageCache.cs:120`, `WallpaperImageCache.cs:46`, `WallpaperLibrary.cs:291`, `PhotosApp.cs:328` and `:355`, `PhotosWidget.cs:150`. Put the decode in `Core/Media/ImageProcessor.cs` and route all seven through it.

**Comments**: this repo runs near-zero comments (137 `//` and 84 `///` lines across 700 files). This PR adds 109 in 10 files. Keep the EBML/Matroska ones, those are earned. The rationale narratives belong in the commit message, where you've already written them well.

**If this lands**: add a row to the MIT table in `THIRD-PARTY-NOTICES.md`: `| NEbml | 0.11.0 | Oleg Zee (https://github.com/Oleg-Zee/NEbml) |`

The hard part of this PR is done and done well. The rest is wiring it into what's already in the tree.

src/Aetherphone/Core/Songs/OpusWebmSampleProvider.cs
BluntEXE: Fixed. SeekTo reopens the source and forward-decodes to the target on every call, regardless of direction. This also covers the decode-ahead-of-audible case: it no longer matters whether the target sits behind positionSeconds. Cost: every seek re-reads from the start now, flagged as a follow-up in the summary comment.

src/Aetherphone/Core/Songs/OpusWebmSampleProvider.cs
BluntEXE: Fixed, with one adjustment: I didn't add the CreateDecoder(int channels) overload to OpusAudio.cs, since that file's outside this PR's diff and telephony has no reported issues. Instead OpusWebmSampleProvider calls OpusCodecFactory.CreateDecoder(OpusAudio.SampleRate, channels) directly, referencing OpusAudio.SampleRate to trigger its static constructor (AttemptToUseNativeLibrary = false) without touching the class. Same managed-only guarantee, no changes to telephony code.

src/Aetherphone/Core/Songs/OpusWebmSampleProvider.cs
BluntEXE: Fixed. pending is now sized once for MaxOpusFrameSamples = 5760 (120ms), allocated in the constructor. No more per-frame resize check.

src/Aetherphone/Core/Songs/OpusWebmSampleProvider.cs
BluntEXE: Not done. Deferred: pre-skip needs OpusHead parsing this demuxer doesn't do yet, and reading Audio/Channels without the pre-skip fix only solves half the problem. Left as a known gap rather than a partial fix. Can add it as a follow-up if you want it in this PR instead.

src/Aetherphone/Core/Songs/SongPlayer.cs
BluntEXE: Fixed. PlayOnce tries the Opus key first, falls back to the legacy videoId key with bytesAreOpus = false. Verified the routing itself works: seeded a legacy-keyed entry, confirmed no re-download and no misrouting into the Opus decoder. Couldn't verify actual mp4 decode correctness, since my install never had a real pre-existing mp4 cache entry to test against (the old streamed-mp4 path was the one that hung). If anyone has an old cache entry lying around, worth a real check before merge.

src/Aetherphone/Core/Net/MediaCache.cs
BluntEXE: Fixed. Moved to ImageProcessor.DecodeToTextureAsync(ITextureProvider, byte[], string, CancellationToken); all 7 sites (MediaCache, PhotosApp x2, RemoteImageCache, WallpaperLibrary, WallpaperImageCache, PhotosWidget) call it now. Also pooled the per-image raw-pixel buffer with ArrayPool<byte>.Shared per your note on the LOH allocation, now that it's shared across all of those call sites.

src/Aetherphone/Core/Net/MediaCache.cs
BluntEXE: Done. The whole block left with the method when it moved to ImageProcessor.

src/Aetherphone/Core/Songs/WebmOpusDemuxer.cs
BluntEXE: Fixed. Moved insideCluster and hasPendingFirstCluster up next to timestampScale/opusTrackNumber. ReadNextPacket's doc comment now sits directly above the method.

src/Aetherphone/Core/Songs/WebmOpusDemuxer.cs
BluntEXE: Fixed. n became bytesRead, both i instances became index, including the test helper.

src/Aetherphone/Core/Songs/ForwardSeekableStream.cs
BluntEXE: Fixed. private readonly byte[] discardBuffer = new byte[8192]; dropped the lazy-alloc check from Seek.

src/Aetherphone/Aetherphone.csproj
BluntEXE: Added, the exact row you gave.

BluntEXE:
Addressed all the blocking items and the style notes. Details in the individual replies above.

Comment count came down too: cut the two narrative rationale blocks (OpusWebmSampleProvider's class doc, SongPlayer.SelectAudioStream's rationale) that duplicated the commit message. Kept the EBML/Matroska mechanics comments and ForwardSeekableStream's bug-history comment as earned.

Manual testing (Linux/Wine, xlcore): fresh network stream playback works, forward and backward drags on the progress slider no longer crash. Verified the legacy mp4-cache routing specifically (no re-download, no Opus-decoder misroute) but not actual mp4 decode correctness. See the note on the mp4-cache thread for detail.

Two follow-ups surfaced during testing, neither blocking:
1. Seek now has a real cost. Every seek re-downloads and re-decodes from zero, so cost scales with how far into the song you seek.
2. That catch-up decode causes a visual stutter, not audio: WebmOpusDemuxer.ReadBlockPayload/ReadNextPacket allocate two byte[] per packet, and a deep seek's GC burst likely stalls the game's render thread briefly, since Dalamud plugins share the game's CLR. Reusing a buffer instead of allocating per packet would probably fix it. Can send that as a follow-up PR.

44/44 tests green, clean Release build.

XeldarAlz:
Re-reviewed `aa02707e`. All four blocking items and the style notes are addressed:

- Backward seek reopens the source and forward-seeks from zero (`OpusWebmSampleProvider.cs` `SeekTo`/`Reopen`)
- `OpusCodecFactory.CreateDecoder(OpusAudio.SampleRate, channels)` replaces the obsolete ctor + pragma
- Decode moved to `Core/Media/ImageProcessor.cs`; all 7 `CreateFromImageAsync` sites route through it
- mp4 cache reads `OpusCacheKey` first, then the legacy key with `bytesAreOpus = false` (`SongPlayer.cs`)
- Pending buffer sized for 5760 samples; NEbml row added to `THIRD-PARTY-NOTICES.md`

Merging. I'll take the two non-blocking follow-ups (per-packet `byte[]` allocations in `WebmOpusDemuxer`, and the seek re-decode cost) directly on master. Nice work on `ForwardSeekableStream`.

If you tell me your Discord Username, I can give you a Developer role.

PR #41, opened 2026-07-23
feat(social): tap-to-scroll-top on Aethergram and Chirper feeds

Kiro-XVI:
## What

Tap to jump back to the top of the feed, in both social apps.

- **Aethergram** — tap the Home tab while already on Home (Instagram behaviour). Hooks into the existing `SelectTab` branch that already handled this case.
- **Chirper** — tap the centred header title (Twitter behaviour). Chirper has no tab bar or bottom nav, so there was no equivalent hook; the title had no click handler, so it takes the tap.

## How

A `feedScrollTopPending` flag per app. The tap sets it; `DrawFeedList` consumes it with `ImGui.SetScrollY(0f)` inside the `AppSurface` scope, since that's the only place the correct scroll region is open.

## Notes / open questions

- **The two gestures differ**, because the two apps are shaped differently. Both match their real-world counterparts, but neither is discoverable without already knowing the convention. Happy to swap either for a visible floating button if you'd prefer — Chirper already has `ComposeFab` in the bottom-right, so a second FAB there would need repositioning.
- On Aethergram the tap **also triggers the existing `store.RefreshFeed`**, since that was already wired to that branch. Left as-is; easy to separate if you want scroll without refresh.
- No new strings, so no localisation changes.

## Testing

CachyOS, wine-xiv-staging-fsync-git 10.8, Dalamud 15.0.2.3. Both feeds verified in-game: scroll down, tap, jumps to top.
Closes #

## Checklist

### DIFF
--- a/src/Aetherphone/Apps/Aethergram/AethergramApp.cs
+++ b/src/Aetherphone/Apps/Aethergram/AethergramApp.cs
@@ -86,6 +86,7 @@ internal sealed partial class AethergramApp : IPhoneApp
     private AethergramTab activeTab = AethergramTab.Home;
     private readonly Spring[] navHover = new Spring[NavTabCount];
     private SocialFeedScope activeScope = SocialFeedScope.ForYou;
+    private bool feedScrollTopPending;
     private bool commentFocusPending;
     private readonly PhotoComposeSession composeSession;
     private bool composeAvatarMode;
@@ -364,6 +365,7 @@ internal sealed partial class AethergramApp : IPhoneApp
     {
         if (tab == AethergramTab.Home && activeTab == AethergramTab.Home)
         {
+            feedScrollTopPending = true;
             store.RefreshFeed(activeScope);
             return;
         }
@@ -485,6 +487,11 @@ internal sealed partial class AethergramApp : IPhoneApp
         var snapshot = store.Feed(scope);
         using (AppSurface.Begin(listRect))
         {
+            if (feedScrollTopPending)
+            {
+                ImGui.SetScrollY(0f);
+                feedScrollTopPending = false;
+            }
             stories.DrawTray(theme);
             if (snapshot.Length == 0)
             {
--- a/src/Aetherphone/Apps/Chirper/ChirperApp.Profile.cs
+++ b/src/Aetherphone/Apps/Chirper/ChirperApp.Profile.cs
@@ -108,6 +108,12 @@ internal sealed partial class ChirperApp
         var rowCenterY = area.Min.Y + AppHeader.Height * scale * 0.5f;
         Typography.DrawCentered(new Vector2(area.Center.X, rowCenterY), DisplayName, AppPalettes.Chirper.TitleInk,
             1.3f, FontWeight.Bold);
+        var titleSize = Typography.Measure(DisplayName, 1.3f, FontWeight.Bold);
+        var titleMin = new Vector2(area.Center.X - titleSize.X * 0.5f, rowCenterY - titleSize.Y * 0.5f);
+        if (UiInteract.HoverClick(titleMin, titleMin + titleSize))
+        {
+            feedScrollTopPending = true;
+        }
         if (store.Me is { } me)
         {
             var radius = 16f * scale;
--- a/src/Aetherphone/Apps/Chirper/ChirperApp.Profile.cs
+++ b/src/Aetherphone/Apps/Chirper/ChirperApp.Profile.cs
@@ -106,13 +106,18 @@ internal sealed partial class ChirperApp
     {
         var scale = ImGuiHelpers.GlobalScale;
         var rowCenterY = area.Min.Y + AppHeader.Height * scale * 0.5f;
-        Typography.DrawCentered(new Vector2(area.Center.X, rowCenterY), DisplayName, AppPalettes.Chirper.TitleInk,
-            1.3f, FontWeight.Bold);
-        var titleSize = Typography.Measure(DisplayName, 1.3f, FontWeight.Bold);
-        var titleMin = new Vector2(area.Center.X - titleSize.X * 0.5f, rowCenterY - titleSize.Y * 0.5f);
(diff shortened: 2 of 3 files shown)


src/Aetherphone/Apps/Chirper/ChirperApp.cs
```diff
@@ -279,11 +280,17 @@ private void DrawFeedList(Rect listRect, SocialFeedScope scope)
...
+                feedScrollTopPending = false;
+            }
+
             if (snapshot.Length == 0)
             {
                 var message = store.IsLoading(scope) ? Loc.T(L.Common.Loading) :
-                    scope == SocialFeedScope.Following ? Loc.T(L.Chirper.FollowingEmpty) :
-                    Loc.T(L.Chirper.ExploreEmpty);
+    scope == SocialFeedScope.Following ? Loc.T(L.Chirper.FollowingEmpty) :
+                Loc.T(L.Chirper.ExploreEmpty);
```
XeldarAlz: Unrelated indentation regression, these lines were aligned before. Please restore:

```suggestion
                    scope == SocialFeedScope.Following ? Loc.T(L.Chirper.FollowingEmpty) :
                    Loc.T(L.Chirper.ExploreEmpty);
```

src/Aetherphone/Apps/Chirper/ChirperApp.Profile.cs
```diff
@@ -108,6 +108,12 @@ private void DrawHomeTopBar(Rect area)
         var rowCenterY = area.Min.Y + AppHeader.Height * scale * 0.5f;
         Typography.DrawCentered(new Vector2(area.Center.X, rowCenterY), DisplayName, AppPalettes.Chirper.TitleInk,
             1.3f, FontWeight.Bold);
+        var titleSize = Typography.Measure(DisplayName, 1.3f, FontWeight.Bold);
+        var titleMin = new Vector2(area.Center.X - titleSize.X * 0.5f, rowCenterY - titleSize.Y * 0.5f);
```
XeldarAlz: `1.3f, FontWeight.Bold` is written twice now. Retune the header and the hit rect silently drifts off the text. Bind them to one local:

```suggestion
        const float titleScale = 1.3f;
        var titleCenter = new Vector2(area.Center.X, rowCenterY);
        Typography.DrawCentered(titleCenter, DisplayName, AppPalettes.Chirper.TitleInk, titleScale, FontWeight.Bold);
        var titleSize = Typography.Measure(DisplayName, titleScale, FontWeight.Bold);
        var titleMin = titleCenter - titleSize * 0.5f;
```

Also: `DrawCentered` already measures internally (`Typography.cs:529`), so this measures twice per frame, and it auto-wraps above `AutoWrapWidth` while the hit rect stays single-line. A fixed header-relative rect would avoid both.

src/Aetherphone/Apps/Chirper/ChirperApp.Profile.cs
```diff
@@ -108,6 +108,12 @@ private void DrawHomeTopBar(Rect area)
...
+        if (UiInteract.HoverClick(titleMin, titleMin + titleSize))
+        {
+            feedScrollTopPending = true;
+        }
```
XeldarAlz: The only cue that the title is tappable is the hand cursor. Clickable regions elsewhere get `UiInteract.HoverHighlight` or the hover label `ui.IconButton` uses for the search and bell icons below.

src/Aetherphone/Apps/Aethergram/AethergramApp.cs
```diff
@@ -364,6 +365,7 @@ private void SelectTab(AethergramTab tab)
     {
         if (tab == AethergramTab.Home && activeTab == AethergramTab.Home)
         {
+            feedScrollTopPending = true;
```
XeldarAlz, line 368: Same gesture, two outcomes: refresh plus scroll here, scroll only on Chirper. Preference is to match by kicking `profile.TickRefresh(activeScope)` on the Chirper tap, but either way the two apps should agree.

src/Aetherphone/Apps/Chirper/ChirperApp.cs
```diff
@@ -279,11 +280,17 @@ private void DrawFeedList(Rect listRect, SocialFeedScope scope)
         var snapshot = store.Feed(scope);
         using (AppSurface.Begin(listRect))
         {
+            if (feedScrollTopPending)
+            {
+                ImGui.SetScrollY(0f);
+                feedScrollTopPending = false;
```
XeldarAlz, line 286: Not a blocker: when signed out `DrawHome` returns before `DrawFeedList` (`ChirperApp.cs:223`), so the flag is set with nothing to consume it. Harmless today since the feed is already at the top, worth knowing if this flag gains a second consumer.

XeldarAlz (commented):
Good feature, right hooks. Built the branch with `dotnet build -c Release`: passes.

**Done well:** the pending-flag-consumed-inside-the-scroll-child pattern matches the rest of the repo (`NewsApp.cs:119`, `PhotosApp.Grid.cs:104`), and `UiInteract.HoverClick` is the correct call since it respects the overlay and input gates.

**Before merge**
1. Revert the unrelated indentation change in `ChirperApp.cs`.
2. Bind the duplicated `1.3f, FontWeight.Bold` literals to one local so the hit rect cannot drift off the text.

**Nice to have**
3. A hover affordance on the Chirper title.
4. Pick one refresh behaviour: Aethergram refreshes on tap, Chirper does not.

Details are inline.

**Not this PR:** Velvet is the third social app and has no equivalent hook (`VelvetShell.cs:324` assigns the tab directly). Worth a follow-up, and that is the point to extract a shared helper rather than a third copy of the flag.

On your gesture question: keep both as they are, a second FAB next to `ComposeFab` would be worse. Minor: the repo calls this flag `resetScroll` elsewhere, and the PR body has the template pasted twice.

PR #42, opened 2026-07-23
feat(scroll): phone-style kinetic scrolling across the phone UI

Valiice:
## What

Adds phone-style drag-to-scroll (momentum + pull-to-refresh) to every scrollable surface, and reworks tap handling so a scroll never activates the item under the cursor.

- Kinetic drag-to-scroll with momentum on every list (KineticScroller + DragScrollHost), active only while Lock Position is on so unlocked drags still move the window.
- Release-triggered taps (UiInteract.Click): a drag never fires the row/button/photo under the cursor, across shared primitives and app-local taps in every app.
- Pull-to-refresh on Chirper, Aethergram, and Velvet, with a rubber-band overscroll that springs back.
- No ImGui scrollbar on app surfaces (fixes a width-feedback vibration); slide layers snap to whole pixels so text doesn't re-wrap mid-animation.
- Notifications list drag-scrolls, axis-locked against swipe-to-dismiss.

## Why

Lists only moved via the mouse wheel or a thin scrollbar, which felt nothing like a phone. This adds finger-style drag scrolling with momentum and overscroll, and distinguishes a scroll from a tap (widgets previously fired on mouse-down, so any drag activated them).

## How to test

Open the phone with `/phone` and enable **Lock Position**.

1. Press-drag any list (Chirper, Music, Photos, Inventory, ...) and flick; it scrolls and coasts.
2. Drag-scroll starting on a post/photo/card/row; it scrolls instead of opening, and a tap still opens it.
3. Pull down at the top of a Chirper / Aethergram / Velvet feed; it rubber-bands and refreshes on release.
4. Open a Velvet profile with photos and a long bio and navigate in/out; no vibration, and the About text holds during the slide.
5. Notifications: drag to scroll; swipe still dismisses.

## Checklist

- [ ] Verified in-game
- [ ] README updated (user-visible change)
- [ ] Matches existing style (shared `Windows/Components/` widgets, no `what` comments)


XeldarAlz (commented):
Verified locally on a clean worktree: `dotnet build -c Release` is green and 42/42 tests pass, including the 7 new `KineticScrollerTests`.

The architecture is the right call: a pure, testable `KineticScroller`, a thin `DragScrollHost` adapter, one injection point in `AppSurface`. The whole-pixel snap in `SceneCompositor.DrawLayer` is the correct fix for the mid-slide reflow.

Notes are inline. Three of them I'd treat as blocking: the removed scrollbar-shake guard (`ScrollLayout.cs`), the Settings primitives still firing on mouse-down (`ProfilePage.cs`), and the unconditional `NoScrollbar` while drag is gated on Lock Position (`PhoneWindow.cs`). The rest are smaller.

src/Aetherphone/Windows/Components/ScrollLayout.cs
```diff
@@ -2,24 +2,8 @@
...
-    {
-        var available = ImGui.GetContentRegionAvail().X;
-        if (ImGui.GetScrollMaxY() > 0f)
-        {
-            return available;
-        }
-
-        return available - ImGui.GetStyle().ScrollbarSize;
-    }
+    public static float StableContentWidth() => ImGui.GetContentRegionAvail().X;
```
XeldarAlz: **Blocking.** This is safe only because `AppSurface` now passes `NoScrollbar`. These scroll children don't go through `AppSurface`, still have scrollbars, and are exactly the fixed-column image grids the old gutter reservation was written for:

- `PhotosApp.Grid.cs:97/113` and `:237/251`
- `WallpaperPage.cs:83/159` and `:354/361`
- `EmojiPicker.cs:176/191`

Narrower avail shrinks the cells, the content gets shorter, the scrollbar drops, the content widens again: the same per-frame flip as before. Either give those children `NoScrollbar` too, or reserve the gutter only when the current window actually shows a scrollbar.

src/Aetherphone/Apps/Settings/Pages/ProfilePage.cs
```diff
@@ -167,7 +167,7 @@ private static bool StepperButton(ImDrawListPtr drawList, Vector2 min, float siz
             ImGui.SetMouseCursor(ImGuiMouseCursor.Hand);
         }
 
-        return hovered && ImGui.IsMouseClicked(ImGuiMouseButton.Left);
+        return UiInteract.Click(min, max, hovered);
```
XeldarAlz, line 170: **Blocking, not about this line.** The conversion covers app-local taps but misses the shared primitives that fill every Settings page: `SettingsRow.cs:70,112,140,167,186`, `Toggle.cs:20-23`, `SettingsHero.cs:65`. All Settings pages are inside an `AppSurface`, so pressing a row to drag-scroll opens it, and pressing a toggle flips it, on the press frame before the 6px threshold is reached.

`Toggle` also uses a raw `IsMouseHoveringRect` instead of `UiInteract.Hover`, so it ignores `InputBlocked` as well.

Same pattern left in `MarketRowViews.cs:70/77/125`, `LinkshellRow.cs:126`, `SegmentStrip.cs:51`, `PersonPicker.cs:129`, `TextButton.cs:27`.

src/Aetherphone/Windows/PhoneWindow.cs
```diff
@@ -130,7 +130,9 @@ public override void PreDraw()
         var size = minimized ? MinimizeTransition.MinimizedSize : PhoneSizeCatalog.SizeFor(configuration.PhoneScale);
         Size = size;
         SizeCondition = ImGuiCond.Always;
-        Flags = !minimized && configuration.LockPosition ? BaseFlags | ImGuiWindowFlags.NoMove : BaseFlags;
+        var locked = !minimized && configuration.LockPosition;
+        Flags = locked ? BaseFlags | ImGuiWindowFlags.NoMove : BaseFlags;
+        Components.DragScrollHost.Enabled = locked;
```
XeldarAlz, line 123: **Blocking.** Drag-scroll is gated on `locked`, but the `NoScrollbar` flag in `AppSurface` is unconditional. A user who keeps the window unlocked loses the scrollbar and gains nothing: wheel only, with no indication of how long the list is.

Either gate `NoScrollbar` on the same flag, or draw a slim fading indicator to replace it.

src/Aetherphone/Windows/Components/UiInteract.cs
```diff
@@ -29,15 +34,58 @@ public static bool HoverOverlay(Rect rect)
...
-    public static bool HoverClick(Vector2 min, Vector2 max)
+    /// <summary>
+    /// Release-triggered activation for a rect the caller has already hover-tested, so a drag can
+    /// cancel the pending tap before it fires. Pass the same rect <paramref name="hovered"/> was
+    /// computed from.
+    /// </summary>
+    public static bool Click(Vector2 min, Vector2 max, bool hovered)
     {
-        if (!Hover(min, max))
+        if (!ImGui.IsMouseDown(ImGuiMouseButton.Left) && !ImGui.IsMouseReleased(ImGuiMouseButton.Left))
```
XeldarAlz, line 54: The pending tap is one global slot keyed by screen position alone, so it can activate a widget the press never landed on: press a row, wheel-scroll while still holding (the 6px threshold is never crossed, so nothing cancels it), release, and whichever row now sits under that pixel fires. An async feed update that shifts the list does the same.

It also makes overlapping hit rects resolve first-drawn-wins, since the first `Click` to activate clears the flag. That's the inverse of z-order.

Keying the pending tap to a widget identity (`ImGui.GetID`, or an id from the caller) rather than a point would close both.

src/Aetherphone/Windows/Components/AppSurface.cs
```diff
@@ -11,8 +11,11 @@ public static SurfaceScope Begin(Rect area)
     {
         var scale = ImGuiHelpers.GlobalScale;
         ImGui.SetCursorScreenPos(area.Min);
+        var key = ImGui.GetID("##appSurface");
         var padding = ImRaii.PushStyle(ImGuiStyleVar.WindowPadding, new Vector2(16f * scale, 8f * scale));
-        var child = ImRaii.Child("##appSurface", area.Size, false, ImGuiWindowFlags.NoBackground);
+        var child = ImRaii.Child("##appSurface", area.Size, false,
+            ImGuiWindowFlags.NoBackground | ImGuiWindowFlags.NoScrollbar);
+        DragScrollHost.Begin(key);
```
XeldarAlz: Since this is the only place `DragScrollHost.Begin` is called, every surface that doesn't use `AppSurface` is left out: `PhotosApp.Grid.cs:97/237`, `MusicApp.Playlists.cs:254/341`, `MusicApp.NowPlaying.cs:34/143`, `SkywatcherApp.cs:78`, `EmojiPicker.cs:176`, `WallpaperPage.cs:83/354`, `ConductGateOverlay.cs:160`.

Worth calling out because the PR description says "every scrollable surface" and lists Photos under How to test.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -216,13 +216,12 @@ private void DrawPhotoRow(ImDrawListPtr drawList, GridBand band, float leftX, fl
...
-            if (!hovered)
+            if (hovered)
             {
-                continue;
+                ImGui.SetMouseCursor(ImGuiMouseCursor.Hand);
             }
 
-            ImGui.SetMouseCursor(ImGuiMouseCursor.Hand);
-            if (ImGui.IsMouseClicked(ImGuiMouseButton.Left))
+            if (UiInteract.Click(min, max, hovered))
```
XeldarAlz, line 227: The tap conversion landed here, but Photos never opens an `AppSurface` (`##photoGrid` at :97, `##photoAlbums` at :237), so this grid gets neither drag-scroll nor `NoScrollbar`. It's the one surface the How to test section names explicitly.

src/Aetherphone/Windows/Components/DragScrollHost.cs
```diff
@@ -0,0 +1,123 @@
...
+            currentPull = 0f;
+            currentDragging = false;
+            return;
+        }
+
+        var io = ImGui.GetIO();
+        var dt = io.DeltaTime;
+        var pointerY = io.MousePos.Y;
+        var down = ImGui.IsMouseDown(ImGuiMouseButton.Left);
+        var widgetOwnsClick = ImGui.IsAnyItemActive();
```
XeldarAlz, line 126: `IsAnyItemActive()` kills the drag whenever the press lands on a real ImGui item, so the `InvisibleButton` picker grids can only be scrolled by starting on a gap between thumbnails: `FeedbackApp.cs:305`, `ImagePickCrop.cs:123`, `ChatThreadView.cs:657`, `AethergramApp.Compose.cs:116`, `DevApp.Chat.cs:503`.

src/Aetherphone/Windows/Components/DragScrollHost.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            if (region.Pressed)
+            {
+                var wasDragging = scroller.IsDragging;
+                scroller.Release();
+                region.Pressed = false;
+                shouldBlock = wasDragging;
+            }
+
+            scroller.Tick(dt);
```
XeldarAlz: `Tick` only runs in the `!down` branch, so holding the button anywhere else freezes an in-flight fling while it stays `IsControlling`. Combined with the forced `SetScrollY` below, the wheel is also dead for the whole duration of a fling.

src/Aetherphone/Windows/Components/DragScrollHost.cs
```diff
@@ -0,0 +1,123 @@
...
+        public readonly KineticScroller Scroller = new();
+        public int LastFrame = -2;
+        public bool Pressed;
+    }
+
+    private static readonly Dictionary<uint, Region> Regions = new();
+    private static float currentPull;
+    private static bool currentDragging;
+
+    public static float CurrentPull => currentPull;
```
XeldarAlz: These are last-`Begin`-wins statics, so `PullToRefresh.Draw` reads the right region only by call ordering. Returning the region (or the pull/dragging pair) from `Begin` would make that structural rather than conventional.

Minor: `Region.LastFrame` is tracked but never used to evict, so the dictionary only grows.

src/Aetherphone/Core/Animation/KineticScroller.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+        else
+        {
+            offset = MathF.Min(target, maxOffset);
+        }
+
+        if (dt > 0f)
+        {
+            var instant = -dy / dt;
+            velocity = velocity * VelocitySmoothing + instant * (1f - VelocitySmoothing);
```
XeldarAlz: The smoothing factor is applied per frame rather than per `dt`, so the same flick produces a different velocity at 60 fps and at 144 fps. `MathF.Exp(-rate * dt)` would make it frame-rate independent, like the decay in `Tick` already is.

src/Aetherphone/Windows/Components/PullToRefresh.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    public void Draw(Rect area, float pull, bool dragging, bool loading, Vector4 ink, Action onRefresh)
+    {
+        var scale = ImGuiHelpers.GlobalScale;
+        var dt = ImGui.GetIO().DeltaTime;
+
+        if (refreshing)
+        {
+            spinnerElapsed += dt;
+            if (!loading && spinnerElapsed >= MinSpinnerSeconds)
```
XeldarAlz: If `loading` stays true (stalled request), the spinner never stops: there's a minimum duration but no maximum.

src/Aetherphone.Tests/KineticScrollerTests.cs
```diff
@@ -0,0 +1,124 @@
+using Aetherphone.Core.Animation;
+using Xunit;
+
+namespace Aetherphone.Tests;
+
+public class KineticScrollerTests
+{
+    private static KineticScroller Scroller(float bounds)
+    {
+        var s = new KineticScroller();
```
XeldarAlz: Nit, style guide: `s` here and the `for (var i` loops below want spelling out, same for `dt` / `dy` / `d` in `KineticScroller` and `NotificationCenter`.

Valiice:
Thanks for the thorough review, and for building it locally. Pushed a commit addressing everything; details inline per thread.

Summary: all three blocking items are fixed (lock-aware scrollbar/width, Settings primitives release-gated, scrollbar gated on lock). Also fixed the fling/wheel handling, fps-independent velocity, spinner cap, region eviction, the naming nits, and the newly-merged EmojiPicker/ConductGateOverlay. Extended drag-scroll to Photos, Wallpaper, the music and emoji grids, ConductGateOverlay, and Skywatcher. The only item left partial is the pending-tap re-keying, noted below. I also merged latest master so the branch is current.

XeldarAlz:
Re-checked at ffe9c79: Release build succeeds (0 errors, 4 warnings), 66/66 tests pass.

Eight of the twelve findings from the first pass are fixed, including all three blocking ones. The gutter is now reserved only when the window has no scrollbar (`ScrollLayout.cs:12-16`), `NoScrollbar` follows the lock state (`DragScrollHost.ScrollFlags`, `AppSurface.cs:17`), and the shared Settings primitives are all on `UiInteract` now. Drag-scroll also reaches Photos, Music playlists, Skywatcher, the emoji picker, wallpapers and the conduct gate. Velocity smoothing is frame-rate independent (`KineticScroller.cs:100`), the spinner has a 20s ceiling (`PullToRefresh.cs:13`), the wheel cancels momentum instead of fighting it (`DragScrollHost.cs:82-89`), and stale regions are evicted.

**Blocking before merge**

- Merge conflict with master in `src/Aetherphone/Apps/Jobs/JobsApp.cs`, against ef4af7d.

**Non-blocking, still open from the first pass**

- `UiInteract.cs:51,60`: the pending tap is still keyed by screen position. The wheel case is closed by `CancelPendingTap`, but a list that shifts under a held pointer still activates whichever row lands on that pixel, and overlapping rects still resolve first-drawn-wins.
- `DragScrollHost.cs:78,112`: the `IsAnyItemActive()` gate is unchanged, and the `InvisibleButton` grids are still there in `FeedbackApp`, `ImagePickCrop` and `ChatThreadView`. This may be moot since `Begin` runs before the content draws, so one manual drag over a thumbnail would settle it.
- `DragScrollHost.cs:22-23`: `currentPull` and `currentDragging` are still last-`Begin`-wins, so `PullToRefresh.Draw` reads the right region only by call ordering.

Nits: `dt` survives at `DragScrollHost.cs:75` and `PullToRefresh.cs:23`.

Valiice:
@XeldarAlz All three remaining findings are addressed:

- **Pending tap** (`UiInteract.cs`): no longer keyed by screen position. The press is claimed by the rect itself, stored in the window's content space, so a list that scrolls or flings under a held pointer can't hand the tap to whatever drifted onto that pixel. Overlapping rects now resolve last-claim-wins, matching the draw order that put the top one on top. No id parameter, so no churn across the ~60 call sites.
- **`IsAnyItemActive` / `InvisibleButton`**: all of them are gone. The three you listed plus four with the same problem, namely the Aethergram profile grid, `PhotoComposeSession`, `DevApp.Chat`, and the Calendar day cells, which sit inside `AppSurface` too.
- **`currentPull` / `currentDragging`**: removed. `Begin` returns a `Surface` carrying that region's pull, drag state and `JumpToTop`, threaded through `AppSurface`, so call ordering no longer matters. The one consumer that genuinely runs before any surface begins uses a `DragScrollHost.AnyDragging` scan instead.

Also fixed `dt` in both spots, and merged master through `8692cb6`. The new Skywatcher Control tab put a horizontal scrub slider inside a vertical drag-scroll surface, so `Surface.CancelDrag()` hands the gesture to the scrubber without moving the scroll offset, covered by a test.

XeldarAlz:
All three blocking items from my review, plus the smaller notes, are addressed as of `7afb24c`. Verified on a clean checkout: `dotnet build -c Release` is green and 67/67 tests pass.

**Blocking, all resolved:**

- Scrollbar-shake guard (`ScrollLayout.cs`): the width-driven image grids now get `NoScrollbar` while the phone is locked. Photos (`PhotosApp.Grid.cs:99,241`) and the wallpaper photo picker (`WallpaperPage.cs:359`) go through `DragScrollHost` directly; the wallpaper grid and emoji picker inherit it from their host. `StableContentWidth` reserves the gutter only on the unlocked path, when the window actually shows a scrollbar.
- Settings primitives firing on mouse-down (`ProfilePage.cs`): `SettingsRow`, `Toggle`, `SettingsHero`, `SegmentStrip`, `TextButton`, `MarketRowViews`, `LinkshellRow` and `PersonPicker` now activate on release through `UiInteract.Click`, and `Toggle` uses `UiInteract.Hover` so it honors `InputBlocked`.
- Unconditional `NoScrollbar` (`PhoneWindow.cs:123`): `DragScrollHost.ScrollFlags` gates the flag on the lock state (`DragScrollHost.Enabled = locked`), so an unlocked window keeps its scrollbar.

**Smaller notes, all resolved:**

- The pending tap is now claimed by rect in content space and dropped when the window moves under the pointer (`UiInteract.cs`), so a shifting list or a fling can no longer hand it to the wrong widget.
- Coverage extends to Photos, Emoji, Skywatcher, Music playlist rows, Wallpaper and the conduct gate. No `InvisibleButton` grids remain, so picker thumbnails drag-scroll (`Dummy` + `UiInteract.Click`, no active-item conflict).
- `DragScrollHost` state is per-region keyed by surface id with frame eviction; `Tick` runs during a fling and the wheel cancels momentum.
- `KineticScroller` velocity smoothing is frame-rate independent (`MathF.Exp(-rate * dt)`), `PullToRefresh` caps the spinner at 20s, and the naming nits are cleaned up in `KineticScroller`, the tests, and `NotificationCenter`.

Please tell me your discord name so I can give you a Developer role in the server.

Valiice:
@XeldarAlz it's icevali, im in the server already

PR #47, opened 2026-07-23
fix(app/dailies): Use RowId instead of CompletionArrayIndex

Syrilai:
## What

This makes the `DailyBonusRouleteIndices` method return the `RowId` of a row instead of the `CompletionArrayIndex`.

## Why

`CompletionArrayIndex` is used in `IPlayerState`, and not in the native struct methods. Those methods use the `RowId` instead.

This closes the bug report posted on [discord](https://discord.com/channels/1527213298087493732/1529524429044973689).

## How to test

1. Open the phone via any preferred method of choice
2. Navigate to the Dailies App
3. Check on both a character that did not do the roulette, and one that did do it.

## Checklist

### DIFF
--- a/src/Aetherphone/Core/Game/GameData.cs
+++ b/src/Aetherphone/Core/Game/GameData.cs
@@ -416,7 +416,7 @@ internal sealed class GameData
                 continue;
             }
 
-            indices.Add((byte)row.CompletionArrayIndex);
+            indices.Add((byte)row.RowId);
         }
 
         dailyBonusRouletteIndices = indices.Count > 0 ? indices.ToArray() : Array.Empty<byte>();


XeldarAlz (approved):
Verified against FFXIVClientStructs: `IsRouletteComplete(byte)` wraps the native `IsRouletteIncomplete` member function, which takes the ContentRoulette RowId. `CompletionArrayIndex` only indexes the raw `PlayerState._contentRouletteCompletion` buffer, which we never read directly. DailyDuty, SimpleTweaks and Altoholic all pass `(byte)row.RowId` as well, so the fix is right.

Non-blocking, naming: `GameData.cs:400` `DailyBonusRouletteIndices()` and `DailiesReader.cs:133` `bonusIndices` now carry row ids, not indices. I will rename those in a follow-up so the names stop lying.

Non-blocking, testing: worth one pass on a low-level alt. `IsRouletteComplete` is the inverse of `IsRouletteIncomplete`, so a not-yet-unlocked roulette must not read as complete, otherwise the daily would show done permanently.

Thanks for the catch.

Please give me your discord username so I can give you a Developer role.

Syrilai:
Hey, thanks for merging! My discord tag is `syrilai`.

PR #49, opened 2026-07-23
feat(social): let users reopen community rules from Aethergram/Chirper/Velvet

Kiro-XVI:
## Summary
- Adds a "?" button to the own-profile view in Aethergram, Chirper, and Velvet that reopens the same community-rules popup shown on first open, reusing the existing `ConductGateOverlay` / `L.Conduct.*` content rather than building a second rules screen.
- Placement matches each app's existing header conventions: next to the "Edit Profile" pill for Aethergram/Chirper (mirroring where the Report flag sits on other users' profiles), and in the page header for Velvet (mirroring the same Report-flag placement there).
- New `ConductGateService.ShowRules(appId)` opens the gate on demand, bypassing the `HasAcknowledgedConduct` check used for the first-open flow.
- When reopened this way, the panel skips the read-it countdown and "I understand and agree" button entirely, showing a plain X in the top-right corner to close instead — no re-acknowledgement needed for a voluntary review. The original first-open gate (`NotifyAppOpened`) is untouched and still enforces the countdown + explicit agreement.

## Test plan
- [x] `dotnet build -c Release` / `dotnet test` pass locally (42/42).
- [x] In-game: open Aethergram's own profile, tap the "?" next to Edit Profile, confirm the rules popup appears with just an X (no countdown/agree button) and closes on tap.
- [x] Repeat for Chirper (same placement) and Velvet (header placement).
- [x] Confirm the first-time gate (fresh account / cleared `ConductAcknowledged`) still shows the countdown and requires agreeing.


XeldarAlz:
Reviewed and merged. Build + 44/44 tests green.

One follow-up pushed on top (a9256d0):
- `ConductGateOverlay.DrawCloseButton` hit-tested with a raw `ImGui.IsMouseHoveringRect`, bypassing `UiInteract.InputBlocked` and the overlay reservation every other button in the phone goes through. Swapped it for the shared `AppSkin.IconButton` (also gives the standard hover tint) and gated the dismiss on `opacity > 0.5f` like the agree button, so it cannot fire mid fade-in.
- Folded the duplicated open guard in `ConductGateService` into a single `Open(gate, review)`, and dropped the doc/inline comments to match the surrounding style.

Placement, the review-vs-gate split and the untouched `NotifyAppOpened` path all looked right. Thanks!

PR #50, opened 2026-07-23
feat(collection): Use local data over remote data

Syrilai:
## What

Adds local (in-client) checking of owned/total counts for everything except Achievements, using excel sheet data and `IUnlockState` instead of relying exclusively on the ffxivcollect API. Achievements still go through the API only due to that it would require the user to open their achievements menu to compare unlock states. Total counts from ffxivcollect are prioritized over the ones from the excel sheets.

## Why

Data was not always in-sync, or even unreadable, due to Lodestone privacy rules, and because the data is available locally already, its much better to fetch it from there instead.

This closes the bug report posted on [discord](https://discord.com/channels/1527213298087493732/1529580288525602917).

## How to test

1. Open the phone via any preferred method of choice
2. Navigate to the Collections App
3. It should now show the local data for everything except Achievements

## Checklist

### DIFF
--- a/src/Aetherphone/Core/Collections/CollectionsCatalogService.cs
+++ b/src/Aetherphone/Core/Collections/CollectionsCatalogService.cs
@@ -1,6 +1,8 @@
 using System.Collections.Concurrent;
 using System.Text.Json;
 using Aetherphone.Core.Net;
+using Dalamud.Plugin.Services;
+using Lumina.Excel.Sheets;
 
 namespace Aetherphone.Core.Collections;
 
@@ -26,6 +28,13 @@ internal sealed class CategoryProgress
     public CollectionAccess Access;
 
     public bool HasPercent => Access == CollectionAccess.Public || Count > 0;
+}
+
+internal sealed class CategoryDescriptor
+{
+    public required Func<IDataManager, IEnumerable<object>> GetAll;
+    public required Func<IUnlockState, object, bool> IsUnlocked;
+    public required Func<object, int> GetId;
 }
 
 internal sealed class SummaryEntry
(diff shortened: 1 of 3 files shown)


XeldarAlz:
Merged, thanks. Verified `dotnet build -c Release` clean and checked the id spaces against the live API first: ffxivcollect ids are sheet RowIds for mounts/minions/emotes/orchestrions/facewear/cards, and your `UnlockLink` choice for hairstyles is right (46/46 overlap, RowId would have matched nothing).

Followed up in df59f96 with four fixes.

**Thread safety (the blocking one).** `MakeStatFromLocalData` ran inside `LoadSummaryAsync` after `await http.GetJsonAsync(...).ConfigureAwait(false)`, so all seven scans ran on a threadpool thread. Dalamud's `UnlockState` dereferences `PlayerState.Instance()`/`UIState.Instance()` directly, with no marshalling and no assert to catch you. Local sets are now collected on the framework thread only, cached once per app open, and shared by the summary and the owned lists. `LoadOwned` also ran unguarded on the draw thread, so a throw escaped into the frame instead of being logged and marked failed.

**Hairstyles.** `RowId <= 100 || 2050..2100` yields 46 rows while ffxivcollect lists 52. The six race-specific ones (A Wicked Wake, Clowning Around, Scion Special Issue, Scion Special Issue III, Liberating Locks, Simple and Clean) could never be marked owned, and a completionist read 46/52. Now the whole sheet is scanned and deduped by `UnlockLink`.

**Mixed populations.** `stat?.Total ?? totalInCategory` paired a local count with the remote total. Filter sizes match exactly for mounts (349), orchestrions (862), cards (475) and facewear (57), but not for minions (581 vs 575) or emotes (201 vs 198), so the header could read past 100%. Count and Total now come from the same source.

**The reported bug.** `dto is null` returned before applying anything, and both request paths were gated on `lodestoneId is not null`, so a private or unlisted character still saw nothing. The seven local categories now fill regardless of the fetch; only Achievements still needs the remote profile.

Also replaced the boxing `Cast<object>()` descriptors with typed single-pass scans (no LINQ, no cached boxes, one scan feeding both the counts and the owned set) to match the repo's allocation rules.

Please share your discord username so I can give you Developer role.

Syrilai:
Ahh yeah, good catches on that, thanks! I would have missed it otherwise. 

@XeldarAlz My discord tag is `syrilai`.

PR #52, opened 2026-07-23
Fix Wine encryption and imported audio playback

daviirodrig:
## What this fixes

   Two failures reproduced on Linux with XLCore 1.4.0, wine-xiv 10.8, and .NET 10:

   - Platform P-256 ECDH fails with `0x80090029`, leaving encryption unsupported.
   - `MediaFoundationReader` fails with `0xC00D36BB` for common imported WAV and MP3 files, making those sounds silent.

   ## Changes

   - Use BouncyCastle for managed P-256 ECDH.
   - Preserve interoperability with the existing SPKI public-key format, PKCS#8 private-key format, and `EC1` wrapped-key format.
   - Handle malformed server-provided public, private, and ephemeral keys without interrupting processing of other keys or wraps.
   - Decode PCM and IEEE-float WAV files with `WaveFileReader`.
   - Decode MP3 files with the managed NLayer decoder.
   - Continue using `MediaFoundationReader` for compressed WAV and other audio formats.
   - Add native/managed ECDH interoperability tests and managed WAV/MP3 reader tests.
   - Add the required BouncyCastle and NLayer third-party notices.

   ### PKCS#8 encoding note

   BouncyCastle produces a standards-compatible 150-byte PKCS#8 representation for newly exported P-256 private keys, whereas .NET's `ExportPkcs8PrivateKey`
 produced 138 bytes. Existing 138-byte private keys remain importable, but newly exported keys use BouncyCastle's 150-byte encoding.

   ## Verification

   ### Linux / Wine

   - Encryption reached the active state and survived a restart under Wine.
   - PCM WAV, IEEE-float WAV, and MP3 files played through XLCore using managed readers.
   - Radio and WebM/Opus song playback continued to work.
   - Formats routed through Media Foundation remain dependent on codecs available in the Wine prefix.

   ### Windows

   Reviewer verification at commit `1ce97e4`:

   - Clean Release build.
   - 62/62 tests passed.
   - PCM WAV and MP3 managed paths opened successfully.
   - μ-law WAV and a non-WAV/MP3 format opened through Media Foundation.
   - Not tested in game.


src/Aetherphone/Aetherphone.csproj
```diff
@@ -52,8 +52,10 @@
 
   <ItemGroup>
     <PackageReference Include="Concentus" Version="2.2.2" />
+    <PackageReference Include="BouncyCastle.Cryptography" Version="2.6.2" />
```
XeldarAlz, line 55: Non-blocking: BouncyCastle.Cryptography.dll is 4.8 MB in the Release output, which makes it the largest file we ship, ahead of Aetherphone.dll at 4.5 MB, and pushes the release zip to roughly 13 MB. There is no trimming path for a Dalamud plugin, so this is a conscious accept rather than something to fix, but it is worth deciding on deliberately.

src/Aetherphone/Aetherphone.csproj
```diff
@@ -52,8 +52,10 @@
...
     <PackageReference Include="NEbml" Version="0.11.0" />
     <PackageReference Include="NAudio.WinMM" Version="2.3.0" />
+    <PackageReference Include="NLayer.NAudioSupport" Version="2.0.1" />
```
XeldarAlz, line 58: Blocking: THIRD-PARTY-NOTICES.md needs entries for the three new redistributed DLLs (BouncyCastle.Cryptography 2.6.2, NLayer 2.0.1, NLayer.NAudioSupport 2.0.1). That file ships inside every release archive and already covers transitive deps like JsonExtensions and SharpGen.Runtime, so leaving these out breaks the notice chain the AGPL release depends on. BouncyCastle carries its own MIT-derived licence text, so give it a section of its own rather than a row in the MIT table.

src/Aetherphone/Core/Notifications/SoundEffectPlayer.cs
```diff
@@ -160,6 +161,17 @@ private void RunOnce(string path, float volume)
         }
     }
 
+    internal static WaveStream OpenReader(string path)
+    {
+        return Path.GetExtension(path).ToLowerInvariant() switch
+        {
+            ".wav" => new WaveFileReader(path),
```
XeldarAlz: Blocking: WaveFileReader only covers PCM and IEEE float. I built a mu-law WAV and ran both paths against it: MediaFoundationReader decodes it, this one throws ArgumentException "Unsupported source encoding" out of ToSampleProvider(). The same applies to A-law, ADPCM and MP3-in-WAV, so a player using one of those as a ringtone gets silence plus a warning log and no visible error.

src/Aetherphone/Core/Notifications/SoundEffectPlayer.cs
```diff
@@ -160,6 +161,17 @@ private void RunOnce(string path, float volume)
...
+            ".mp3" => new Mp3FileReaderBase(path,
```
XeldarAlz, line 169: Nit: FrameDecompressorBuilder is handed a WaveFormat, not a frame, so `waveFormat => new Mp3FrameDecompressor(waveFormat)` reads truer to what actually arrives.

src/Aetherphone/Core/Notifications/SoundEffectPlayer.cs
```diff
@@ -160,6 +161,17 @@ private void RunOnce(string path, float volume)
...
+                frame => new Mp3FrameDecompressor(frame)),
+            _ => throw new NotSupportedException($"Unsupported sound format: {Path.GetExtension(path)}"),
```
XeldarAlz: Blocking: this drops every other format Media Foundation used to handle. FilePicker.cs:8 is "Audio{.mp3,.wav},.*", so the all-files option lets a user pick .m4a, .ogg, .flac or .wma, SoundLibrary.AddUserFile copies it without validating the extension, and TryResolvePath then resolves the saved token by filename regardless. Falling back to `new MediaFoundationReader(path)` here instead of throwing keeps Wine fixed and leaves existing Windows setups working.

src/Aetherphone/Core/Crypto/CryptoBox.cs
```diff
@@ -1,8 +1,42 @@
...
+    internal EcPrivateKey(ECPrivateKeyParameters privateKey, ECPublicKeyParameters publicKey)
+    {
+        PrivateKey = privateKey;
+        PublicKey = publicKey;
+    }
+
+    internal ECPrivateKeyParameters PrivateKey { get; private set; }
+    internal ECPublicKeyParameters PublicKey { get; private set; }
+
+    public void Dispose()
```
XeldarAlz: Non-blocking: this only nulls the references. ECDiffieHellman.Dispose() released the underlying key handle, whereas the private scalar now lives on as a managed BigInteger until GC collects it, so the IDisposable shape implies a cleanup that does not happen. Either clear D properly or drop IDisposable and the using statements at the call sites.

src/Aetherphone/Core/Crypto/CryptoBox.cs
```diff
@@ -11,104 +45,120 @@ internal static class CryptoBox
...
-
         try
         {
-            var parameters = new CngKeyCreationParameters
-            {
-                ExportPolicy = CngExportPolicies.AllowPlaintextExport,
-            };
-            return new ECDiffieHellmanCng(CngKey.Create(CngAlgorithm.ECDiffieHellmanP256, null, parameters));
+            var generator = new ECKeyPairGenerator();
+            generator.Init(new ECKeyGenerationParameters(P256, new SecureRandom()));
```
XeldarAlz: Non-blocking: this allocates and auto-seeds a fresh DRBG on every call, and WrapCek now routes through TryGenerateIdentity, so ConversationKeyStore.BuildWraps pays that cost once per group recipient on every rekey. A static readonly SecureRandom covers it, since BouncyCastle's is thread-safe.

src/Aetherphone/Core/Crypto/CryptoBox.cs
```diff
@@ -11,104 +45,120 @@ internal static class CryptoBox
...
         }
     }
 
-    public static byte[]? TryExportPrivateKey(ECDiffieHellman key)
+    public static byte[]? TryExportPrivateKey(EcPrivateKey key)
     {
         try
         {
-            return key.ExportPkcs8PrivateKey();
+            return PrivateKeyInfoFactory.CreatePrivateKeyInfo(key.PrivateKey).GetDerEncoded();
```
XeldarAlz, line 96: Non-blocking: worth a line in the PR description. BouncyCastle emits 150 bytes here where ExportPkcs8PrivateKey emitted 138, because it also writes the redundant [0] parameters field alongside the public key. Harmless in practice since both sides parse both shapes and nothing byte-compares the blob, but the next person to diff a recovery escrow will wonder what changed.

src/Aetherphone/Core/Crypto/CryptoBox.cs
```diff
@@ -186,15 +243,16 @@ public static byte[] GenerateCek()
...
+                    payload.AsSpan(cipherOffset + CekBytes, TagBytes), cek);
                 return cek;
             }
             finally
             {
                 CryptographicOperations.ZeroMemory(wrapKey);
             }
         }
-        catch (CryptographicException)
+        catch (Exception)
```
XeldarAlz: Non-blocking: this was CryptographicException, and the same broadening lands in five other places in this file. Catching everything turns any future bug in the BouncyCastle path into a silent null rather than a log line, which is exactly the class of failure this PR is fixing. Narrowing to what PublicKeyFactory and ECDHBasicAgreement actually throw would keep it visible.

src/Aetherphone.Tests/CryptoBoxTests.cs
```diff
@@ -66,4 +67,81 @@ public void WrappedCekIsFreshPerCall()
         Assert.NotNull(second);
         Assert.NotEqual(first, second);
     }
+    [Fact]
+    public void ManagedP256KeysInteroperateWithPlatformEcdh()
```
XeldarAlz, line 68: Non-blocking: this generates with BouncyCastle and then imports into .NET, which is the reverse of the case that decides whether current players keep their keys. Every existing install has a .NET-generated PKCS#8 in EncryptionKeyCache, so the test worth adding is ECDiffieHellman.Create(nistP256).ExportPkcs8PrivateKey() -> CryptoBox.ImportPrivateKey -> ExportPublicKey equal to the original SPKI base64. I verified it passes, and that string comparison at KeyVault.cs:322 is what stands between Unlocked and Locked.

XeldarAlz (changes requested):
Built at b6a91c2 on Windows: clean Release build, 61/61 tests pass, not tested in game. I also ran a standalone BouncyCastle vs .NET harness on the migration path: an existing player's .NET-generated PKCS#8 imports fine under BouncyCastle and re-exports a byte-identical SubjectPublicKeyInfo, and 200 randomized cross-implementation wrap/unwrap ops agreed in both directions, so current keys survive and old and new clients stay interoperable. Nice work on keeping the formats stable; findings are inline.

daviirodrig:
@XeldarAlz thanks for the review, I fixed the mentioned issues on [1ce97e4](https://github.com/XeldarAlz/FFXIV-Aetherphone/pull/52/commits/1ce97e412ed3c679fe45d997a71d2643338a1cf0)

src/Aetherphone/Core/Crypto/CryptoBox.cs
```diff
@@ -1,112 +1,155 @@
...
         try
         {
-            return Convert.ToBase64String(key.ExportSubjectPublicKeyInfo());
+            var parsed = PublicKeyFactory.CreateKey(Convert.FromBase64String(publicKeyBase64));
+            return parsed is ECPublicKeyParameters publicKey && IsP256(publicKey.Parameters)
+                ? new EcPublicKey(publicKey)
+                : null;
         }
-        catch (CryptographicException)
+        catch (Exception exception) when (exception is ArgumentException or IOException or SecurityUtilityException)
```
XeldarAlz: Blocking: this filter is written against .NET's exception taxonomy, but BouncyCastle's is wider. Fuzzing ~1200 malformed, wrong-curve and bit-flipped keys through here and through ImportPrivateKey (line 118) escapes NullReferenceException on empty input, FormatException, InvalidOperationException ("unexpected object") and Asn1ParsingException. Both sites were catch (Exception) on master, so this is a regression rather than a tightening: ConversationKeyStore.BuildWraps at line 394 feeds server-supplied peer keys straight in and depends on the null return. Restoring catch (Exception) on both is the safe call, my earlier comment was aimed at the newly broadened ones and should not have swept these in.

src/Aetherphone/Core/Crypto/CryptoBox.cs
```diff
@@ -186,15 +236,17 @@ public static byte[] GenerateCek()
...
+                    payload.AsSpan(cipherOffset + CekBytes, TagBytes), cek);
                 return cek;
             }
             finally
             {
                 CryptographicOperations.ZeroMemory(wrapKey);
             }
         }
-        catch (CryptographicException)
+        catch (Exception exception) when (exception is ArgumentException or IOException or SecurityUtilityException
```
XeldarAlz, line 248: Blocking: same root cause, different fix. PublicKeyFactory.CreateKey throws FormatException ("Invalid point encoding") and InvalidCastException on a corrupted ephemeral key, and neither is in this filter; master's catch (CryptographicException) caught them because .NET's ImportSubjectPublicKeyInfo funneled every parse failure into CryptographicException. CacheWraps at ConversationKeyStore.cs:422 loops over server-supplied wraps and is built to skip a bad one, so a single corrupt wrap now throws out of the loop instead of being skipped. Adding FormatException and InvalidCastException closes it.

src/Aetherphone/Core/Crypto/CryptoBox.cs
```diff
@@ -1,112 +1,155 @@
...
 
-    public static ECDiffieHellman? ImportPublicKey(string publicKeyBase64)
+    public static byte[]? TryExportPrivateKey(EcPrivateKey key)
     {
         try
         {
-            var key = ECDiffieHellman.Create();
-            key.ImportSubjectPublicKeyInfo(Convert.FromBase64String(publicKeyBase64), out _);
-            return key;
+            return PrivateKeyInfoFactory.CreatePrivateKeyInfo(key.PrivateKey).GetDerEncoded();
```
XeldarAlz, line 96: Nit: still worth a line in the PR description. BouncyCastle emits 150 bytes here where ExportPkcs8PrivateKey emitted 138, and nothing in the description records the change.

XeldarAlz (changes requested):
Rebuilt at 1ce97e4 on Windows: clean Release build, 62/62 tests pass, not tested in game. I confirmed both audio fallbacks against real files, the mu-law WAV and the non-wav/mp3 extension that used to throw now open through Media Foundation, and the notices, disposal, SecureRandom and migration-test items are all addressed. The exception narrowing I asked for went one step too far and regressed against master; details inline.

PR #53, opened 2026-07-24
feat(gallery): Add a button to open the photos folder

Syrilai:
## What

This adds a simple button in the photos app that allows the user to open the folder where the images are stored in.

## Why

It was suggested in one or more feature requests.

## Important
The translation keys exist, but **are not filled out** for all languages except english and german.

## How to test

1. Open the phone via any preferred method
2. Open the Photos app
3. Click the wonderful button in the bottom right **while** on the library tab.

## Checklist

### DIFF
--- a/src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
+++ b/src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
@@ -2,6 +2,7 @@ using System.Diagnostics;
 using Aetherphone.Core;
 using Aetherphone.Core.Localization;
 using Aetherphone.Core.Onboarding;
+using Aetherphone.Windows;
 using Aetherphone.Windows.Components;
 using Dalamud.Bindings.ImGui;
 using Dalamud.Interface;
@@ -299,12 +300,7 @@ internal sealed partial class PhotosApp
         if (!ComposeFab.Draw(rect, "##openFolderFab", Accent, FontAwesomeIcon.Folder.ToIconString(),
                              Loc.T(L.Photos.OpenFolder), "photos.openFolder")) return;
         
-        Process.Start(new ProcessStartInfo
-        {
-            FileName = "explorer.exe",
-            Arguments = $"\"{library.GetDirectory()}\"",
-            UseShellExecute = true
-        });
+        UrlActions.OpenFolder(library.GetDirectory());
     }
 
     private bool DrawAlbumCard(ImDrawListPtr drawList, Rect rect, string title, int coverStart, int coverCount,
--- a/src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
+++ b/src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
@@ -297,7 +297,7 @@ internal sealed partial class PhotosApp
 
     private void DrawOpenFolder(Rect rect)
     {
-        if (!ComposeFab.Draw(rect, "##openFolderFab", Accent, FontAwesomeIcon.Folder.ToIconString(),
+        if (ComposeFab.Draw(rect, "##openFolderFab", Accent, FontAwesomeIcon.Folder.ToIconString(),
                              Loc.T(L.Photos.OpenFolder), "photos.openFolder"))
         {
             UrlActions.OpenFolder(library.DirectoryPath);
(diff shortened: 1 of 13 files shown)


XeldarAlz (changes requested):
Verified: Release build passes (0 errors, only the pre-existing AngleSharp NU1902 warning), tests 67/67 pass. Not run in-game on my side. Feedback is in the inline comments: two blocking (empty translation entries, explorer.exe launch), the rest minor.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -292,6 +294,19 @@ private void DrawAlbumsGrid(Rect body)
...
     }
 
+    private void DrawOpenFolder(Rect rect)
+    {
+        if (!ComposeFab.Draw(rect, "##openFolderFab", Accent, FontAwesomeIcon.Folder.ToIconString(),
+                             Loc.T(L.Photos.OpenFolder), "photos.openFolder")) return;
+        
+        Process.Start(new ProcessStartInfo
+        {
+            FileName = "explorer.exe",
```
XeldarAlz: Blocking: shell-execute the folder itself (FileName = library.GetDirectory(), UseShellExecute = true) instead of hardcoding explorer.exe. Under Wine (Linux/Mac users) that lets the prefix route folders to the host file manager, and it drops the manual quoting. Please also wrap the launch in try/catch with Plugin.Log.Error like UrlActions.OpenInBrowser does, since a Process.Start failure here would throw inside the draw loop.

src/Aetherphone/Core/Photos/PhotoLibrary.cs
```diff
@@ -10,6 +10,8 @@ public PhotoLibrary(DirectoryInfo configDirectory)
         Directory.CreateDirectory(directory);
     }
 
+    public string GetDirectory() => directory;
```
XeldarAlz: Non-blocking: prefer a get-only property here, e.g. public string DirectoryPath => directory;

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -292,6 +294,19 @@ private void DrawAlbumsGrid(Rect body)
         }
     }
 
+    private void DrawOpenFolder(Rect rect)
+    {
+        if (!ComposeFab.Draw(rect, "##openFolderFab", Accent, FontAwesomeIcon.Folder.ToIconString(),
+                             Loc.T(L.Photos.OpenFolder), "photos.openFolder")) return;
```
XeldarAlz: Nit: braces around the if body per codebase style, and the blank line below has trailing whitespace.

src/Aetherphone/Localization/en.json
```diff
@@ -1721,6 +1721,7 @@
   "photos.recents": "Recents",
   "photos.today": "Today",
   "photos.yesterday": "Yesterday",
+  "photos.openFolder": "Open folder",
```
XeldarAlz, line 1724: Non-blocking: the source in L.cs says "Open Folder", en.json says "Open folder". en.json is the translator reference, please keep it identical to L.cs.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -292,6 +295,15 @@ private void DrawAlbumsGrid(Rect body)
         }
     }
 
+    private void DrawOpenFolder(Rect rect)
+    {
+        if (!ComposeFab.Draw(rect, "##openFolderFab", Accent, FontAwesomeIcon.Folder.ToIconString(),
```
XeldarAlz: Blocking: the condition is inverted now. ComposeFab.Draw returns true only on click, so with the ! this opens the folder on every frame the button is NOT clicked, spawning an Explorer window per frame while the Library tab is visible. Drop the ! so it opens only on click.

src/Aetherphone/Windows/UrlActions.cs
```diff
@@ -24,6 +24,23 @@ public static void OpenInBrowser(string url, Action<Exception>? onError = null)
...
+            Process.Start(new ProcessStartInfo
+            {
+                FileName = path,
+                UseShellExecute = true
+            });
+
+        }
+        catch (Exception e)
+        {
+            Plugin.Log.Error($"Failed to open the folder: {path}", e);
```
XeldarAlz: Non-blocking: this overload treats e as a format value, so the stack trace is lost. Use the exception-first overload like PhotoLibrary does: Plugin.Log.Error(e, $"Failed to open the folder: {path}"). Nits: exception instead of e per codebase naming, and there is a stray blank line in the try block.

XeldarAlz (approved):
Verified at 76844513: Release build passes (0 errors, only the pre-existing AngleSharp NU1902 warning), tests 67/67 pass. All blocking items resolved: translations filled, shell-execute via the shared UrlActions.OpenFolder, click check un-inverted. Only leftover is the stray blank line inside the try block in UrlActions.OpenFolder, fine to leave. Thanks for the quick turnarounds!

PR #55, opened 2026-07-24
Fix Music app favorite star not working on station rows

Hubkaw:
## What

Change star click detection to UiInteract and check for it before play check is done.

## Why

UiInteract.Click was claiming the click before star's logic happened after https://github.com/XeldarAlz/FFXIV-Aetherphone/pull/42 which made the star completely ignored on station list view

## Test

1. open phone
2. navigate to Music app
3. navigate to any radio category
4. click on the star next to any radio station

## Checklist

### DIFF
--- a/src/Aetherphone/Apps/Music/MusicApp.Home.cs
+++ b/src/Aetherphone/Apps/Music/MusicApp.Home.cs
@@ -432,6 +432,7 @@ internal sealed partial class MusicApp
         }
 
         var favoriteStar = false;
+        var overStar = false;
         var isFavoriteStation = favoriteRadioStations.Contains(station);
 
         if (isFavoriteStation || hovered)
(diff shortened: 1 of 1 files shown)


XeldarAlz (approved):
Verified in code: the row's UiInteract.Click at MusicApp.Home.cs:453 overwrote the star's press-frame claim (last claim wins, UiInteract.cs:62-68), so the star could never activate. Skipping the row's click while over the star is the right fix, and the hover rect matches IconButton's hit rect exactly (AppSkin.cs:259-260).

Non-blocking: the IconButton call could reuse the new starY, and 14f * scale now appears three times; I'll hoist locals post-merge. Also spotted the same claim steal in MapsApp.cs:308-309, will fix that separately.

Thanks for the find and the clean fix!

PR #57, opened 2026-07-24
feat(photos): Add custom albums

Syrilai:
## What

This adds custom albums to the photos app.

## Why

We like to keep it tidy and clean, and I was asked if I could make it.

## How to test

1. Open the phone via any preferred method
2. Open the photos app
3. Navigate to Albums tab
4. Click the + button in the bottom right
5. Name the album, and click new album
6. Enter an album name, click create album
7. Enter the new album
8. Click "Add Photos", and select the photos to add
9. Head back out, right click the album, and rename

## Checklist


src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -347,6 +529,310 @@ private bool DrawAlbumCard(ImDrawListPtr drawList, Rect rect, string title, int
...
+        ImGui.SetCursorScreenPos(body.Min);
+        using (ImRaii.PushStyle(ImGuiStyleVar.WindowPadding, Vector2.Zero))
+        using (var child = ImRaii.Child("##albumPicker", body.Size, false,
+                   DragScrollHost.ScrollFlags(ImGuiWindowFlags.NoBackground)))
+        {
+            if (!child)
+                return;
+
+            var surface = DragScrollHost.Begin(gridKey);
+            surface.JumpToTop();
```
XeldarAlz: **Blocking: the picker can never be scrolled.**

`JumpToTop()` is `ImGui.SetScrollY(0f)` plus `region.Reset()`, and this call is unconditional, so it fires every frame. The grid is pinned to the top permanently and you can only ever select photos in the first viewport. Anyone with more than ~12 photos cannot reach the rest.

`DrawCustomAlbumGrid` gets this right at line 777. Same guard needed here:

```csharp
if (resetScroll)
{
    surface.JumpToTop();
    resetScroll = false;
}
```

(or a dedicated `resetPickerScroll` flag set when the picker opens, since `resetScroll` is already shared with the album grid).

XeldarAlz (commented):
Nice feature, and the persistence + prune model is the right shape. Details are inline.

Verified: read `PhotosApp.cs` and `PhotosApp.Grid.cs` in full on the branch, plus `DragScrollHost.JumpToTop`, `DropdownMenu`, `ComposeFab`, `ResolveTaken`, and the 9 locale JSONs against `L.cs`.

**3 blocking bugs**, two of which are reachable in normal use and contradict the "Verified in-game" checkbox:

1. The Add Photos picker cannot be scrolled at all (`PhotosApp.Grid.cs:693`).
2. New Album is unreachable with an empty library (`PhotosApp.Grid.cs:57`).
3. A full sort with filesystem stats runs every frame inside a custom album (`PhotosApp.Grid.cs:124`).

The rest are our global rules (no `what` comments, no LINQ in hot paths, no abbreviated names, braces on single-line bodies, YAGNI, DRY) plus some non-blocking robustness and UX notes.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -52,6 +54,7 @@ private void DrawRoot(Rect area)
         }
 
         DrawAlbumsGrid(body);
+        DrawNewAlbumFab(body);
```
XeldarAlz, line 58: **Blocking: New Album is unreachable with an empty library.**

`DrawRoot` returns at the `entries.Length == 0` guard above (line 50-54) before it ever reaches `DrawAlbumsGrid` / `DrawNewAlbumFab`. Two consequences:

- A brand new user with zero photos cannot create an album at all, the FAB does not exist.
- If a user deletes every photo, their existing custom albums disappear from the Albums tab entirely (the albums still exist in config, they just cannot be seen or reached).

The empty-state guard should apply to the Library segment only. The Albums segment needs to render whenever `customAlbums.Count > 0`, and the FAB unconditionally.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -88,6 +96,172 @@ private void DrawAlbum(Rect area, int key)
...
+        DrawNavBar(area, album.Name, back);
+        if (ui.HeaderAction(area, Loc.T(L.Photos.AddPhotos), entries.Length > 0))
+        {
+            pickerAlbumKey = key;
+            showAlbumPicker = true;
+            pickerSelection.Clear();
+        }
+
+        var body = new Rect(new Vector2(area.Min.X, area.Min.Y + AppHeader.Height * scale), area.Max);
+        var paths = SortedCustomAlbumPaths(key);
```
XeldarAlz: **Blocking: full sort with disk I/O on every frame.**

This is in the draw path, so per frame it does `unordered.ToArray()` (one array allocation) and then `Array.Sort` with a comparator that calls `ResolveTaken` twice per comparison. `ResolveTaken` does a `Path.GetFileNameWithoutExtension` (string allocation) and, for anything not matching the `AEP_` naming convention, falls through to `File.GetLastWriteTime`, which is a filesystem stat.

A 200 photo album is roughly 1500 comparisons, so up to 6000 stat calls and 6000 string allocations per frame at 60fps. Photos imported from outside the camera all take the `File.GetLastWriteTime` path.

Sort once and cache it. `BuildCustomAlbums` already runs on every mutation and on `Refresh`, so the sorted array can be built there and just read here.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -347,6 +529,310 @@ private bool DrawAlbumCard(ImDrawListPtr drawList, Rect rect, string title, int
...
+    private void DrawAlbumPicker(Rect area, int key)
+    {
+        var scale = ImGuiHelpers.GlobalScale;
+        if (!TryFindCustomAlbum(key, out var album))
+        {
+            showAlbumPicker = false;
+            return;
+        }
+
+        DrawNavBar(area, string.Format(Loc.Culture, "{0} — {1}", album.Name, Loc.T(L.Photos.AddPhotos)), () =>
```
XeldarAlz: Em dash. We have a hard zero em dash rule across the repo (copy, locales, comments), and this is the only one in `src/**/*.cs` apart from one pre-existing string in `L.cs`.

Also, hardcoding a separator format here means the title structure is not localizable. Prefer a dedicated `LocString` for the picker title, or just use the album name alone since "Add Photos" is already the header action label.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -347,6 +529,310 @@ private bool DrawAlbumCard(ImDrawListPtr drawList, Rect rect, string title, int
...
+
+        var textTop = coverCoverMax.Y + 7f * scale;
+        var name = Typography.FitText(album.Name, rect.Width - 4f * scale, TextStyles.SubheadlineEmphasized);
+        Typography.Draw(drawList, new Vector2(rect.Min.X + 2f * scale, textTop), name, ui.TitleInk,
+            TextStyles.SubheadlineEmphasized);
+        var countLabel = Loc.Plural(L.Photos.Count, album.Count);
+        Typography.Draw(drawList, new Vector2(rect.Min.X + 2f * scale, textTop + 19f * scale), countLabel, ui.MutedInk,
+            TextStyles.Footnote);
+
+        // Right-click to open context menu
```
XeldarAlz: `what` comments. This one, plus `// Context menu` (597), `// Tap to open` (605) and `// Right-click to remove from album` (816), all restate what the next line already says.

Self-documenting code is a global rule, and the PR checklist claims "no `what` comments". Please drop all four. If a block needs a label, that is usually the signal to extract it into a named method instead.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -88,6 +96,172 @@ private void DrawAlbum(Rect area, int key)
 
         DrawPhotoGrid(body, start, count);
     }
+    
+    private void DrawCustomAlbumView(Rect area, int key)
+    {
+        if (showAlbumPicker && pickerAlbumKey == key)
```
XeldarAlz: The picker is a boolean sub-mode rather than a router route, which goes against the ViewRouter convention (all sub-screen nav goes through the router).

Concrete bug it causes: the nav bar back here is a custom lambda that clears `showAlbumPicker`, but the phone's own back gesture calls `router.Pop()` directly. Using that pops you out of the album entirely while `showAlbumPicker` stays `true`. It is only cleared in `OnOpened`, so re-entering that same album drops you straight back into the picker.

A `PhotoRoute.AlbumPicker` route would make back work uniformly and let you delete `showAlbumPicker` / `pickerAlbumKey`.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -88,6 +96,172 @@ private void DrawAlbum(Rect area, int key)
...
+        }
+
+        DrawCustomAlbumGrid(body, paths, key);
+    }
+
+    private void DrawCreateAlbumPage(Rect area)
+    {
+        var scale = ImGuiHelpers.GlobalScale;
+
+        Action cancel = () =>
```
XeldarAlz: This closure is allocated every frame the page is drawn, and it captures `this`. Make it a private method and pass the method group, or hoist it to a cached field.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -88,6 +96,172 @@ private void DrawAlbum(Rect area, int key)
...
+            if (!canCreate && newAlbumDraft.Trim().Length > 0)
+            {
+                ImGui.Dummy(new Vector2(0f, 10f * scale));
+                using (ImRaii.PushColor(ImGuiCol.Text, ui.Palette.Accent))
+                    Typography.Wrapped(Loc.T(L.Photos.AlbumExists));
+            }
+        }
+    }
+
+    private void DrawRenameAlbumSheet(Rect body, Action cancel)
```
XeldarAlz: DRY: this and `DrawCreateAlbumSheet` (line 154) are about 50 near-identical lines. Same squircle field construction, same invisible `FrameBg` input, same dummy-over-origin trick, same four-color accent button push, differing only in the draft field, the validation predicate and the commit call.

Collapse to one method taking the label, the `ref` draft, a validity flag and a commit action. That also fixes the divergence flagged below, where only one of the two sheets shows the "name already exists" hint.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -88,6 +96,172 @@ private void DrawAlbum(Rect area, int key)
...
+                       .Push(ImGuiCol.Text, ui.TitleInk))
+            {
+                ImGui.InputText("##renameAlbumName", ref renameAlbumDraft, 64);
+            }
+
+            ImGui.SetCursorScreenPos(origin);
+            ImGui.Dummy(new Vector2(width, height));
+
+            ImGui.Dummy(new Vector2(0f, 18f * scale));
+            var canRename = renameAlbumDraft.Trim().Length > 0;
```
XeldarAlz: Two gaps in the rename validation:

1. **Case-only renames are impossible.** The `OrdinalIgnoreCase` equality check below rejects "trip" to "Trip", so a user cannot fix the capitalization of their own album.
2. **No feedback when the name is taken.** `DrawCreateAlbumSheet` shows the `AlbumExists` hint, this one does not. The button just sits dimmed with no explanation of why.

Both resolve naturally if the two sheets are unified.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -347,6 +529,310 @@ private bool DrawAlbumCard(ImDrawListPtr drawList, Rect rect, string title, int
...
+        // Tap to open
+        if (UiInteract.Click(rect.Min, rect.Max, hovered))
+        {
+            OpenAlbum(album.Key);
+        }
+    }
+    
+    private void DrawCustomAlbumContextMenu(int key, Rect screen)
+    {
+        DropdownMenu.Item[] items =
```
XeldarAlz: This array plus both `Loc.T` calls are allocated every frame the menu is open. `DropdownMenu.Draw` takes a `ReadOnlySpan<Item>`, so a cached field or a `stackalloc`-friendly local works here.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -347,6 +529,310 @@ private bool DrawAlbumCard(ImDrawListPtr drawList, Rect rect, string title, int
...
+            surface.JumpToTop();
+
+            var origin = ImGui.GetCursorScreenPos();
+            var side = 2f * scale;
+            var gap = 3f * scale;
+            var avail = ScrollLayout.StableContentWidth();
+            var cell = (avail - side * 2f - gap * (Columns - 1)) / Columns;
+            var drawList = ImGui.GetWindowDrawList();
+
+            var inAlbum = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
XeldarAlz: Per-frame allocation in the picker draw path: a `HashSet<string>` built fresh every frame, plus a LINQ `FirstOrDefault` with a capturing closure to find the album name. Build the membership set once when the picker opens.

Separately, this grid has no viewport culling. `DrawCustomAlbumGrid` (line 780ish) and `DrawPhotoGrid` both cull against `scrollY` and `viewHeight`. With a large library the picker draws and hit-tests every thumbnail in the library every frame.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -347,6 +529,310 @@ private bool DrawAlbumCard(ImDrawListPtr drawList, Rect rect, string title, int
...
+                    AppSkin.Icon(drawList, checkCenter, FontAwesomeIcon.Check.ToIconString(),
+                        new Vector4(0.4f, 0.8f, 0.4f, 0.8f), 0.8f);
+                }
+                else if (isSelected)
+                {
+                    drawList.AddRectFilled(min, max, ImGui.GetColorU32(Palette.WithAlpha(ui.Accent, 0.35f)), 7f * scale);
+                    var radius = 11f * scale;
+                    var badgeCenter = new Vector2(max.X - radius - 6f * scale, min.Y + radius + 6f * scale);
+                    drawList.AddCircleFilled(badgeCenter, radius, ImGui.GetColorU32(ui.Accent), 20);
+                    var order = pickerSelection.IndexOf(path) + 1;
```
XeldarAlz: `List.IndexOf` inside the per-item loop makes the selection badge numbering O(n^2) in the number of selected photos. Since you already know the item is selected, track the ordinal in a `Dictionary<string, int>` alongside `pickerSelection`, or store the selection as an index list.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -347,6 +529,310 @@ private bool DrawAlbumCard(ImDrawListPtr drawList, Rect rect, string title, int
...
+                {
+                    ImGui.SetMouseCursor(ImGuiMouseCursor.Hand);
+                }
+
+                if (UiInteract.Click(min, max, hovered))
+                {
+                    OpenViewerFromPaths(paths, index);
+                }
+
+                // Right-click to remove from album
```
XeldarAlz: Removing a photo from an album is right-click only with no visual affordance at all, one screen after the album card itself uses a proper `DropdownMenu` for the same class of action.

Users will not find this. Suggest the same `DropdownMenu` treatment for consistency (a Remove from Album row), or at minimum a hover affordance.

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -1,5 +1,6 @@
 using System.Collections.Concurrent;
 using System.Globalization;
+using Aetherphone;
```
XeldarAlz: Redundant. This file's namespace is `Aetherphone.Apps.Photos`, so `Aetherphone` is already an ancestor namespace and in scope.

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -43,6 +45,18 @@ internal sealed partial class PhotosApp : IPhoneApp
     private readonly List<MonthAlbum> albums = new();
     private readonly List<GridBand> bands = new();
     private readonly string[] segmentLabels = new string[2];
+    private readonly List<CustomAlbum> customAlbums = new();
+    private readonly List<string> customAlbumOrder = new();
+    private readonly Dictionary<string, List<string>> customAlbumPhotos = new(StringComparer.OrdinalIgnoreCase);
+    private string[] customAlbumFlat = Array.Empty<string>();
```
XeldarAlz: `customAlbumFlat` is written on every rebuild (`flat.ToArray()` at line 255, plus the `List` and all the `AddRange` copying that feeds it) and never read anywhere in the codebase.

`CustomAlbum.Start` exists only to index into it, so that is dead too. Both should go, along with the `flat` list in `BuildCustomAlbums`.

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -43,6 +45,18 @@ internal sealed partial class PhotosApp : IPhoneApp
...
+    private readonly DropdownMenu albumMenu = new();
+    private string newAlbumDraft = string.Empty;
+    private bool renaming;
```
XeldarAlz: `renaming` makes `PhotoRoute.CreateAlbum` mean two different screens depending on hidden state, so the route no longer describes what is on screen. A separate `PhotoRoute.RenameAlbum` would carry the album key on the view itself and let both `renaming` and `renameAlbumKey` disappear.

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -154,6 +182,36 @@ private void Refresh()
         Array.Sort(built, static (left, right) => right.Taken.CompareTo(left.Taken));
         entries = built;
         BuildAlbums();
+        PruneCustomAlbumPaths(pathSet);
+        BuildCustomAlbums();
+        SaveCustomAlbums();
```
XeldarAlz: `Refresh()` saves unconditionally, and `SaveNow()` is `SavePluginConfig`, a synchronous disk write.

`Refresh()` runs on every `OnOpened` and after every photo delete, so opening the Photos app writes the whole plugin config to disk every single time, including for the majority of users who have no custom albums at all. Nothing has usually changed.

Have `PruneCustomAlbumPaths` report whether it actually removed anything and only save when it did (the mutator methods already save on their own).

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -154,6 +182,36 @@ private void Refresh()
...
+    }
+    
+    private void PruneCustomAlbumPaths(HashSet<string> validPaths)
+    {
+        foreach (var name in customAlbumOrder)
+        {
+            if (!customAlbumPhotos.TryGetValue(name, out var photos))
+                continue;
+            photos.RemoveAll(p => !validPaths.Contains(p));
```
XeldarAlz: `RemoveAll` with a capturing lambda allocates a closure and a delegate per album per refresh. A reverse `for` loop with `RemoveAt` avoids both.

Also worth noting: this method correctly uses `TryGetValue` and skips names with no map entry, but `BuildCustomAlbums` at line 248 indexes the same dictionary directly. Those two should agree, see the comment there.

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -176,8 +234,127 @@ private void BuildAlbums()
...
-            albums.Add(new MonthAlbum(key, new DateTime(taken.Year, taken.Month, 1), start, index - start));
+            albums.Add(new MonthAlbum(key, start, index - start, new DateTime(taken.Year, taken.Month, 1)));
+        }
+    }
+    
+    private void BuildCustomAlbums()
+    {
+        customAlbums.Clear();
+        var flat = new List<string>();
+        for (var i = 0; i < customAlbumOrder.Count; i++)
```
XeldarAlz: `i` is an abbreviated name. Global rule is explicit loop variable names, `index` here (and at line 260). Same applies to the lambda parameters introduced across this file: `c`, `p`, `a`, `b`, `ta`, `tb`, `kvp`.

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -176,8 +234,127 @@ private void BuildAlbums()
...
+        {
+            var name = customAlbumOrder[i];
+            var photos = customAlbumPhotos[name];
```
XeldarAlz: Unguarded indexer, throws `KeyNotFoundException` if `customAlbumOrder` ever contains a name with no entry in `customAlbumPhotos`.

The in-app flows keep the two in sync today, but both are deserialized independently from user-editable config JSON, so a hand-edited or partially written config crashes the app on open. `RenameCustomAlbumInternal` at line 308 has the same exposure.

`PruneCustomAlbumPaths` already defends against exactly this with `TryGetValue`. Cheap to be consistent.

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -176,8 +234,127 @@ private void BuildAlbums()
...
+            var start = flat.Count;
+            flat.AddRange(photos);
+            // Negative keys avoid any collision with MonthAlbum keys (year*100+month, always positive).
```
XeldarAlz: Deriving the key from the list position means deleting an album silently renumbers every album after it.

`pickerAlbumKey`, `renameAlbumKey` and the view pushed into the router all hold these keys across frames, so this is a latent identity bug even though I could not reach it from the current UI flows. Album identity should be the name (already unique, already the dictionary key) or a stable id allocated at creation.

The `+ 2` offset also needs explaining or removing, `-1` is unused and the `key < 0` branch accepts it fine.

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -176,8 +234,127 @@ private void BuildAlbums()
...
+            }
+        }
+        result = default;
+        return false;
+    }
+    
+    private void CreateCustomAlbumInternal(string name)
+    {
+        name = name.Trim();
+        if (name.Length == 0)
```
XeldarAlz, line 310: Missing braces on a single-line `if` body. Global rule is braces always, including single-line bodies.

There are roughly 20 of these across the new methods in this file and in `PhotosApp.Grid.cs`, worth a pass over the whole diff rather than fixing them one by one.

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -176,8 +234,127 @@ private void BuildAlbums()
...
+            return;
+        customAlbumOrder.Add(name);
+        customAlbumPhotos[name] = new List<string>();
+        BuildCustomAlbums();
+        SaveCustomAlbums();
+    }
+
+    private void DeleteCustomAlbumInternal(int key)
+    {
+        var found = customAlbums.FirstOrDefault(c => c.Key == key);
```
XeldarAlz: LINQ in a hot path. `FirstOrDefault` with a `key`-capturing lambda appears 8 times across this file and `PhotosApp.Grid.cs`, and several of those run per frame (`DrawAlbumPickerGrid`, `DrawCustomAlbumContextMenu`). Each call allocates a closure plus a delegate.

`TryFindCustomAlbum` right above is already the allocation-free version. Use it everywhere.

Same rule applies to `Enumerable.Contains(..., StringComparer)` at lines 277 and in `AddPhotosToCustomAlbum`.

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -410,19 +595,49 @@ public PhotoEntry(string path, DateTime taken)
...
 
-    private readonly struct MonthAlbum
+    private readonly struct AlbumInfo
+    {
+        public int Key { get; init; }
+        public int Start { get; init; }
+        public int Count { get; init; }
+    }
+
+    private interface IAlbum
```
XeldarAlz: YAGNI: `IAlbum` is declared and never consumed. Nothing takes an `IAlbum`, no method is generic over it, there is no shared call site. It is pure speculative structure.

It is also actively risky: `MonthAlbum` and `CustomAlbum` are structs, so the first thing that ever does use this interface boxes them on every access.

Same for `AlbumInfo`, which wraps three ints behind `init` properties and adds a level of indirection to every `Key` / `Start` / `Count` read for no gain. Please drop both and give `CustomAlbum` plain readonly fields like `MonthAlbum` had.

src/Aetherphone/Apps/Photos/PhotosApp.cs
```diff
@@ -410,19 +595,49 @@ public PhotoEntry(string path, DateTime taken)
...
+    {
+        int Key { get; }
+        int Start { get; }
+        int Count { get; }
+    }
+
+    private readonly struct MonthAlbum : IAlbum
```
XeldarAlz: This rewrites a working struct that the feature does not need to touch: readonly fields became `init` properties behind `AlbumInfo`, and the constructor parameter order changed from `(key, month, start, count)` to `(key, start, count, month)`.

Unrelated churn on existing code makes the diff harder to review and risks a silent argument-order mistake at the call site. Please revert `MonthAlbum` to its original form and let `CustomAlbum` stand on its own.

src/Aetherphone/Localization/en.json
```diff
@@ -1745,6 +1745,20 @@
   "photos.today": "Today",
   "photos.yesterday": "Yesterday",
   "photos.openFolder": "Open folder",
+  "photos.createAlbum": "New Album",
+  "photos.albumName": "Album name",
+  "photos.albumNamePlaceholder": "My Album",
```
XeldarAlz, line 1768: `photos.albumNamePlaceholder` and `photos.noAlbums` (line 1758) are added to all 9 locale files but never declared in `L.cs`, so nothing can ever resolve them. `L.cs` is the runtime source of truth, a key that is not there is unreachable.

Either add the `LocString` entries and use them (a placeholder in the empty name field and a proper albums empty state would both be improvements), or drop the 18 dead lines across the 9 files.

src/Aetherphone/Localization/en.json
```diff
@@ -1745,6 +1745,20 @@
...
+  "photos.createAlbumButton": "Create Album",
+  "photos.addPhotos": "Add Photos",
+  "photos.removeFromAlbum": "Remove from Album",
+  "photos.deleteAlbum": "Delete Album",
+  "photos.deleteAlbumConfirm": "Delete {0}?",
+  "photos.deleteAlbumBody": "Photos in the album won't be deleted.",
+  "photos.renameAlbum": "Rename Album",
```
XeldarAlz, line 1775: Mismatch with `L.cs`, where `Photos.Rename` defaults to `"Rename"` while this says `"Rename Album"`. The `L.cs` default is what English users actually see, so this line is misleading reference text.

Worth deciding which you want, since the same string is used both as the context menu row and as the confirm button on the rename sheet. "Rename" reads better in the menu, the translations here all went with the longer "Rename Album" form.

PR #61, opened 2026-07-25
fix(velvet): treat an unknown race as unavailable instead of allowed

Valiice:
## What

Velvet is meant to be unavailable on Lalafell characters. The check behind that can't tell your race until a character is loaded, and when it can't tell it answers "not a Lalafell". So at the login screen a Lalafell gets in. This gives the check a third answer, "don't know yet", and treats that as unavailable.

## Why

Came out of this report: https://discord.com/channels/1527213298087493732/1530436723899633855

`cachedLalafell` is a plain bool defaulting to false and both early returns in `IsLalafellCharacter()` hand it back, so "couldn't read the race" and "confirmed not a Lalafell" were the same value. With no character loaded it is always false.

Two things consume it:

- the unavailable screen from 048db794, so the app opens at the title screen
- the heartbeat, which sends `&lalafell=false`. `sinceHeartbeat` starts already expired, so the first drawn frame sends it before any read could have succeeded.

The 403 doesn't cover this on its own. `EnsureMe` returns early once `me is not null`, and `me` and `accessBlocked` only reset on an account change, so the 403 can only land on the first profile load after signing in. After one 200 the client side check is the only thing left. The gate at the top of `Draw` also runs before the profile load it depends on has finished.

## How to test

1. Sign in, then sit at the title screen with Velvet open (reloading the plugin there is quickest). On master the app opens normally. With this it shows the Unavailable screen.
2. Log in on a Lalafell with Velvet open. Unavailable, same as master.
3. Log in on anything else. Velvet works as before, though it can take up to a second after the character loads.

Step 1 is the one that matters.

## Extra

`RaceWatch` (new, `Core/Game/`) holds the race id it read, or null if it doesn't have one yet, and `IsLalafell` comes off that. Race id 0 is ignored since it isn't a real race, and a failed read never overwrites an answer it already has. It's a volatile byte because `bool?` is two fields and can be caught half written from the draw thread.

`VelvetShell` reads it on a `FrameworkTicker` (1s, gated on the app being installed) rather than during `Draw`, so it no longer depends on whether `Draw` counts as the framework thread. It forgets the answer only on a real logout, so a loading screen or a zone change keeps the last known race.

The two call sites want opposite defaults, which is the reason for the tri-state: the gate blocks on unknown, the heartbeat sends nothing on unknown instead of a guess.

11 tests in `RaceWatchTests`.

Couple of open questions:

- Does the 403 depend on the `lalafell` flag we send? If it does, some stored tags are probably wrong, since master can send false for a character it never read.
- Skipping the beat also skips region and last seen, so presence stops refreshing at the title screen. Beating without the flag would avoid that, but `&lalafell=` is always appended, so it needs a server side change.

## Checklist

### DIFF
--- a/src/Aetherphone/Apps/Velvet/VelvetShell.cs
+++ b/src/Aetherphone/Apps/Velvet/VelvetShell.cs
@@ -22,7 +22,6 @@ using Aetherphone.Core.Theme;
 using Aetherphone.Core.Wallpapers;
 using Aetherphone.Windows.Components;
 using Dalamud.Bindings.ImGui;
-using Dalamud.Game.ClientState.Objects.Enums;
 using Dalamud.Interface;
 using Dalamud.Interface.Utility;
 
@@ -31,6 +30,7 @@ namespace Aetherphone.Apps.Velvet;
 internal sealed partial class VelvetShell : IPhoneApp
 {
     private const float HeartbeatSeconds = 45f;
+    private const long RaceProbeIntervalMilliseconds = 1000;
 
     private readonly VelvetStore store;
     private readonly StoryPresenter stories;
@@ -66,11 +66,13 @@ internal sealed partial class VelvetShell : IPhoneApp
     private readonly RouterDraw<VelvetView> drawView;
     private readonly Action back;
 
+    private readonly RaceWatch race = new();
+    private readonly FrameworkTicker raceTicker;
+
     private PhoneTheme theme = PhoneTheme.Default;
     private INavigator navigation = null!;
     private VelvetPage activeTab = VelvetPage.Discover;
     private float sinceHeartbeat = HeartbeatSeconds;
-    private bool cachedLalafell;
 
     public VelvetShell(AethernetSession session, AethernetApi net, LodestoneService lodestone,
         Configuration configuration, PhotoLibrary library, HttpService http, RemoteImageCache images,
@@ -109,6 +111,8 @@ internal sealed partial class VelvetShell : IPhoneApp
         drawView = DrawView;
         back = () => router.Pop();
         threadView = new ThreadView(this);
+        raceTicker = new FrameworkTicker(Plugin.Framework, RaceProbeIntervalMilliseconds, ProbeRace,
+            installer.Gate(Id));
     }
 
     public string Id => "velvet";
@@ -206,7 +210,7 @@ internal sealed partial class VelvetShell : IPhoneApp
             return;
         }
 
-        if (IsLalafellCharacter() || store.AccessBlocked)
+        if (Unavailable)
         {
             TourHolds.Hold(Id);
             store.EnsureMe();
@@ -263,6 +267,7 @@ internal sealed partial class VelvetShell : IPhoneApp
 
     public void Dispose()
     {
+        raceTicker.Dispose();
         threadView.Dispose();
         stories.Dispose();
         store.Dispose();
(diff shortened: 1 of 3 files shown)


XeldarAlz (changes requested):
Superseded by the inline review below, ignore this one. Posted twice by mistake.

src/Aetherphone/Core/Game/RaceWatch.cs
```diff
@@ -0,0 +1,35 @@
...
+internal sealed class RaceWatch
+{
+    internal const int RaceIndex = (int)CustomizeIndex.Race;
+    internal const byte LalafellRaceId = 3;
+
+    private const byte UnknownRaceId = 0;
+
+    private volatile byte race = UnknownRaceId;
+
+    public byte? Race => race == UnknownRaceId ? null : race;
```
XeldarAlz, line 14: Two volatile loads. If `Forget()` lands between them this returns a non-null `(byte?)0`, so `IsLalafell` answers `false` for an unknown race. Verified against real Dalamud: 2.2M bad reads in 3s under contention.

```suggestion
    public byte? Race
    {
        get
        {
            var observed = race;
            return observed == UnknownRaceId ? null : observed;
        }
    }
```

src/Aetherphone/Apps/Velvet/VelvetShell.cs
```diff
@@ -206,7 +210,7 @@ public void Draw(in PhoneContext context)
             return;
         }
 
-        if (IsLalafellCharacter() || store.AccessBlocked)
+        if (Unavailable)
```
XeldarAlz, line 213: `velvet.unavailableBody` reads "not available on Lalafell characters". Unknown lands here too now, so logging out to character select tells a Highlander they are a Lalafell. Needs its own empty state: new `LocString` plus all nine JSONs.

src/Aetherphone/Apps/Velvet/VelvetShell.cs
```diff
@@ -31,6 +30,7 @@ namespace Aetherphone.Apps.Velvet;
 internal sealed partial class VelvetShell : IPhoneApp
 {
     private const float HeartbeatSeconds = 45f;
+    private const long RaceProbeIntervalMilliseconds = 1000;
```
XeldarAlz, line 33: This one second floor means every login with Velvet open shows that screen briefly, not just the title screen. Skip the rate limit while the race is unknown.

src/Aetherphone/Apps/Velvet/VelvetShell.cs
```diff
@@ -272,34 +277,38 @@ public void Dispose()
         configuration.VelvetAcknowledgedGate &&
         configuration.VelvetAcknowledgedGateVersion >= Configuration.VelvetGateVersion;
 
+    private bool Unavailable => race.IsLalafell is not false || store.AccessBlocked;
```
XeldarAlz, line 280: Nit: `is not false` is a double negative. `IsLalafellOrUnknown` on `RaceWatch` would read better here.

XeldarAlz (changes requested):
Builds clean in Release. Comments inline.

Branch does not merge: based on `8bf1669f`, 222 commits behind. Watch the `ImGuiHelpers.GlobalScale` to `UiScale.Current` migration on rebase, CI fails on it.

Your open questions: presence stopping at the title screen is correct, leave it. The stored `lalafell` flag may have bad rows from master, I will check Aethernet.

PR #62, opened 2026-07-25
fix(controlcenter): removed modules no longer reappear after a restart

BluntEXE:
## Summary

Removed Control Center modules reappeared on every plugin restart, forcing a manual re-removal each time.

`ControlLayoutService.Load()` derived which modules belong on the grid from what's currently placed, with no record of what the user explicitly removed. A module absent from a saved layout looked identical to a module that had never been added, so the auto-add fallback (meant for modules newly registered since the last save) silently re-added anything the user had removed, every time the plugin loaded. Tile position and span for kept modules were never affected - that path only touches `Items`.

## Root cause

```csharp
private void Load()
{
    // ...loads saved Items into slots...

    var all = registry.Modules;
    for (var index = 0; index < all.Count; index++)
    {
        if (placed.Add(all[index].Id))
        {
            slots.Add(ControlSlot.For(all[index], all[index].DefaultSpan));
        }
    }
    // ...
}
```

That fallback can't distinguish "genuinely new module" from "module the user removed." Both are simply absent from `placed`.

## Fix

`ControlLayout` gains a persisted `Enabled` list, and `ControlLayoutService` mirrors `HomeLayoutService`'s existing `Installed`/`LoadInstalled` pattern, which already solves this exact problem correctly for the home screen's app install/uninstall system:

- Trust the saved `Enabled` set directly when present - no re-deriving from `Items`.
- Seed it only once: every module, on a genuinely first run (`saved is null`); from the current `Items`, on a save that predates this field (existing users keep their current layout as the new baseline instead of losing it).
- `Add`/`Remove` update `Enabled` alongside `Items`.
- `Reset()` already nulls the whole `ControlPanel`, so it clears `Enabled` too - a full reset still restores everything.
- `LoadEnabled` unions `placed` into `enabled` right after the trust/seed branch, so `Items` stays a subset of `Enabled` even if the two ever drift (a hand-edited config, a future migration).

No new mechanism invented - same shape, same load order as the code already uses one directory over.

## Behavior change worth flagging

Dropping the auto-add fallback removes a capability, not just a bug: a Control Center module shipped in a future version no longer appears on the grid automatically. It lands in the "Add Controls" gallery instead, same as a first-run module does today, matching `HomeLayoutService`'s already-tested contract.

Control Center has no `MandatoryApps`-equivalent to force a module visible by default - Home does. If a future module needs to ship visible without user action, that's the first time this gap needs closing.

## Unit tests

Added `ControlLayoutServiceInstallTests.cs`, mirroring `HomeLayoutServiceInstallTests.cs`'s first four cases for this same persistence contract:

- `FreshInstall_EnablesEveryAvailableModule`
- `ModuleShippedAfterTheSaveWasWritten_IsNotEnabled`
- `Add_PutsTheModuleOnTheGridAndSurvivesAReload`
- `Remove_TakesTheModuleOffTheGridAndSurvivesAReload` - the actual restart-regression case pinned down deterministically

Needed two small seams to make `ControlLayoutService` fakeable, both additive, no behavior change:

- `IControlConfiguration` (new): `ControlPanel` + `Save()`, the two members the service actually calls on `Configuration`. `Configuration` picks it up alongside its existing `IHomeConfiguration`.
- `IControlRegistry` (new): `Modules` + `TryGet`, the two members the service actually calls on `ControlRegistry`. `ControlRegistry` now implements it; its constructor and module wiring are untouched.
- `ControlLayoutService`'s constructor takes the two interfaces instead of the concrete types. Every call site already passes a `ControlRegistry`/`Configuration`, so this is source-compatible everywhere it's constructed.

One more wrinkle worth flagging: `IControlModule.GalleryIcon` is typed as `FontAwesomeIcon` (Dalamud), unlike every other fakeable interface in this suite (`IPhoneApp.Glyph` is a plain `string` specifically to keep Dalamud out of the test project). Faking `IControlModule` needed `Aetherphone.Tests.csproj` to gain its own reference to the Dalamud assembly, resolved the same way `Aetherphone.csproj` resolves it, without pulling in `Dalamud.NET.Sdk`'s plugin-packaging targets (those don't apply to a test project and demand a plugin manifest if imported). Test-project build config only; no production code or packaging path touched.

## Test plan

- [x] `dotnet build -c Release`: 0 errors, no new warnings (pre-existing AngleSharp NU1902 advisory only)
- [x] `dotnet test`: 107/107 passing (103 existing + 4 new)
- [x] In-game: removed a module, reloaded the plugin (simulates a restart - `ControlLayoutService` only loads once per instantiation), confirmed the module stayed removed
- [x] In-game: confirmed a kept module's position/span survived the reload unaffected
- [x] In-game: re-added the removed module from the gallery, reloaded again, confirmed it now stays present
- [x] Full FFXIV/XIVLauncher restart: removed tiles confirmed still gone after a real restart, not just a plugin reload

No comments added.

### DIFF
--- a/src/Aetherphone.Tests/ControlLayoutServiceInstallTests.cs
+++ b/src/Aetherphone.Tests/ControlLayoutServiceInstallTests.cs
@@ -1,0 +1,130 @@
+using Aetherphone.Core.ControlCenter;
+using Dalamud.Interface;
+using Xunit;
+
+namespace Aetherphone.Tests;
+
+public sealed class ControlLayoutServiceInstallTests
+{
+    [Fact]
+    public void FreshInstall_EnablesEveryAvailableModule()
+    {
+        var modules = MakeModules();
+        var service = BuildService(modules, new FakeControlConfiguration());
+
+        for (var index = 0; index < modules.Count; index++)
+        {
+            Assert.True(SlotIndexOf(service, modules[index].Id) >= 0,
+                $"Module '{modules[index].Id}' should ship on the grid");
+        }
+    }
+
+    [Fact]
+    public void ModuleShippedAfterTheSaveWasWritten_IsNotEnabled()
+    {
+        var modules = MakeModules();
+        var configuration = SavedWith("a", "b");
+
+        var service = BuildService(modules, configuration);
+
+        Assert.Equal(-1, SlotIndexOf(service, "c"));
+    }
+
+    [Fact]
+    public void Add_PutsTheModuleOnTheGridAndSurvivesAReload()
+    {
+        var modules = MakeModules();
+        var configuration = SavedWith("a", "b");
+
+        var first = BuildService(modules, configuration);
+        first.Add(modules[2]);
+        Assert.True(SlotIndexOf(first, "c") >= 0);
+
+        var reloaded = BuildService(modules, configuration);
+
+        Assert.True(SlotIndexOf(reloaded, "c") >= 0, "A module the user added should still be on the grid after a reload");
+    }
+
+    [Fact]
+    public void Remove_TakesTheModuleOffTheGridAndSurvivesAReload()
+    {
+        var modules = MakeModules();
+        var configuration = SavedWith("a", "b", "c");
+
+        var first = BuildService(modules, configuration);
+        var slot = first.Slots[SlotIndexOf(first, "c")];
+        first.Remove(slot);
+        Assert.Equal(-1, SlotIndexOf(first, "c"));
+
+        var reloaded = BuildService(modules, configuration);
+
+        Assert.True(SlotIndexOf(reloaded, "c") < 0, "A module the user removed must stay removed across a restart");
+    }
+
+    private static int SlotIndexOf(ControlLayoutService service, string moduleId)
+    {
+        for (var index = 0; index < service.Slots.Count; index++)
+        {
+            if (service.Slots[index].Id == moduleId)
+            {
+                return index;
+            }
+        }
+
+        return -1;
+    }
+
+    private static ControlLayoutService BuildService(List<IControlModule> modules, FakeControlConfiguration configuration) =>
+        new(new FakeControlRegistry(modules), configuration);
+
+    private static FakeControlConfiguration SavedWith(params string[] moduleIds)
+    {
+        var layout = new ControlLayout();
+        for (var index = 0; index < moduleIds.Length; index++)
+        {
+            layout.Items.Add(new ControlItem { ModuleId = moduleIds[index], Span = "small" });
+        }
+
+        layout.Enabled.AddRange(moduleIds);
+        return new FakeControlConfiguration { ControlPanel = layout };
+    }
+
+    private static List<IControlModule> MakeModules() =>
+        new()
+        {
+            new FakeControlModule("a"),
+            new FakeControlModule("b"),
+            new FakeControlModule("c"),
+        };
+
+    private sealed class FakeControlModule : IControlModule
+    {
+        public FakeControlModule(string id) => Id = id;
+        public string Id { get; }
+        public string GalleryLabel => Id;
+        public FontAwesomeIcon GalleryIcon => FontAwesomeIcon.Circle;
+        public IReadOnlyList<ControlSpan> Sizes { get; } = new[] { ControlSpan.Small };
+        public ControlSpan DefaultSpan => ControlSpan.Small;
+        public void Draw(in ControlModuleContext context) { }
+    }
+
+    private sealed class FakeControlRegistry : IControlRegistry
+    {
+        private readonly Dictionary<string, IControlModule> byId;
+
+        public FakeControlRegistry(List<IControlModule> modules)
+        {
+            Modules = modules;
+            byId = modules.ToDictionary(module => module.Id);
+        }
+
+        public IReadOnlyList<IControlModule> Modules { get; }
+        public bool TryGet(string id, out IControlModule module) => byId.TryGetValue(id, out module!);
+    }
+
+    private sealed class FakeControlConfiguration : IControlConfiguration
+    {
+        public ControlLayout? ControlPanel { get; set; }
+        public void Save() { }
+    }
+}
--- a/src/Aetherphone/Core/ControlCenter/ControlLayoutService.cs
+++ b/src/Aetherphone/Core/ControlCenter/ControlLayoutService.cs
@@ -7,15 +7,15 @@ internal sealed class ControlLayoutService
     public const int Columns = 4;
     private const int SolverRows = 16;
 
-    private readonly ControlRegistry registry;
-    private readonly Configuration configuration;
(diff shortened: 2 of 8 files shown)


src/Aetherphone/Core/ControlCenter/ControlLayoutService.cs
```diff
@@ -177,19 +180,56 @@ private void Load()
...
+        LoadEnabled(saved, placed);
+
+        placementsDirty = true;
+        if (saved is null)
+        {
+            Save();
+        }
+    }
+
+    private void LoadEnabled(ControlLayout? saved, HashSet<string> placed)
```
XeldarAlz, line 192: Non-blocking, but requested: this persistence contract ships untested, while `HomeLayoutServiceInstallTests.cs` already covers the same four cases for Home.

Cheap to mirror. `IControlModule` is 6 members, so a fake is ~8 lines, and `Configuration` already implements `IHomeConfiguration` (`Configuration.cs:25`), so an `IControlConfiguration` is one more entry on that list. Only real blocker is the concrete `ControlRegistry` in the ctor.

The case your test plan lists as uncovered (a real restart) is precisely what this would pin down deterministically.

src/Aetherphone/Core/ControlCenter/ControlLayoutService.cs
```diff
@@ -177,19 +180,56 @@ private void Load()
...
+            {
+                enabled.Add(stored[index]);
+            }
+        }
+        else
+        {
+            SeedEnabled(saved);
+        }
+
         var all = registry.Modules;
```
XeldarAlz, line 209: Optional hardening: `Items ⊆ Enabled` is not enforced.

A saved `Items` entry missing from `Enabled` still loads a slot in `Load()` (that loop never consults `enabled`), then re-saves forever without it, and `Remove` no-ops on the set. Not reachable through the current API, and `HomeLayoutService` has the identical gap, so leaving it is consistency-preserving.

To close it: union `placed` into `enabled` just above this line, after the trust/seed branch.

src/Aetherphone/Core/ControlCenter/ControlLayoutService.cs
```diff
@@ -177,19 +180,56 @@ private void Load()
...
         for (var index = 0; index < all.Count; index++)
         {
-            if (placed.Add(all[index].Id))
+            if (enabled.Contains(all[index].Id) && placed.Add(all[index].Id))
```
XeldarAlz, line 212: Non-blocking, behavior note: this drops the auto-add fallback outright, so a module shipped in a future version lands in the gallery rather than on the grid.

That matches Home's deliberately tested contract (`AppShippedAfterTheSaveWasWritten_IsNotInstalled`) and iOS, so I think it is the right call. Worth stating in the description though, since it removes a capability rather than only fixing one.

Home backs it with `MandatoryApps` to force-add; Control Center has no equivalent. Worth adding one the first time a module has to ship visible by default.

XeldarAlz (commented):
Verified on a clean worktree at 1642e91: `dotnet build -c Release` gives 0 errors (only the pre-existing AngleSharp NU1902), `dotnet test` gives 103/103.

Root cause and migration both check out. A user who removed a module and never re-committed the grid has it absent from `Items`, so `SeedEnabled` applies the removal retroactively. Removing *every* module also stays empty, since an empty `Enabled` falls through to `SeedEnabled` over an equally empty `Items`. Recovery is intact: the Customize button at `ControlCenter.cs:164` is drawn unconditionally, so a bare grid can always reach the gallery.

Approach is right, mirroring `HomeLayoutService.LoadInstalled` is exactly what I'd want here. Nothing blocking. Three comments inline.

src/Aetherphone/Core/ControlCenter/ControlLayoutService.cs
BluntEXE: Fixed: added ControlLayoutServiceInstallTests.cs mirroring HomeLayoutServiceInstallTests.cs's first four cases, including the restart-regression case itself. Needed IControlConfiguration and IControlRegistry seams to make the service fakeable (Configuration/ControlRegistry already implement them, ctor now takes the interfaces), plus a scoped Dalamud assembly reference in Aetherphone.Tests.csproj since GalleryIcon exposes FontAwesomeIcon directly - full writeup in the PR description.

src/Aetherphone/Core/ControlCenter/ControlLayoutService.cs
BluntEXE: Fixed: LoadEnabled now unions placed into enabled right after the trust/seed branch.

src/Aetherphone/Core/ControlCenter/ControlLayoutService.cs
BluntEXE: Added a 'Behavior change worth flagging' section to the PR description covering this.

src/Aetherphone/Core/ControlCenter/ControlLayoutService.cs
```diff
@@ -177,19 +180,58 @@ private void Load()
...
+            SeedEnabled(saved);
+        }
+
+        enabled.UnionWith(placed);
+
         var all = registry.Modules;
         for (var index = 0; index < all.Count; index++)
         {
-            if (placed.Add(all[index].Id))
+            if (enabled.Contains(all[index].Id) && placed.Add(all[index].Id))
```
XeldarAlz, line 212: Non-blocking: this drops auto-add with no `MandatoryApps` equivalent, so a future module stays invisible until someone opens the gallery. A persisted `Removed` set keeps both behaviours: seed it on migration from `registry.Modules` minus `saved.Items`, and existing removals still stick. Your call whether Home-parity is worth more.

src/Aetherphone.Tests/ControlLayoutServiceInstallTests.cs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private sealed class FakeControlRegistry : IControlRegistry
+    {
+        private readonly Dictionary<string, IControlModule> byId;
+
+        public FakeControlRegistry(List<IControlModule> modules)
+        {
+            Modules = modules;
+            byId = modules.ToDictionary(module => module.Id);
```
XeldarAlz, line 118: Nit: house style is index loops, no LINQ, and this is the only LINQ call in `Aetherphone.Tests`. A `for` over `modules` filling `byId` matches the rest of the suite.

XeldarAlz (commented):
Verified locally: `dotnet build -c Release` clean (only the pre-existing AngleSharp NU1902), `dotnet test` 107/107 including the 4 new cases. Not run in-game on my side, your test plan covers the restart path.

Migration checks out: `Enabled` seeds from `Items`, and the module set has not changed since Control Center shipped, so nothing currently on anyone's grid disappears on upgrade. Two comments inline, neither blocking.

PR #66, opened 2026-07-26
fix(yellowpages): decrypt the inquiry list preview

BluntEXE:
## What this fixes

The Yellow Pages Inquiries list row was rendering `thread.LastBody` directly — the raw encrypted envelope straight off the wire (e.g. `AE1.1.HrkznRRZnOer...`) — instead of the decrypted last message. So the card showed ciphertext where the message preview should be.

## Fix

`AdInquiryStore.RevealPreview(thread)` already existed and does exactly what's needed here — checks the envelope version and decrypts via the cipher if encrypted. It just wasn't wired up to the row. One-line fix: read from `RevealPreview` instead of the raw DTO field.

## Verification

Checked the other three E2E-encrypted list surfaces (Message/Linkpearl, Aethergram DMs, Velvet Messages) for the same pattern — all three decrypt `LastMessagePreview` once at the store layer (`cipher.ResolvePreview(...)`) before the field ever reaches the UI, so this was isolated to Yellow Pages, not systemic.

Verified in-game: before the fix the card showed the encrypted blob; after, it shows the actual message text.

## Test plan
- [x] Build succeeds
- [x] Reproduced the bug on current master, confirmed the fix resolves it in-game
- [x] Swept the other three encrypted list surfaces for the same pattern (none affected)

### DIFF
--- a/src/Aetherphone/Apps/YellowPages/YellowPagesApp.Inquiries.cs
+++ b/src/Aetherphone/Apps/YellowPages/YellowPagesApp.Inquiries.cs
@@ -131,7 +131,7 @@ internal sealed partial class YellowPagesApp
         var whoLine = Typography.FitText(who, textRight - textLeft, TextStyles.Caption1);
         Typography.Draw(drawList, new Vector2(textLeft, card.Min.Y + 30f * scale), whoLine, ui.Accent,
             TextStyles.Caption1);
-        var preview = thread.LastBody.Replace('\n', ' ');
+        var preview = inquiries.RevealPreview(thread).Replace('\n', ' ');
         var previewWidth = textRight - textLeft - (thread.UnreadCount > 0 ? 26f * scale : 0f);
         Typography.Draw(drawList, new Vector2(textLeft, card.Min.Y + 47f * scale),
             Typography.FitText(preview, previewWidth, TextStyles.Footnote),
--- a/src/Aetherphone/Core/YellowPages/AdInquiryStore.cs
+++ b/src/Aetherphone/Core/YellowPages/AdInquiryStore.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using Aetherphone.Core.Aethernet;
 using Aetherphone.Core.Aethernet.Clients;
 using Aetherphone.Core.Aethernet.Contracts;
@@ -31,7 +32,7 @@ internal sealed class AdInquiryStore : IDisposable
     private volatile bool sending;
     private string? openThreadId;
     private DateTime nextThreadPoll = DateTime.MinValue;
-    private readonly Dictionary<string, string> scopeCache = new(StringComparer.Ordinal);
+    private readonly ConcurrentDictionary<string, string> scopeCache = new(StringComparer.Ordinal);
 
     public AdInquiryStore(AethernetSession session, YellowPagesClient client, KeyVault vault,
         ConversationKeyStore keys, PhoneVisibility visibility, RealtimeSignalBus signals, AppGate gate)
@@ -59,14 +60,8 @@ internal sealed class AdInquiryStore : IDisposable
 
     public string ScopeFor(string otherUserId)
     {
-        if (scopeCache.TryGetValue(otherUserId, out var cached))
-        {
-            return cached;
-        }
-
-        var scope = ConversationKeyStore.AdScope(ConversationKeyStore.Pair(MyUserId, otherUserId));
-        scopeCache[otherUserId] = scope;
-        return scope;
+        return scopeCache.GetOrAdd(otherUserId,
+            id => ConversationKeyStore.AdScope(ConversationKeyStore.Pair(MyUserId, id)));
     }
 
     /// <summary>Plaintext for display. Bodies always travel and rest as sealed envelopes.</summary>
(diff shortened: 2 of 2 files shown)


Syrilai (approved):
Fix works

XeldarAlz (commented):
Release build succeeds and 139/139 tests pass at 9654ed9; not run in-game on my side, taking your in-game check for the render.

Fix is right. `RevealPreview` has had zero callers since fb44a8f6, so the row was the one surface that was never wired up, and routing through it also upgrades the failure modes: a missing key now shows the placeholder instead of the envelope. Swept the other consumers too, `LastBody` appears nowhere else and the inquiry notification formats the ad title rather than the body, so the isolation claim holds. No blocking findings, two notes inline.

src/Aetherphone/Apps/YellowPages/YellowPagesApp.Inquiries.cs
```diff
@@ -131,7 +131,7 @@ private bool DrawInquiryRow(AdInquiryDto thread, float scale)
         var whoLine = Typography.FitText(who, textRight - textLeft, TextStyles.Caption1);
         Typography.Draw(drawList, new Vector2(textLeft, card.Min.Y + 30f * scale), whoLine, ui.Accent,
             TextStyles.Caption1);
-        var preview = thread.LastBody.Replace('\n', ' ');
+        var preview = inquiries.RevealPreview(thread).Replace('\n', ' ');
```
XeldarAlz, line 134: Non-blocking: `ResolveBody` caches by message id, so this row seeds the entry the thread view later reads (`AdInquiryStore.cs:70`). If the server leaves `LastCommitmentTag` null on the list DTO, that message stays cached with `Verified=false`, which would read as a false tamper hint if YP ever adopts the check at `MessageApp.Thread.cs:364`. Can you confirm the field is populated?

XeldarAlz, line 134: Nit: this runs every frame for every thread (the list is not virtualized), so `ScopeFor` allocates two strings per row per frame, and rows with no CEK yet re-enter the decode path each frame since `ResolveBody` only caches Decrypted/Malformed. Fair trade for recovering the instant keys land, but hoisting the scope string would be cheap.

BluntEXE:
Confirmed the field with a temporary diagnostic log against a live dev inquiry: `LastCommitmentTag` is populated (`Ntv8wqMoyT3SfDOfyTnq0jz62So81SbkwyK+RGH61mo=` on the thread I checked, across 6 polls), not null. So the cache-poisoning scenario you flagged isn't live currently — noted for the record in case that ever changes server-side.

Hoisted the `ScopeFor` scope string into a per-`otherUserId` cache (cleared on account switch) in 2edcb35d — addresses the per-frame allocation nit directly without touching the caching/verification path.

src/Aetherphone/Core/YellowPages/AdInquiryStore.cs
```diff
@@ -56,8 +57,17 @@ public AdInquiryStore(AethernetSession session, YellowPagesClient client, KeyVau
...
-        ConversationKeyStore.AdScope(ConversationKeyStore.Pair(MyUserId, otherUserId));
+    public string ScopeFor(string otherUserId)
+    {
+        if (scopeCache.TryGetValue(otherUserId, out var cached))
+        {
+            return cached;
+        }
+
+        var scope = ConversationKeyStore.AdScope(ConversationKeyStore.Pair(MyUserId, otherUserId));
+        scopeCache[otherUserId] = scope;
```
XeldarAlz: Blocking: `ScopeFor` now writes, and it is called from the render thread (`RevealPreview` every frame), from the pool via `SealAsync` inside `StoreWork.Run`'s `Task.Run`, while `OnSessionChanged` calls `Clear()` on the framework thread. Concurrent writes to a plain `Dictionary` can corrupt the buckets and spin `TryGetValue` forever, hanging the render thread. Make it a `ConcurrentDictionary` like `MessageCipher:34` and `ConversationKeyStore:20`.

XeldarAlz (changes requested):
Release build succeeds and 139/139 tests pass at 2edcb35.

The nit is addressed, the allocation is gone. But the cache made `ScopeFor` a writing method and it is reached from three threads, so it needs to be concurrent; one blocking comment inline. The `LastCommitmentTag` question from the first pass is still open.

src/Aetherphone/Core/YellowPages/AdInquiryStore.cs
BluntEXE: Fixed in b0516fb3 — switched scopeCache to ConcurrentDictionary with GetOrAdd, matching the pattern already used in MessageCipher and ConversationKeyStore. Confirmed 139/139 tests pass.

XeldarAlz:
Merged into master as cc3af607, both commits included. GitHub did not flip the state automatically because the merge was made locally, so closing by hand.

On the open question: the server does populate `LastCommitmentTag`, `AdInquiryEndpoints.cs:351` sets it from the message alongside `lastMessageId`, so the cache poisoning I worried about cannot happen. That note is resolved, no action needed.

I took the thread-safety fix myself in 1e8d6e33 rather than hold the merge on it, `scopeCache` is now a `ConcurrentDictionary`. Same commit also keeps a deleted last message out of the cipher: the server blanks the body of a soft deleted message but leaves `LastEncVersion` at 1 and still sends `LastMessageId`, so the row resolved an empty envelope and read "can't decrypt this message" where it used to be blank.

Thanks for this one, the diagnosis was right and the sweep of the other three surfaces saved a round trip.

PR #67, opened 2026-07-26
[Velvet] Word Wrapping on Comments Bug

BluntEXE:
## Summary
- Comment text in Velvet's post detail view (`DrawCommentRow`) could overflow the card's right edge instead of wrapping, for any comment that doesn't contain a `@mention`, link, or `:shortcode:` emoji (those route through the RichText renderer instead, which wraps correctly).
- Root cause: `ImGui.PushTextWrapPos` expects a window-local X coordinate, but the call site was passing an absolute screen-space X (`textLeft + wrapWidth`). Every other wrap call site in the codebase already subtracts `ImGui.GetWindowPos().X` for exactly this reason — this was the one spot missing it.
- Reused the existing, already-established pattern from the other correct call sites rather than introducing new wrapping logic.

## Test plan
- [x] `dotnet test` — 139/139 pass
- [x] Reproduced in-game: long plain-text comment on a post overflowed past the card edge; confirmed fixed after the change
- [ ] Xeldar to confirm on their end if desired

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Aetherphone/Apps/Velvet/VelvetShell.Detail.cs
+++ b/src/Aetherphone/Apps/Velvet/VelvetShell.Detail.cs
@@ -291,7 +291,7 @@ internal sealed partial class VelvetShell
 
         if (commentLayout is null)
         {
-            ImGui.PushTextWrapPos(textLeft + wrapWidth);
+            ImGui.PushTextWrapPos(textLeft + wrapWidth - ImGui.GetWindowPos().X);
             using (ImRaii.PushColor(ImGuiCol.Text, VelvetTheme.BodyInk))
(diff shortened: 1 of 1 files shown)


XeldarAlz (approved):
Verified: branch builds clean in a fresh worktree (0 errors) and `dotnet test` is 139/139. Not run in-game on my side.

Root cause is correct: `PushTextWrapPos` takes window-local X, and this was the only screen-space call site still missing the `GetWindowPos().X` subtraction (checked all 22 sites; the `0f` and window-local ones are fine as-is). Row height already measured against `wrapWidth`, so the drawn text now matches the space that was already being reserved. No blocking findings.

PR #68, opened 2026-07-27
AetherStream: world-anchored screen, position controls, watch-along sync (joining still WIP)

Kiro-XVI:
## Summary

Reworks AetherStream's in-world screen and adds control/sync features on top of it.

**Screen rendering**
- Replaces the actor-VFX/Penumbra temp-mod mounting with a self-drawn world-space quad (`ScreenPainter`): a textured triangle strip rendered from the game's own Present hook, positioned by an absolute world position/yaw/scale instead of being cast onto the local player's own model. The screen stays exactly where it's placed instead of riding along attached to the player.
- Drops the Penumbra dependency for this feature entirely - no temp collections, no resource-hook interception, no bundled avfx/atex assets.
- Adds a slight cylindrical curve to the screen (procedural, generated in the same shader - no CPU-side vertex buffer needed), scaled proportionally with the screen's own size.

**Position/scale/rotate controls (Casting tab)**
- X/Y/Z as click-and-drag fields instead of type-and-enter text boxes.
- Scale as a slider, now bounded 0.1x-8x (previously unbounded).
- Rotate as a full-turn slider - not present in the old mounting at all, since yaw was fixed at spawn time.
- Position sliders are bounded to a window around wherever the screen was last deliberately placed (spawn/recenter/preset apply), not an arbitrary world origin.
- Recenter button - re-spawn the screen in front of the player without touching playback.
- Named presets: save/apply/remove, now capturing rotation too so a saved spot restores the exact facing.

**Watch-along screen sync**
- Extends the existing `stream.state` envelope with an optional screen transform (`screenX/y/z/yaw/scale`), so a host's screen position/rotation/scale carries over to viewers - each viewer's own client draws its own local copy at the synced coordinates, no shared/networked object involved.
- Fixes a client-side ordering bug found while testing this: `Mode` was being set to `Hosting` optimistically before the first `stream.state` was even sent, instead of waiting for the server's echo to confirm it, per the existing protocol's own contract.

**Zone-scoped discovery ("Streaming nearby")**
- New signal pair, `stream.nearby` / `stream.nearby.roster`: ask "what's live in my current zone" and get back whoever's hosting there, with no mutual-contact relationship required. Surfaced as a "Streaming nearby" list on the Join screen, refreshed on a timer while it's open.

**Connection reachability**
- Pulled forward the realtime-socket reachability fix from master (keep the socket connected while signed in even with Calls disabled, instead of stopping the router entirely) - relevant here since AetherStream's signaling rides the same connection as calls.

## Status / known gaps

**Joining and watch-along generally is still in active development.** The client side here is complete and has been tested against the existing `stream.state`/`stream.join`/`stream.declined` protocol, but:
- `stream.nearby`/`stream.nearby.roster` needs corresponding server-side support to do anything - right now it's a client-only no-op until the server recognizes the new signal types.
- The screen-transform fields on `stream.state` need to be explicitly relayed server-side (not just tolerated) for sync to actually reach other viewers.
- End-to-end join reliability between two live clients is still being worked through - not yet confirmed fully working in real multi-client testing.

## Test plan

- [x] `dotnet build -c Release` - 0 errors
- [x] Play a video/ROM - screen spawns in front of the player, independent of movement
- [x] Casting tab - drag X/Y/Z, Scale, Rotate live; Recenter; save/apply/remove a preset
- [x] Confirm curve renders as intended (subtle bow, not inverted)
- [ ] Two-client watch-along: host screen position syncs to a viewer once server-side relay is in place
- [x] "Streaming nearby" list populates once server-side `stream.nearby` support ships


src/Aetherphone/Core/Video/DxHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+		}
+		catch (Exception e)
+		{
+			AepLog.Error($"[DxHandler] OnPresent subscriber failed: {e}");
+		}
+
+		return _presentHook!.Original(swapChain, syncInterval, flags);
+	}
+
+	public static void Dispose()
```
XeldarAlz, line 81: Blocking: `DxHandler.Dispose()` has no callers. `VideoEngine.Dispose` (VideoEngine.cs:449) tears down the renderers but never unhooks Present, so the hook stays live after the plugin unloads and the next reload or update detours into freed managed code. Call it from `VideoEngine.Dispose`.

src/Aetherphone/Aetherphone.csproj
```diff
@@ -36,6 +36,12 @@
     <Content Include="Sounds\*.mp3;Sounds\*.wav">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="Native\*.dll;Native\*.exe">
```
XeldarAlz, line 39: Blocking: neither bundled binary is ever loaded. `NativeLoader.Resolve` (Resources.cs:344) resolves libmpv from `GetLocationMPV()` and mpv's `ytdl_path` from `GetLocationYTDLP()`, both of which read `ConfigDirectory` and are filled by the runtime downloader. That is 116 MB of dead weight in git and in the zip: plugin output measures 216 MB. The `Resources/Video` include on line 42 also points at a directory that does not exist.

src/Aetherphone/Aetherphone.csproj
```diff
@@ -62,6 +68,14 @@
     <PackageReference Include="SixLabors.ImageSharp" Version="3.1.12" />
     <PackageReference Include="System.Security.Cryptography.ProtectedData" Version="10.0.0" />
     <PackageReference Include="Vortice.Direct3D11" Version="3.8.3" />
+    <PackageReference Include="Vortice.D3DCompiler" Version="3.8.3" />
+    <PackageReference Include="Penumbra.Api" Version="5.15.1" />
     <PackageReference Include="YoutubeExplode" Version="6.6.0" />
+    <PackageReference Include="SharpDX.Direct3D11" Version="4.2.0" />
+    <PackageReference Include="SharpDX.D3DCompiler" Version="4.2.0" />
+    <PackageReference Include="SharpDX.XAudio2" Version="4.2.0" />
+    <PackageReference Include="NoireLib" Version="1.8.0.3" />
```
XeldarAlz, line 77: Blocking: `NoireLib` is not referenced anywhere in the code, but its analyzer adds 650 `NoireLib_002` warnings across pre-existing files (HttpService, RealtimeConnection, every JsonContext) telling us to drop System.Text.Json, and it pulls System.Data.SQLite and Castle.Core into the plugin. Drop the reference.

src/Aetherphone/Aetherphone.csproj
```diff
@@ -62,6 +68,14 @@
     <PackageReference Include="SixLabors.ImageSharp" Version="3.1.12" />
     <PackageReference Include="System.Security.Cryptography.ProtectedData" Version="10.0.0" />
     <PackageReference Include="Vortice.Direct3D11" Version="3.8.3" />
+    <PackageReference Include="Vortice.D3DCompiler" Version="3.8.3" />
+    <PackageReference Include="Penumbra.Api" Version="5.15.1" />
```
XeldarAlz, line 72: Blocking: `Penumbra.Api` and `ZXing.Net` (line 79) exist only for `Core/Video/TextureTranslate.cs`, which nothing instantiates. It is the QR-code-into-Penumbra-.tex path from the VFX approach this PR removes. `APIHelper.cs` and `APIProvider.cs` are dead the same way, `ApiProvider.Init` is never called. Delete the three files and both packages.

src/Aetherphone/Core/Video/WatchAlongSession.cs
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    // Zone-scoped discovery - unlike Join, this needs no hostId and no mutual-contact relationship;
+    // the server answers with whoever is currently hosting (see OnFrameworkUpdate's PublishState
+    // call below) in the caller's own current territory. Response lands in Nearby via OnNearby.
+    public void RequestNearbyStreams() => stream.RequestNearby(Plugin.ClientState.TerritoryType);
+
+    public void Join(string hostId)
+    {
+        if (Mode == WatchAlongMode.Hosting)
```
XeldarAlz, line 108: Blocking: `awaitingHostAck` is only cleared when `Mode == Hosting`. Join between the first `PublishState` and its echo and the flag survives the join; the host's next `stream.state` then takes the `awaitingHostAck` branch in `OnState` and flips a viewer to `Hosting`, skipping the url/seek/pause apply. Gate on `Mode == Hosting || awaitingHostAck`.

src/Aetherphone/Apps/AetherStream/AetherStreamApp.Player.cs
```diff
@@ -0,0 +1,461 @@
...
+        {
+            ImGui.InputTextWithHint("##aetherstreamUrl", Loc.T(L.AetherStream.UrlHint), ref urlInput, 2000,
+                ImGuiInputTextFlags.None);
+        }
+
+        var browseCenter = new Vector2(fieldRect.Max.X + 22f * scale, fieldRect.Center.Y);
+        if (ui.IconButton(browseCenter, 16f * scale, FontAwesomeIcon.FolderOpen.ToIconString(), ui.TitleInk,
+                ui.FieldSurface, 0.6f, Loc.T(L.AetherStream.BrowseLocalFile)))
+        {
+            NativeFileDialog.PickVideo(Loc.T(L.AetherStream.BrowseLocalFile),
```
XeldarAlz, line 374: Blocking: this calls `NativeFileDialog` directly instead of `FilePicker`. That skips the ReShade guard (native shell pickers crash ReShade users on thumbnail resolution) and ignores the user's `UseNativeFileDialog` setting. Add a `PickVideo` to `FilePicker` and route through it, like every other picker in the repo.

src/Aetherphone/Core/Video/VideoEngine.cs
```diff
@@ -0,0 +1,458 @@
...
+        bool isNewSession = !_isActive;
+        _screenPainter.SetTarget(screenTexture);
+
+        if (isNewSession)
+        {
+            SpawnScreenInFrontOfLocalPlayer();
+        }
+    }
+
+    public void Dispose()
```
XeldarAlz, line 449: Blocking: `_screenTexture` and `_snesScreenTexture` are never disposed. Two 1920x1080 D3D11 textures leak on every plugin reload.

src/Aetherphone/Core/Video/VideoEngine.cs
```diff
@@ -0,0 +1,458 @@
...
+
+        Task.Run(async () =>
+        {
+            if (IsYTURL(url))
+            {
+                TimeSpan elapsed = DateTime.Now - _lastLoadYT;
+                if (elapsed.TotalSeconds < 7)
+                {
+                    int sleepTime = Math.Min(Math.Max((int)(7000 - elapsed.TotalMilliseconds), 0), 7000); //Add some sleep time to avoid hitting rate limits
+                    Thread.Sleep(sleepTime);
```
XeldarAlz, line 155: Non-blocking: `Thread.Sleep` parks a thread-pool thread for up to 7 seconds, and the `while (true)` render loop on line 175 holds another one for the entire video. The lambda is `async` but never awaits. Use `await Task.Delay` here and a dedicated `LongRunning` task for the render loop.

src/Aetherphone/Core/Video/VideoEngine.cs
```diff
@@ -0,0 +1,458 @@
...
+        MipLevels = 1,
+        ArraySize = 1,
+        Format = Format.B5G6R5_UNorm,
+        BindFlags = BindFlags.ShaderResource | BindFlags.RenderTarget,
+        CpuAccessFlags = CpuAccessFlags.None,
+        SampleDescription = new SampleDescription(1, 0),
+        Usage = ResourceUsage.Default,
+        OptionFlags = ResourceOptionFlags.None,
+    };
+    private CancellationTokenSource _renderCancellation = new();
```
XeldarAlz, line 77: Non-blocking: `_renderCancellation` is never cancelled or disposed, so all seven `IsCancellationRequested` guards below are permanently false and do nothing. `StopVideo` nulls `_mpvRenderer` while the render loop still reads it, which only surfaces as a caught NRE. `_lastIdle` on line 86 is written every tick and never read.

src/Aetherphone/Plugin.cs
```diff
@@ -39,6 +42,9 @@ public sealed class Plugin : IDalamudPlugin
     [PluginService] internal static IContextMenu ContextMenu { get; private set; } = null!;
     [PluginService] internal static IPluginLog Log { get; private set; } = null!;
     [PluginService] internal static IUnlockState UnlockState { get; private set; } = null!;
+    [PluginService] internal static IGameInteropProvider InteropProvider { get; private set; } = null!;
+    [PluginService] internal static IKeyState KeyState { get; private set; } = null!;
```
XeldarAlz, line 46: Non-blocking: `KeyState`/`GamepadState` and the whole SNES path (Snes9xRenderer 841 lines, InputManager, WndProcKeyUpReader, the two Snes config fields) have no UI entry point, `PlaySnes` has no callers outside Core/Video. It still installs a WndProc hook from the plugin constructor. Either wire it up or cut it from this PR.

src/Aetherphone/Core/Video/Resources.cs
```diff
@@ -0,0 +1,366 @@
...
+			AepLog.Debug("Finished Downloading " + downloadURL);
+			if (nameEndsWith == ".7z")
+			{
+				string localFolder = Path.Combine(configDir, Path.GetRandomFileName());
+				Directory.CreateDirectory(localFolder);
+				using (var archive = ArchiveFactory.OpenArchive(tempFile))
+				{
+					foreach (var entry in archive.Entries.Where(e => !e.IsDirectory))
+					{
+						entry.WriteToDirectory(localFolder, new ExtractionOptions
```
XeldarAlz, line 209: Non-blocking: `WriteToDirectory` with `ExtractFullPath = true` on a downloaded archive has no entry-path validation (zip slip), and neither download is hash-checked before being loaded as native code into the game process. Also `Path.GetTempFileName()` leaks its 0-byte file every time, only `tempFile + ext` is deleted.

src/Aetherphone/Core/Video/ScreenPainter.cs
```diff
@@ -0,0 +1,479 @@
...
+using Buffer = SharpDX.Direct3D11.Buffer;
+using GfxKernel = FFXIVClientStructs.FFXIV.Client.Graphics.Kernel;
+using GfxScene = FFXIVClientStructs.FFXIV.Client.Graphics.Scene;
+using GfxGui = FFXIVClientStructs.FFXIV.Component.GUI;
+using GameControl = FFXIVClientStructs.FFXIV.Client.Game.Control;
+using NumericsMatrix4x4 = System.Numerics.Matrix4x4;
+
+namespace Aetherphone.Core.Video;
+
+// Ported from AlphaChannel's ScreenPainter (Voudi, GPL-3.0, tag v1.1.20260725.1088) - the revamp that
```
XeldarAlz, line 16: Nit: this file plus DxHandler, Resources, MpvRenderer and Snes9xRenderer use tabs, `_field` prefixes, block-scoped namespaces and unsealed classes. The rest of the repo is 4 spaces, unprefixed fields, file-scoped namespaces and sealed. Worth normalizing on the way in so the ported files do not read as a separate codebase.

src/Aetherphone/Apps/AetherStream/AetherStreamApp.Casting.cs
```diff
@@ -0,0 +1,215 @@
...
+        if (SmallButton(savePresetRect, Loc.T(L.AetherStream.CastingSavePreset), !string.IsNullOrWhiteSpace(screenPresetName), scale)
+            && !string.IsNullOrWhiteSpace(screenPresetName))
+        {
+            screen.Engine.SaveScreenPreset(screenPresetName);
+            screenPresetName = "";
+        }
+
+        bodyTop = savePresetRect.Max.Y + 12f * scale;
+
+        var presets = screen.Engine.GetScreenPresets();
```
XeldarAlz, line 175: Nit: `GetScreenPresets()` returns a fresh copy, so this allocates a new List every frame the Casting tab is open. Expose the backing list as `IReadOnlyList<ScreenPositionPreset>` instead.

src/Aetherphone/Core/Localization/L.cs
```diff
@@ -1857,6 +1858,105 @@ internal static class Calculator
         public static readonly LocString Error = new("calculator.error", "Error");
     }
 
+    internal static class AetherStream
```
XeldarAlz, line 1861: Nit: ~100 new strings land in L.cs only, none of the 8 non-English JSONs got keys, so AetherStream is English-only in every other language. There is also no `Icons/aetherstream.png`, so the home tile falls back to the letter glyph instead of a Tabler icon like every other app.

XeldarAlz (changes requested):
Built `Aetherphone.sln -c Release` off the branch head: 0 errors, 658 warnings (650 of them newly introduced), 67/67 tests pass. Not run in-game.

The world-space rendering itself reads well and the ScreenPainter notes on depth and viewport are genuinely useful. Findings are inline. Separately, the branch conflicts with master in 7 files (CallHub, PhoneServices, Plugin, AppRegistry, AppAccents, SettingsRow, SocialNotificationService), several of which already carry the poll-interval and Reconcile changes re-applied here, so it needs a rebase before any of this is actionable.

Kiro-XVI:
Superseded by https://github.com/XeldarAlz/FFXIV-Aetherphone/pull/76, which builds on this same port/alphachannel-engine work and adds join-approval scaffolding, party mode, and the watch-along sync fixes (main-thread deferral, screen-transform timing) that make joining actually work. Closing this one in favor of that.

PR #69, opened 2026-07-27
Reject oversized images before decode to prevent OOM DoS

BluntEXE:
## Summary

`ImageProcessor.DecodeToTextureAsync` called `Image.Load<Rgba32>` directly on
remote bytes. ImageSharp allocates the full RGBA buffer from the
header-declared width/height before it validates any pixel data. A 389KB PNG
with a 20000x20000 header pushes RSS from 47MB to 2.3GB. A 30000x30000
variant hits 3.6GB and overflows the `int` in `image.Width * image.Height * 4`.

This path takes bytes straight from the network. `MediaCache`,
`RemoteImageCache`, and `WallpaperLibrary` all feed avatars, chat media, and
wallpapers into it. Any peer can crash every client that renders their
content.

## Fix

- `EnsureDecodable` calls `Image.Identify` first and rejects anything over
  64M pixels (~256MB RGBA buffer) before the real decode runs.
- Wired into `DecodeToTextureAsync` (the network path) and the three
  local-file `Bake*` methods, same bug class.
- `checked` on the `Width * Height * 4` multiply, plus a static assert on
  `MaxDecodePixels` at class init, so raising the cap later without checking
  the math fails at load instead of overflowing silently at runtime.
- The path-based overload delegates to the stream-based one, no duplicated
  validation logic.

## Testing

- Reproduced the reported PoC dimensions (20000x20000 and 30000x30000)
  against the real ImageSharp version this project uses: both rejected via
  `Image.Identify` before any pixel buffer is allocated, no RSS spike.
- A normal image still decodes without behavior change.
- All six call sites already catch `Exception` around the decode call, so
  rejected images fail like any other decode error — logged, marked failed,
  no new crash surface.
- `dotnet build` clean.

## Open question for maintainers

Does the server enforce any dimension or file-size limit on avatar/media
uploads independent of this client-side fix? Other clients stay exposed
until they update.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Aetherphone/Core/Media/ImageProcessor.cs
+++ b/src/Aetherphone/Core/Media/ImageProcessor.cs
@@ -30,24 +30,27 @@ internal static class ImageProcessor
     private const int JpegQuality = 88;
 
     private const long MaxDecodePixels = 4096L * 4096L;
+    private const long MaxLocalDecodePixels = 8192L * 8192L;
     private static readonly DecoderOptions SingleFrame = new() { MaxFrames = 1 };
 
-    private static void EnsureDecodable(Stream stream)
+    private static void EnsureDecodable(Stream stream, long maxPixels)
     {
         var info = Image.Identify(stream);
         stream.Position = 0;
 
-        if ((long)info.Width * info.Height > MaxDecodePixels)
+        if ((long)info.Width * info.Height > maxPixels)
         {
             throw new InvalidImageContentException(
-                $"Image dimensions {info.Width}x{info.Height} exceed the {MaxDecodePixels} pixel limit.");
+                $"Image dimensions {info.Width}x{info.Height} exceed the {maxPixels} pixel limit.");
         }
     }
+
+    private static void EnsureDecodable(Stream stream) => EnsureDecodable(stream, MaxDecodePixels);
 
     private static void EnsureDecodable(string path)
     {
         using var stream = File.OpenRead(path);
-        EnsureDecodable(stream);
+        EnsureDecodable(stream, MaxLocalDecodePixels);
     }
 
     public static BakedImage BakeSquareJpeg(string sourcePath, WallpaperCrop crop, int target)
@@ -105,17 +108,12 @@ internal static class ImageProcessor
     public static (byte[] Pixels, int Width, int Height) DecodeRgba32(byte[] bytes)
     {
         using var stream = new MemoryStream(bytes);
-        var (pooled, length, width, height) = DecodeRgba32Pooled(stream);
-        try
-        {
-            var pixels = new byte[length];
-            pooled.AsSpan(0, length).CopyTo(pixels);
-            return (pixels, width, height);
-        }
-        finally
-        {
-            ArrayPool<byte>.Shared.Return(pooled);
-        }
+        EnsureDecodable(stream);
+        using var image = Image.Load<Rgba32>(SingleFrame, stream);
+        var length = checked(image.Width * image.Height * 4);
+        var pixels = new byte[length];
+        image.CopyPixelDataTo(pixels);
+        return (pixels, image.Width, image.Height);
     }
 
     public static async Task<IDalamudTextureWrap> DecodeToTextureAsync(ITextureProvider textures, byte[] bytes,
--- a/src/Aetherphone/Core/Media/ImageProcessor.cs
+++ b/src/Aetherphone/Core/Media/ImageProcessor.cs
@@ -28,6 +28,41 @@ internal static class ImageProcessor
 {
     private const int JpegQuality = 88;
 
+    // Rgba32 buffer is Width*Height*4 bytes; cap keeps that under ~256MB and
+    // keeps Width*Height*4 from overflowing int32 (see DecodeToTextureAsync).
+    private const long MaxDecodePixels = 64L * 1024 * 1024;
+
+    static ImageProcessor()
+    {
+        if (MaxDecodePixels * 4 > int.MaxValue)
+        {
+            throw new InvalidOperationException(
+                $"{nameof(MaxDecodePixels)} is too large: Width*Height*4 would overflow int32.");
+        }
+    }
+
+    private static void EnsureDecodable(Stream stream)
+    {
+        var info = Image.Identify(stream);
+        stream.Position = 0;
+        if (info is null)
+        {
+            throw new InvalidImageContentException("Unrecognized image format.");
+        }
+
+        if ((long)info.Width * info.Height > MaxDecodePixels)
+        {
+            throw new InvalidImageContentException(
+                $"Image dimensions {info.Width}x{info.Height} exceed the {MaxDecodePixels} pixel limit.");
+        }
+    }
+
+    private static void EnsureDecodable(string path)
+    {
+        using var stream = File.OpenRead(path);
+        EnsureDecodable(stream);
+    }
+
     public static BakedImage BakeSquareJpeg(string sourcePath, WallpaperCrop crop, int target)
     {
         return BakeCroppedJpeg(sourcePath, crop, target, target);
@@ -35,6 +70,7 @@ internal static class ImageProcessor
 
     public static BakedImage BakeCroppedJpeg(string sourcePath, WallpaperCrop crop, int targetWidth, int targetHeight)
     {
+        EnsureDecodable(sourcePath);
         using var image = Image.Load(sourcePath);
         var size = new Vector2(image.Width, image.Height);
         var aspect = (float)targetWidth / targetHeight;
@@ -52,6 +88,7 @@ internal static class ImageProcessor
 
     public static BakedImage BakeJpeg(string sourcePath, int maxDimension)
     {
+        EnsureDecodable(sourcePath);
         using var image = Image.Load(sourcePath);
         var width = image.Width;
         var height = image.Height;
@@ -73,8 +110,10 @@ internal static class ImageProcessor
     {
         var (pixels, length, width, height) = await Task.Run(() =>
         {
-            using var image = Image.Load<Rgba32>(bytes);
-            var length = image.Width * image.Height * 4;
+            using var stream = new MemoryStream(bytes);
+            EnsureDecodable(stream);
+            using var image = Image.Load<Rgba32>(stream);
+            var length = checked(image.Width * image.Height * 4);
             var buffer = ArrayPool<byte>.Shared.Rent(length);
             image.CopyPixelDataTo(buffer.AsSpan(0, length));
             return (buffer, length, image.Width, image.Height);
--- a/src/Aetherphone/Core/Media/ImageProcessor.cs
+++ b/src/Aetherphone/Core/Media/ImageProcessor.cs
@@ -29,9 +29,9 @@ internal static class ImageProcessor
 {
     private const int JpegQuality = 88;
 
-    private const long MaxDecodePixels = 4096L * 4096L;
-    private const long MaxLocalDecodePixels = 8192L * 8192L;
-    private static readonly DecoderOptions SingleFrame = new() { MaxFrames = 1 };
+    public const long MaxDecodePixels = 4096L * 4096L;
+    public const long MaxLocalDecodePixels = 8192L * 8192L;
+    internal static readonly DecoderOptions SingleFrame = new() { MaxFrames = 1 };
 
     private static void EnsureDecodable(Stream stream, long maxPixels)
     {
@@ -45,12 +45,12 @@ internal static class ImageProcessor
         }
     }
 
-    private static void EnsureDecodable(Stream stream) => EnsureDecodable(stream, MaxDecodePixels);
-
-    private static void EnsureDecodable(string path)
+    private static Image<Rgba32> LoadRgba32(Stream stream, long maxPixels, out int length)
     {
-        using var stream = File.OpenRead(path);
-        EnsureDecodable(stream, MaxLocalDecodePixels);
+        EnsureDecodable(stream, maxPixels);
+        var image = Image.Load<Rgba32>(SingleFrame, stream);
+        length = checked(image.Width * image.Height * 4);
+        return image;
     }
 
     public static BakedImage BakeSquareJpeg(string sourcePath, WallpaperCrop crop, int target)
@@ -60,8 +60,9 @@ internal static class ImageProcessor
 
     public static BakedImage BakeCroppedJpeg(string sourcePath, WallpaperCrop crop, int targetWidth, int targetHeight)
     {
-        EnsureDecodable(sourcePath);
-        using var image = Image.Load(SingleFrame, sourcePath);
+        using var sourceStream = File.OpenRead(sourcePath);
+        EnsureDecodable(sourceStream, MaxLocalDecodePixels);
+        using var image = Image.Load(SingleFrame, sourceStream);
         var size = new Vector2(image.Width, image.Height);
         var aspect = (float)targetWidth / targetHeight;
         var clamped = crop.Clamped(size, aspect);
@@ -78,8 +79,9 @@ internal static class ImageProcessor
 
     public static BakedImage BakeJpeg(string sourcePath, int maxDimension)
     {
-        EnsureDecodable(sourcePath);
-        using var image = Image.Load(SingleFrame, sourcePath);
+        using var sourceStream = File.OpenRead(sourcePath);
+        EnsureDecodable(sourceStream, MaxLocalDecodePixels);
+        using var image = Image.Load(SingleFrame, sourceStream);
         var width = image.Width;
         var height = image.Height;
         if (width > maxDimension || height > maxDimension)
@@ -95,11 +97,9 @@ internal static class ImageProcessor
         return new BakedImage(stream.ToArray(), width, height);
     }
 
-    private static (byte[] Pixels, int Length, int Width, int Height) DecodeRgba32Pooled(Stream stream)
+    private static (byte[] Pixels, int Length, int Width, int Height) DecodeRgba32Pooled(Stream stream, long maxPixels)
     {
-        EnsureDecodable(stream);
-        using var image = Image.Load<Rgba32>(SingleFrame, stream);
-        var length = checked(image.Width * image.Height * 4);
+        using var image = LoadRgba32(stream, maxPixels, out var length);
         var buffer = ArrayPool<byte>.Shared.Rent(length);
         image.CopyPixelDataTo(buffer.AsSpan(0, length));
         return (buffer, length, image.Width, image.Height);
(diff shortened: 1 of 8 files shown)


src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -28,13 +28,49 @@ internal static class ImageProcessor
...
+    private static void EnsureDecodable(Stream stream)
+    {
+        var info = Image.Identify(stream);
+        stream.Position = 0;
+        if (info is null)
+        {
+            throw new InvalidImageContentException("Unrecognized image format.");
+        }
+
+        if ((long)info.Width * info.Height > MaxDecodePixels)
```
XeldarAlz: Blocking: the cap is per frame, so multi-frame formats walk straight past it. A 1,174 byte GIF with a 2000x2000 logical screen and 40 one-pixel frame blocks: `Identify` reports 2000x2000 (4M pixels, accepted), `Image.Load` then materialises all 40 full size frames and working set goes 30MB to 643MB. At the 64M cap that is 256MB per frame with no bound on the count, and `image/gif` is an allowed upload type. `Identify` already gives you `info.FrameMetadataCollection.Count`: fold it into the multiply, or decode with `new DecoderOptions { MaxFrames = 1 }` (`CopyPixelDataTo` only reads the root frame, so callers are unaffected).

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -73,8 +110,10 @@ public static async Task<IDalamudTextureWrap> DecodeToTextureAsync(ITextureProvi
     {
         var (pixels, length, width, height) = await Task.Run(() =>
         {
-            using var image = Image.Load<Rgba32>(bytes);
-            var length = image.Width * image.Height * 4;
+            using var stream = new MemoryStream(bytes);
+            EnsureDecodable(stream);
```
XeldarAlz: Blocking: one network decode is still unguarded. `ChatThreadView.SaveImage` (src/Aetherphone/Windows/Components/ChatThreadView.cs:721) fetches DM media over HTTP and calls `Image.Load<Rgba32>(bytes)` directly, then `new byte[image.Width * image.Height * 4]` unchecked. Same peer supplied bytes, same bug class, and that file is shared by ChocoChat and Velvet.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -28,13 +28,49 @@ internal static class ImageProcessor
 {
     private const int JpegQuality = 88;
 
+    // Rgba32 buffer is Width*Height*4 bytes; cap keeps that under ~256MB and
+    // keeps Width*Height*4 from overflowing int32 (see DecodeToTextureAsync).
+    private const long MaxDecodePixels = 64L * 1024 * 1024;
```
XeldarAlz: Non-blocking: 64M pixels is 256MB of Rgba32 per image, and `RemoteImageCache.LoadAsync` starts one decode per key with no concurrency limit, so a feed of distinct 8191x8191 avatars still stacks them. Nothing in a phone sized UI needs past 4096x4096 (64MB); the tighter cap costs nothing real.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -28,13 +28,49 @@ internal static class ImageProcessor
...
+
+    static ImageProcessor()
```
XeldarAlz: Non-blocking: this can never fire. Both operands are compile time constants so the branch folds away, and the static ctor costs the type its `beforefieldinit`, adding an init check to every static call into this path. The `checked()` below already catches the overflow if the cap is ever raised.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -28,13 +28,49 @@ internal static class ImageProcessor
...
+            throw new InvalidOperationException(
+                $"{nameof(MaxDecodePixels)} is too large: Width*Height*4 would overflow int32.");
+        }
+    }
+
+    private static void EnsureDecodable(Stream stream)
+    {
+        var info = Image.Identify(stream);
+        stream.Position = 0;
+        if (info is null)
```
XeldarAlz: Nit: dead on ImageSharp 3.1.12. `Image.Identify` returns a non-nullable `ImageInfo` and throws `UnknownImageFormatException` for unrecognised input, so this branch never runs.

XeldarAlz (changes requested):
Built the branch clean in a scratch worktree (Release, only the pre-existing NU1902 AngleSharp warnings); not run in-game. Reproduced the guard standalone against ImageSharp 3.1.12: it does stop the reported 20000x20000 and 30000x30000 header bombs, but a multi-frame image walks straight past it, and one network decode path is left unguarded. Two blocking comments inline, three smaller ones.

On your open question: no. `/media/upload-url` allowlists png/jpeg/webp/gif, and the 8MB cap in `MediaEndpoints` only covers the local dev storage path; on R2 the client PUTs straight to a presigned bucket URL, so the API never sees the bytes. Nothing validates dimensions or frame count server side. Worth a separate backend issue.

src/Aetherphone/Core/Media/ImageProcessor.cs
BluntEXE: Confirmed and fixed in bb2b4b7e. Reproduced it to be sure before touching the code: a 40-frame 2000x2000 GIF (8.4KB on disk) went 30MB -> 643MB with the old guard, matches your numbers. `DecoderOptions { MaxFrames = 1 }` is now passed to every decode call (`Image.Load`/`Image.Load<Rgba32>`), verified the same GIF now decodes 1 frame at 15MB instead of 40 at 610MB. Went with your MaxFrames suggestion over folding frame count into the multiply since it also bounds decode time, not just memory.

src/Aetherphone/Core/Media/ImageProcessor.cs
BluntEXE: Good catch, missed this one. Filtered my original review by callers of DecodeToTextureAsync instead of checking all 5 files that hit Image.Load/Rgba32 directly, and this one inlines its own decode instead of calling into ImageProcessor. Fixed in bb2b4b7e: added ImageProcessor.DecodeRgba32(byte[]) as the guarded equivalent of the old inline block, ChatThreadView.SaveImage now calls that instead. Same EnsureDecodable + MaxFrames=1 path as DecodeToTextureAsync, no duplicated validation logic.

src/Aetherphone/Core/Media/ImageProcessor.cs
BluntEXE: Agreed, done in bb2b4b7e. Tightened to 4096x4096 (16.7M px / 64MB). Doesn't fully solve the no-concurrency-limit stacking issue on RemoteImageCache you flagged, but it does shrink the worst case ~4x; a concurrency cap there feels like a separate change outside this PR's scope.

src/Aetherphone/Core/Media/ImageProcessor.cs
BluntEXE: Fair, removed in bb2b4b7e. The checked() below is the right place for it.

src/Aetherphone/Core/Media/ImageProcessor.cs
BluntEXE: Confirmed via reflection against 3.1.12: Identify(Stream) returns ImageInfo, not ImageInfo?, and throws UnknownImageFormatException instead of returning null. Removed the dead branch in bb2b4b7e.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -28,14 +29,36 @@ internal static class ImageProcessor
 {
     private const int JpegQuality = 88;
 
+    private const long MaxDecodePixels = 4096L * 4096L;
```
XeldarAlz: Blocking, and my fault: one shared cap now gates local files too, so this rejects images that worked before the PR.

`PhotosApp.LoadThumbnailAsync` (:651, bake from path) and `LoadFullAsync` (:683, decode raw file bytes) both run through here, so anything over 16.7M pixels on your own disk is now refused. That band is real in this game: 5760x3240 (DLDSR 2.25x from 4K) is 18.7M and 7680x4320 (DSR 4x) is 33.2M, both rejected, while 3840x2160 and 5120x2880 still pass. Same for setting one as a wallpaper, posting it, or attaching it to Feedback / Yellow Pages. It fails to an `AepLog.Warning` and a failed tile with nothing telling the user why.

My "nothing in a phone sized UI needs past 4096x4096" argument only holds for bytes arriving off the network, where the point is that you did not choose them. A file the user picked and that is about to be downscaled to 256px or 1600px anyway is a different trust tier. Suggested shape: keep 4096x4096 for the byte[] entry points fed by peers (`DecodeRgba32`, the remote caches) and give the path-based bakes a `MaxLocalDecodePixels` of 8192x8192, which is where 78df4d14 had it and covers every setup above.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -68,16 +92,39 @@ public static BakedImage BakeJpeg(string sourcePath, int maxDimension)
...
+        return (buffer, length, image.Width, image.Height);
+    }
+
+    public static (byte[] Pixels, int Width, int Height) DecodeRgba32(byte[] bytes)
+    {
+        using var stream = new MemoryStream(bytes);
+        var (pooled, length, width, height) = DecodeRgba32Pooled(stream);
+        try
+        {
+            var pixels = new byte[length];
```
XeldarAlz: Non-blocking: this rents a pooled buffer, copies into it, allocates the real array, copies again, then returns the pooled one. The inline code it replaced did one allocation and one copy. On a 64MB save that is an extra full-buffer pass bought purely for DRY. Decoding straight into `new byte[length]` keeps the shared `EnsureDecodable` + `MaxFrames` path without the second copy.

XeldarAlz (changes requested):
Re-reviewed at bb2b4b7e against the code, not the replies. Built clean in a scratch worktree (Release, only the pre-existing NU1902 AngleSharp warnings); not run in-game.

Both blockers are genuinely closed. `MaxFrames = 1` is on all three `Image.Load` calls in `ImageProcessor` (:61, :79, :98), and `ChatThreadView.SaveImage` (:721) now routes through `DecodeRgba32` with no duplicated validation. Swept the branch for `Image.Load`/`Identify`: every site is inside `ImageProcessor` now except `WallpaperLibrary.MeasureBrightness:357`, which is fed only by `File.ReadAllBytesAsync` and sits two lines after a `DecodeToTextureAsync` that throws first, so it is transitively covered. Both network callers already `catch (Exception)` and mark failed, so the new throw degrades instead of killing a task. The three smaller items are done too.

One regression to fix before merge, and it is mine: the 4096x4096 cap I asked for landed on the local-file paths as well, not just peer-supplied bytes. Detail inline.

src/Aetherphone/Core/Media/ImageProcessor.cs
BluntEXE: Fixed: split into MaxDecodePixels (4096x4096) for byte[] entry points fed by peers, and MaxLocalDecodePixels (8192x8192) for path-based bakes. Covers DLDSR/DSR screenshots up to 33.2M pixels.

src/Aetherphone/Core/Media/ImageProcessor.cs
BluntEXE: Fixed: DecodeRgba32 now decodes straight into the returned byte[], no pooled intermediate.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -28,14 +29,39 @@ internal static class ImageProcessor
...
+        stream.Position = 0;
+
+        if ((long)info.Width * info.Height > maxPixels)
+        {
+            throw new InvalidImageContentException(
+                $"Image dimensions {info.Width}x{info.Height} exceed the {maxPixels} pixel limit.");
+        }
+    }
+
+    private static void EnsureDecodable(Stream stream) => EnsureDecodable(stream, MaxDecodePixels);
```
XeldarAlz: Blocking: the split is on API shape (`string` vs `Stream`/`byte[]`), but provenance does not follow that shape. Four local-file callers read their own bytes off disk and land on this overload, so they get the 4096x4096 peer cap:

- `PhotosApp.LoadFullAsync` (src/Aetherphone/Apps/Photos/PhotosApp.cs:682-683)
- `WallpaperLibrary.LoadAsync` (src/Aetherphone/Core/Wallpapers/WallpaperLibrary.cs:320-321)
- `WallpaperImageCache.LoadAsync` (src/Aetherphone/Core/Wallpapers/WallpaperImageCache.cs:45-46)
- `PhotosWidget.LoadAsync` (src/Aetherphone/Windows/Widgets/PhotosWidget.cs:150-151)

All four are `File.ReadAllBytesAsync(path)` into `DecodeToTextureAsync`. So 5760x3240 (18.7M) and 7680x4320 (33.2M) are still rejected, just one call deeper than in bb2b4b7e.

It is also worse than a uniform reject now. `LoadThumbnailAsync` bakes through `BakeJpeg(path, ThumbnailMaxDimension)` at the 8192 cap and succeeds, so the tile renders in the grid; tapping it runs `LoadFullAsync` on the raw file at 4096 and fails. Before this commit both failed together. Same shape for a wallpaper the user already had set: it is on disk, it was working, and it stops rendering.

The cap has to be chosen by the caller, not by the overload it happens to reach. Take a `maxPixels` argument on `DecodeToTextureAsync` and thread it through `DecodeRgba32Pooled`; pass `MaxLocalDecodePixels` from the four above and `MaxDecodePixels` from `MediaCache:86` and `RemoteImageCache:134`, which are the actual peer-fed entry points.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -28,14 +29,39 @@ internal static class ImageProcessor
 {
     private const int JpegQuality = 88;
 
+    private const long MaxDecodePixels = 4096L * 4096L;
+    private const long MaxLocalDecodePixels = 8192L * 8192L;
+    private static readonly DecoderOptions SingleFrame = new() { MaxFrames = 1 };
```
XeldarAlz: Non-blocking: `MaxFrames = 1` does not reach the last decode left outside this class. `WallpaperLibrary.MeasureBrightness` (src/Aetherphone/Core/Wallpapers/WallpaperLibrary.cs:357) still calls `Image.Load<Rgba32>(bytes)` with no options.

Dimensions there are transitively covered, since the `DecodeToTextureAsync` two lines above at :321 throws first. Frame count is not, and that is a consequence of this line specifically: the guarded decode now passes on a multi-frame image because it only reads one frame, and then `RecordBrightness` re-decodes the same bytes with every frame materialised. 3.1.12 does decode APNG (`AnimationControl`, `FrameControl` and `PngFrameMetadata` are all in the assembly) and `.png` is in `BuiltInPatterns` (:23), so a 4096x4096 animated PNG is 64MB per frame with no bound on the count.

Local files only, so self-inflicted rather than a peer vector, which is why this is not blocking. But it is the exact bug class this PR exists to close and it is one argument.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -28,14 +29,39 @@ internal static class ImageProcessor
...
+                $"Image dimensions {info.Width}x{info.Height} exceed the {maxPixels} pixel limit.");
+        }
+    }
+
+    private static void EnsureDecodable(Stream stream) => EnsureDecodable(stream, MaxDecodePixels);
+
+    private static void EnsureDecodable(string path)
+    {
+        using var stream = File.OpenRead(path);
+        EnsureDecodable(stream, MaxLocalDecodePixels);
```
XeldarAlz: Non-blocking: this opens the file and parses the header, closes it, then `Image.Load(SingleFrame, sourcePath)` at :64 and :82 reopens the same file and parses the header again. Two opens and two header parses per bake, on a path that already runs on every Photos thumbnail miss.

Holding the stream and handing it to `Image.Load` does both in one pass, and incidentally closes the window where the file changes between the check and the load.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -68,16 +95,34 @@ public static BakedImage BakeJpeg(string sourcePath, int maxDimension)
...
+        var buffer = ArrayPool<byte>.Shared.Rent(length);
+        image.CopyPixelDataTo(buffer.AsSpan(0, length));
+        return (buffer, length, image.Width, image.Height);
+    }
+
+    public static (byte[] Pixels, int Width, int Height) DecodeRgba32(byte[] bytes)
+    {
+        using var stream = new MemoryStream(bytes);
+        EnsureDecodable(stream);
+        using var image = Image.Load<Rgba32>(SingleFrame, stream);
```
XeldarAlz: Nit: this and `DecodeRgba32Pooled` (:98) now repeat the same guard, `Image.Load<Rgba32>(SingleFrame, stream)` and `checked` length triple, differing only in where the buffer comes from.

Not worth reviving the pooled indirection I asked you to drop. But threading `maxPixels` for the blocking comment above means making the same edit in both bodies, which is the drift this invites. A private helper that returns the loaded image plus the length would leave one copy of the policy.

XeldarAlz (changes requested):
Re-reviewed at 31dcfc20 against the code. Built clean in a scratch worktree (Release, 0 errors, only the pre-existing NU1902 AngleSharp warnings); not run in-game.

The pooled double-copy is gone and the local/peer split exists, so both items from the last round are addressed as written. Also checked the shared `SingleFrame` static while I was in there: every `DecoderOptions` property on 3.1.12 is `init`-only, so reusing one instance across concurrent decodes is safe. No issue there.

The split is drawn on the wrong axis though, so the regression I blocked on last round is only half fixed, and Photos now fails asymmetrically because of it. One blocking comment, three smaller ones.

src/Aetherphone/Core/Media/ImageProcessor.cs
BluntEXE: Fixed in 50193317: `DecodeToTextureAsync` now takes a `maxPixels` argument, threaded through `DecodeRgba32Pooled`. The 4 local-file callers listed (PhotosApp.LoadFullAsync, WallpaperLibrary.LoadAsync, WallpaperImageCache.LoadAsync, PhotosWidget.LoadAsync) now pass `MaxLocalDecodePixels`; MediaCache:86 and RemoteImageCache:134 pass `MaxDecodePixels`. Verified via grep that every `DecodeToTextureAsync` call site in the tree resolves to one or the other.

src/Aetherphone/Core/Media/ImageProcessor.cs
BluntEXE: Fixed in 50193317: `WallpaperLibrary.MeasureBrightness` now calls `Image.Load<Rgba32>(ImageProcessor.SingleFrame, bytes)`, reusing the same `MaxFrames = 1` guard (made `internal` so it's visible outside `ImageProcessor`). Bounds the animated-PNG frame count same as every other decode site.

src/Aetherphone/Core/Media/ImageProcessor.cs
BluntEXE: Fixed in 50193317: `BakeCroppedJpeg`/`BakeJpeg` now open the file stream once and hand that same stream to both `EnsureDecodable` and `Image.Load`. The string-overload of `EnsureDecodable` is gone, one open + one header parse per bake.

src/Aetherphone/Core/Media/ImageProcessor.cs
BluntEXE: Fixed in 50193317: extracted a single `LoadRgba32(stream, maxPixels, out length)` helper doing the guard + load + length calc. `DecodeRgba32` and `DecodeRgba32Pooled` both call it, so the `maxPixels` threading only had to land in one place.

XeldarAlz (approved):
Re-reviewed at 50193317 against the code. Built clean in a scratch worktree (Release, 0 errors, only the pre-existing NU1902 AngleSharp warnings); 139/139 tests pass. Not run in-game.

All four items from the last round are closed as written. The cap split is now on provenance: `maxPixels` is a required parameter on `DecodeToTextureAsync` with no defaulted overload, so the compiler forces every call site to choose. Verified all six callers pick the right cap, including the thumbnail path staying on the peer cap (those bytes are self-baked JPEGs, so that is correct). Swept the tree: the only decodes left outside `ImageProcessor` are `WallpaperLibrary.MeasureBrightness:357` (now on `SingleFrame`) and `ChatThreadView:721` (through `DecodeRgba32`, peer cap). Merging.

PR #70, opened 2026-07-27
Fix text truncation, hover-scroll, and layout overlap across the app

BluntEXE:
## Summary

Fixes a batch of text truncation, hover-scroll, and layout overlap bugs reported across the app (Photos, Notes, Timers, Dailies, Activity, Fishing, Market, Clock, Aethergram, Velvet, Music, Health, Yellow Pages, Muster, Direct Messages, Linkshells, Collections, and the three arcade mini-games), plus a couple of framework-level bugs found along the way that were causing some of them.

## Root causes found

Most of these bugs traced back to a small number of repeating patterns rather than being independent one-offs:

- **Shared hover state on stacked text.** A single hover boolean, computed for a whole row, was passed into two separate `Marquee` calls (title + subtitle). Hovering either line scrolled both. Fixed by giving each line its own hover check, using `Marquee.DrawLeftAuto`/`DrawCenteredAuto` where possible.
- **Unclamped sibling eating shared width.** A row split between two elements (e.g. a name and a value, or two pills) where one side was measured and clamped but the other was drawn at full natural width. When the unclamped side grew, it pushed into or past its sibling. Fixed by giving both sides a computed floor and letting them shrink or scroll together, following the same pattern used in `MarketApp.Detail.cs`'s hero card fix.
- **`Typography.DrawCentered`'s auto-wrap heuristic.** The no-`maxWidth` overload estimates a safe wrap width from the *current ImGui window's* edges relative to the text's center point. That assumption breaks for anything not roughly centered in a full-size window: a small host window (the minimized-phone icon), or an element sitting near the edge of a multi-column layout. Found three independent instances of this (Yellow Pages intent-tile category count, the minimized-phone notification badge, hover tooltips in general). Fixed each call site by bypassing the auto-wrap path with an explicit-width `Marquee` call or a manual measure-and-draw, rather than patching the heuristic itself (too many other call sites depend on its current behavior to change it globally without a full audit).
- **Padding equal to, not greater than, corner radius.** A pill button's text padding must exceed its corner radius (`height * 0.5`), not just be some positive number, or the text visually crowds the curve.

## Per-area changes

- **Photos, Notes, Timers, Dailies, Activity, Fishing, Market, Clock, Aethergram, Velvet, Health, Yellow Pages, Muster, Direct Messages**: converted static `Typography.FitText` truncation to `Marquee` hover-scroll wherever text could realistically overflow, applying the shared-hover and unclamped-sibling fixes above where relevant. This includes Muster's host name/status/meta text (`MusterApp.Detail.cs`, `.Directory.cs`, `.Manage.cs`, `MusterCard.cs`), the DM/Aethergram thread header name and subtitle (`MessageApp.Thread.cs`, `AethergramApp.Thread.cs`), and a Velvet comment timestamp that could overlap a review badge.
- **Aethergram**: the multi-photo page-indicator dots on both the feed card and detail page now only render at phone size M and up; below that there isn't room to show them without colliding with the share button, so they're hidden entirely instead of squeezed in.
- **Music**: Recently Played chip title/author text is hidden below phone size M for the same reason; the cover art expands to fill the freed space instead of leaving a gap.
- **Yellow Pages**: `DrawIntent`'s header (the screen you land on from Go Somewhere/Hire Someone/Join Something) now reserves space for the scope pill instead of using a plain centered title. `DrawRailCard` (the "Open Now" rail) converted to `Marquee`. The intent tile's category count no longer goes through `Typography.DrawCentered`'s auto-wrap path.
- **Linkshells**: `LinkshellRow`'s title width reservation only accounted for the mute-bell icon, never the timestamp drawn on the same row; now reserves the timestamp's measured width first.
- **Collections**: the per-tile completion-percentage ring is slightly larger and better-proportioned relative to its icon (radius 23 to 25px, icon gap 18 to 8px).
- **Games (Tetris, Watersort, Breakout)**: added `GameHud.PillWidth` (and a `sizeScale` parameter on `Pill`/`ScorePill`) so HUD layout can be computed from a pill's real measured width instead of guessed fixed offsets, which is what was causing Score/Level/Best to overlap the Hold/Next preview slots and each other. Watersort and Breakout's HUD rows now solve their layout as one group (fixed anchors define a middle zone, flexible elements fit inside it) rather than two independent one-sided clamps that could fight each other. Tetris's HUD went through several rounds of layout iteration; final shape is Hold/Next in the corners (fixed size, since their piece-preview rendering assumes a fixed slot size), Restart in the gap between them, and Level/Score/Best as their own row below, spanning the full board width for maximum room against large scores. Breakout's top bar was changed to match Beat's shape (Restart pinned to a corner independent of the score pills, lives shown as a vertical 3-slot indicator instead of a horizontal row) per direct request.
- **`HoverTooltip`**: wrap-width calculation now uses the viewport's bounds instead of the current ImGui window's content region, since tooltips render on the foreground draw list and can extend past a small host window.
- **`StepperField`**: chevron button width now clamps to a fraction of the actual rect width instead of a fixed constant, so it can't exceed the available space on a narrow rect.
- **`AppHeader.DrawTitleWithReserve`**: added an optional `TextStyle` parameter so callers with a non-default title size (Photos) can use the shared helper instead of duplicating its logic inline.

## Test plan

- `dotnet build` clean after every change in this batch
- Each fix confirmed in-game via full plugin reload (not hot-reload) before moving to the next, including several rounds of screenshot-driven iteration on the Tetris HUD and the Yellow Pages venue-ad detail page
- No changes to game logic, only rendering/layout


BluntEXE:
Closing in favor of #71 — this branch (fix/home-screen-ui) had unrelated encryption/Wine-Proton commits mixed into its history that were never their own PR, and was also 21 commits behind current master. #71 is the same fix cherry-picked onto a clean branch off current master.

PR #71, opened 2026-07-27
Fix text truncation, hover-scroll, and layout overlap across the app

BluntEXE:
## Summary

Fixes a batch of text truncation, hover-scroll, and layout overlap bugs reported across the app (Photos, Notes, Timers, Dailies, Activity, Fishing, Market, Clock, Aethergram, Velvet, Music, Health, Yellow Pages, Muster, Direct Messages, Linkshells, Collections, and the three arcade mini-games), plus a couple of framework-level bugs found along the way that were causing some of them.

## Root causes found

Most of these bugs traced back to a small number of repeating patterns rather than being independent one-offs:

- **Shared hover state on stacked text.** A single hover boolean, computed for a whole row, was passed into two separate `Marquee` calls (title + subtitle). Hovering either line scrolled both. Fixed by giving each line its own hover check, using `Marquee.DrawLeftAuto`/`DrawCenteredAuto` where possible.
- **Unclamped sibling eating shared width.** A row split between two elements (e.g. a name and a value, or two pills) where one side was measured and clamped but the other was drawn at full natural width. When the unclamped side grew, it pushed into or past its sibling. Fixed by giving both sides a computed floor and letting them shrink or scroll together, following the same pattern used in `MarketApp.Detail.cs`'s hero card fix.
- **`Typography.DrawCentered`'s auto-wrap heuristic.** The no-`maxWidth` overload estimates a safe wrap width from the *current ImGui window's* edges relative to the text's center point. That assumption breaks for anything not roughly centered in a full-size window: a small host window (the minimized-phone icon), or an element sitting near the edge of a multi-column layout. Found three independent instances of this (Yellow Pages intent-tile category count, the minimized-phone notification badge, hover tooltips in general). Fixed each call site by bypassing the auto-wrap path with an explicit-width `Marquee` call or a manual measure-and-draw, rather than patching the heuristic itself (too many other call sites depend on its current behavior to change it globally without a full audit).
- **Padding equal to, not greater than, corner radius.** A pill button's text padding must exceed its corner radius (`height * 0.5`), not just be some positive number, or the text visually crowds the curve.

## Per-area changes

- **Photos, Notes, Timers, Dailies, Activity, Fishing, Market, Clock, Aethergram, Velvet, Health, Yellow Pages, Muster, Direct Messages**: converted static `Typography.FitText` truncation to `Marquee` hover-scroll wherever text could realistically overflow, applying the shared-hover and unclamped-sibling fixes above where relevant. This includes Muster's host name/status/meta text (`MusterApp.Detail.cs`, `.Directory.cs`, `.Manage.cs`, `MusterCard.cs`), the DM/Aethergram thread header name and subtitle (`MessageApp.Thread.cs`, `AethergramApp.Thread.cs`), and a Velvet comment timestamp that could overlap a review badge.
- **Aethergram**: the multi-photo page-indicator dots on both the feed card and detail page now only render at phone size M and up; below that there isn't room to show them without colliding with the share button, so they're hidden entirely instead of squeezed in.
- **Music**: Recently Played chip title/author text is hidden below phone size M for the same reason; the cover art expands to fill the freed space instead of leaving a gap.
- **Yellow Pages**: `DrawIntent`'s header (the screen you land on from Go Somewhere/Hire Someone/Join Something) now reserves space for the scope pill instead of using a plain centered title. `DrawRailCard` (the "Open Now" rail) converted to `Marquee`. The intent tile's category count no longer goes through `Typography.DrawCentered`'s auto-wrap path.
- **Linkshells**: `LinkshellRow`'s title width reservation only accounted for the mute-bell icon, never the timestamp drawn on the same row; now reserves the timestamp's measured width first.
- **Collections**: the per-tile completion-percentage ring is slightly larger and better-proportioned relative to its icon (radius 23 to 25px, icon gap 18 to 8px).
- **Games (Tetris, Watersort, Breakout)**: added `GameHud.PillWidth` (and a `sizeScale` parameter on `Pill`/`ScorePill`) so HUD layout can be computed from a pill's real measured width instead of guessed fixed offsets, which is what was causing Score/Level/Best to overlap the Hold/Next preview slots and each other. Watersort and Breakout's HUD rows now solve their layout as one group (fixed anchors define a middle zone, flexible elements fit inside it) rather than two independent one-sided clamps that could fight each other. Tetris's HUD went through several rounds of layout iteration; final shape is Hold/Next in the corners (fixed size, since their piece-preview rendering assumes a fixed slot size), Restart in the gap between them, and Level/Score/Best as their own row below, spanning the full board width for maximum room against large scores. Breakout's top bar was changed to match Beat's shape (Restart pinned to a corner independent of the score pills, lives shown as a vertical 3-slot indicator instead of a horizontal row) per direct request.
- **`HoverTooltip`**: wrap-width calculation now uses the viewport's bounds instead of the current ImGui window's content region, since tooltips render on the foreground draw list and can extend past a small host window.
- **`StepperField`**: chevron button width now clamps to a fraction of the actual rect width instead of a fixed constant, so it can't exceed the available space on a narrow rect.
- **`AppHeader.DrawTitleWithReserve`**: added an optional `TextStyle` parameter so callers with a non-default title size (Photos) can use the shared helper instead of duplicating its logic inline.

## Test plan

- `dotnet build` clean after every change in this batch
- Each fix confirmed in-game via full plugin reload (not hot-reload) before moving to the next, including several rounds of screenshot-driven iteration on the Tetris HUD and the Yellow Pages venue-ad detail page
- No changes to game logic, only rendering/layout


src/Aetherphone/Windows/Widgets/WidgetChrome.cs
```diff
@@ -45,6 +45,53 @@ public static float EyebrowWidth(string text, float scale)
...
+    private static float MeasureTrackedWidth(string text, float fontScale, FontWeight weight, float tracking)
+    {
+        using (Plugin.Fonts.Push(fontScale, weight))
+        {
+            var width = 0f;
+            Span<char> buffer = stackalloc char[1];
+            for (var index = 0; index < text.Length; index++)
+            {
+                buffer[0] = text[index];
+                width += ImGui.CalcTextSize(new string(buffer)).X + tracking;
```
XeldarAlz: Blocking: this adds `tracking` after the last character too, so `fullWidth` is always one tracking step wider than `EyebrowWidth` - which is exactly what `ResetsWidget.DrawColumn` passes as `maxWidth`. A label that fits therefore never takes the plain path: it falls into the `FitText` branch on a budget recomputed as `(a + b) - b` in float, so it can drop the last character to an ellipsis, and it marquee-scrolls ~1.6px on hover. Drop the trailing tracking, or just reuse `EyebrowWidth` here.

src/Aetherphone/Apps/YellowPages/YellowPagesApp.Detail.cs
```diff
@@ -540,22 +541,25 @@ private void DrawTagRow(AdDto ad, float scale)
...
                 lineTop += rowHeight + gap;
             }
 
-            var pillTop = new Vector2(cursorX, lineTop);
-            DrawPill(drawList, pillTop, label, fill, Palette.WithAlpha(ui.Accent, 0.92f), TextStyles.Footnote, scale);
-            Squircle.Stroke(drawList, pillTop, pillTop + new Vector2(pillWidth, pillHeight), pillHeight * 0.5f,
-                ImGui.GetColorU32(Palette.WithAlpha(ui.Accent, 0.30f)), 1f);
+            var pillMin = new Vector2(cursorX, lineTop);
+            var pillMax = pillMin + new Vector2(pillWidth, labelSize.Y + 8f * scale);
+            Squircle.Fill(drawList, pillMin, pillMax, (pillMax.Y - pillMin.Y) * 0.5f, ImGui.GetColorU32(fill));
```
XeldarAlz: Non-blocking: the old tag pill also drew `Squircle.Stroke` at accent 0.30 over the fill. That outline is gone now and it is not called out in the description. Separately, this inlines `DrawPill`'s 9/4/18/8 geometry rather than reusing it, so the two pill styles will drift.

src/Aetherphone/Apps/YellowPages/YellowPagesApp.Browse.cs
```diff
@@ -390,11 +401,13 @@ private bool DrawIntentTile(ImDrawListPtr drawList, int intent, Vector2 min, Vec
             ImGui.GetColorU32(Palette.WithAlpha(ui.Accent, 0.18f)), 32);
         AppSkin.Icon(drawList, glyphCenter, IntentIcon(intent).ToIconString(), ui.Accent, 0.92f);
         var textWidth = tileWidth - 12f * scale;
-        Typography.DrawWrappedCentered(drawList, Loc.T(AdIntents.Label(intent)), TextStyles.FootnoteEmphasized,
-            AppPalettes.YellowPages.TitleInk, new Vector2(glyphCenter.X, min.Y + 50f * scale), textWidth);
+        var label = Loc.T(AdIntents.Label(intent));
+        Typography.DrawWrappedCentered(drawList, label, labelStyle, AppPalettes.YellowPages.TitleInk,
```
XeldarAlz, line 407: Non-blocking: the label style moved from `FootnoteEmphasized` to `SubheadlineEmphasized` (0.80 to 0.88) and the count below from `Caption1` to `Footnote` (0.72 to 0.80). Neither bump is in the description, and the larger label is what makes `FitLabelStyle` necessary at all. Intended?

src/Aetherphone/Apps/YellowPages/YellowPagesApp.Browse.cs
```diff
@@ -405,6 +418,29 @@ private bool DrawIntentTile(ImDrawListPtr drawList, int intent, Vector2 min, Vec
         return UiInteract.Click(min, max, hovered);
     }
 
+    private static TextStyle FitLabelStyle(string label, float maxWidth, in TextStyle style)
+    {
+        var words = label.Split(' ');
```
XeldarAlz: Non-blocking: `Split(' ')` allocates an array plus a string per word, and `DrawIntentTiles` calls this once per intent every frame. `Typography.FitScale` underneath is cached, this is not. Intent labels are static per language, so cache the resolved style instead of re-splitting in the draw path.

src/Aetherphone/Windows/Components/HoverTooltip.cs
```diff
@@ -110,9 +110,9 @@ private static void DrawLabel(ImDrawListPtr dl, Rect target, string label, float
         var padX = 11f * scale;
         var padY = 7f * scale;
         var margin = 8f * scale;
-        var windowLeft = ImGui.GetWindowPos().X;
-        var left = windowLeft + ImGui.GetWindowContentRegionMin().X + margin;
-        var right = windowLeft + ImGui.GetWindowContentRegionMax().X - margin;
+        var viewport = ImGui.GetMainViewport();
+        var left = viewport.Pos.X + margin;
+        var right = viewport.Pos.X + viewport.Size.X - margin;
```
XeldarAlz, line 115: Nit: `maxTextWidth` is now the full viewport width, so a long label produces a tooltip spanning the whole screen instead of wrapping. Viewport bounds are right for the clamp; consider capping the wrap width, e.g. `MathF.Min(right - left, 420f * scale)`.

XeldarAlz (changes requested):
Verified on `39e06bc` in a throwaway worktree: `dotnet build` clean (0 errors, 4 pre-existing NU1902 warnings), `dotnet test` 139/139 passing. Not run in-game here, so the screenshot iteration is taken on trust.

Approach is right: fixing `GameHud.PillWidth`, `HoverTooltip` and `AppHeader.DrawTitleWithReserve` at the component level instead of patching every call site is the correct call. Five inline notes, one blocking.

src/Aetherphone/Windows/Widgets/WidgetChrome.cs
BluntEXE: Fixed in ed685cc5. `MeasureTrackedWidth` was adding `tracking` after every character including the last; it now matches `EyebrowWidth`'s formula (tracking only between characters), so `fullWidth` no longer disagrees with the `maxWidth` callers pass in.

src/Aetherphone/Apps/YellowPages/YellowPagesApp.Detail.cs
BluntEXE: Fixed in ed685cc5 — restored the `Squircle.Stroke` at accent 0.30 on the tag pill fill. Left the geometry inlined rather than routing through `DrawPill` since that helper doesn't support the marquee-fit label this pill needs; happy to extract a shared constants/geometry helper separately if you'd rather not carry the duplication.

src/Aetherphone/Apps/YellowPages/YellowPagesApp.Browse.cs
BluntEXE: Intentional, not called out in the description — my bad. The bump from FootnoteEmphasized/Caption1 to SubheadlineEmphasized/Footnote is what makes `FitLabelStyle` worth having; at the old sizes labels basically never overflowed the tile width, so the fit-down path would rarely fire. Left as-is.

src/Aetherphone/Apps/YellowPages/YellowPagesApp.Browse.cs
BluntEXE: Fixed in ed685cc5. Added a small `IntentLabelWords` cache keyed by label so `Split(' ')` runs once per distinct label instead of once per tile per frame.

src/Aetherphone/Windows/Components/HoverTooltip.cs
BluntEXE: Fixed in ed685cc5 — capped `maxTextWidth` at `MathF.Min(right - left, 420f * scale)` as suggested.

XeldarAlz (commented):
Reviewed all 41 files. Verified: built `ed685cc5` clean in a detached worktree (0 errors, only the pre-existing AngleSharp NU1902 warnings).

The three root causes in the description hold up, and the fixes are the right shape: per-line hover instead of one shared boolean, both sides of a row given a floor, and the `Typography.DrawCentered` auto-wrap path bypassed at the call site rather than patched globally. Checked and happy with: the `GameHud` `sizeScale` plumbing (measure and draw both land on the same `FontService.NearestSize` bucket, so `PillWidth` tracks the rendered text), the `HoverTooltip` viewport switch, the Market hero card reflow, and the Linkshell timestamp reserve.

2 blocking (both in `WidgetChrome.cs`, both small), 6 non-blocking. Details inline.

One housekeeping note: `ed685cc5` carries a `Co-Authored-By: Claude Sonnet 5` trailer. I keep those out of this repo's history, so please drop it before merge (or I'll strip it on squash).

src/Aetherphone/Windows/Widgets/WidgetChrome.cs
```diff
@@ -45,6 +45,55 @@ public static float EyebrowWidth(string text, float scale)
...
+
+        var offset = Marquee.Offset(id, fullWidth - maxWidth);
+        var slack = 4f * scale;
+        drawList.PushClipRect(new Vector2(position.X, position.Y - slack), new Vector2(position.X + maxWidth, position.Y + height + slack),
+            true);
+        Tracked(drawList, position with { X = position.X - offset }, upper, tinted, EyebrowFontScale, FontWeight.SemiBold, tracking);
+        drawList.PopClipRect();
+    }
+
+    private static float MeasureTrackedWidth(string text, float fontScale, FontWeight weight, float tracking)
```
XeldarAlz: **Blocking.** This is a second formula for a width `EyebrowWidth` (line 40) already computes, and the two don't agree: `EyebrowWidth` measures the whole string once and adds `tracking * (n-1)`, this sums per-glyph `CalcTextSize` and adds the same tracking. ImGui rounds each `CalcTextSize` up, so the per-glyph sum is >= the whole-string measure.

`ResetsWidget.cs:70` passes `EyebrowWidth` in as `maxWidth`, so `fullWidth <= maxWidth` on line 55 is effectively never true: a label that fits still takes the clip/marquee path and shimmers a few pixels on hover.

Use `EyebrowWidth(text, scale)` here and delete `MeasureTrackedWidth`. One formula, agrees with the caller by construction, and it drops a `new string(buffer)` allocation per character per frame.

src/Aetherphone/Windows/Widgets/WidgetChrome.cs
```diff
@@ -45,6 +45,55 @@ public static float EyebrowWidth(string text, float scale)
...
+    {
+        using (Plugin.Fonts.Push(fontScale, weight))
+        {
+            var width = 0f;
+            Span<char> buffer = stackalloc char[1];
+            for (var index = 0; index < text.Length; index++)
+            {
+                buffer[0] = text[index];
+                width += ImGui.CalcTextSize(new string(buffer)).X;
+                if (index < text.Length - 1)
```
XeldarAlz: **Blocking (style).** Braces on every single-line `if` body, per the repo rule.

src/Aetherphone/Apps/Notes/NotesApp.cs
```diff
@@ -266,16 +266,15 @@ private void DrawReminderRow(Rect row, ReminderItem reminder, float scale)
         var hasDue = reminder.DueAt.HasValue;
         var titleY = hasDue ? row.Center.Y - 16f * scale : row.Center.Y - 9f * scale;
         var title = reminder.Title.Length > 0 ? reminder.Title : Loc.T(L.Notes.ReminderHint);
-        var rowHovered = UiInteract.Hover(textRect.Min, textRect.Max);
-        Marquee.DrawLeft("notes.reminderRow.title." + reminder.Id, title, textLeft, titleY, textRect.Width,
-            TextStyles.Body, titleInk, rowHovered);
+        Marquee.DrawLeftAuto("notes.reminderRow.title." + reminder.Id, title, textLeft, titleY, textRect.Width,
```
XeldarAlz, line 269: Non-blocking. This dropped the `UiInteract.Hover` call, and `DrawLeftAuto` hover-tests with a raw `ImGui.IsMouseHoveringRect` - so the title now marquees while an overlay or open menu sits on top of the row. `UiInteract.Hover` gates on `InputBlocked` and `MouseOverOverlay`; the raw call doesn't.

Notes was the only site in this PR that had the gating (Dailies and Timers were already raw). Either keep `DrawLeft` + `UiInteract.Hover` here, or point the three `*Auto` helpers in `Marquee.cs` at `UiInteract.Hover`, which fixes every Auto call site at once.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
```diff
@@ -107,8 +107,10 @@ private void DrawCustomAlbumView(Rect area, int key)
             return;
         }
 
-        DrawNavBar(area, album.Name, back);
-        if (ui.HeaderAction(area, Loc.T(L.Photos.AddPhotos), entries.Length > 0))
+        var addPhotosLabel = Loc.T(L.Photos.AddPhotos);
+        var addPhotosWidth = Typography.Measure(addPhotosLabel, 0.9f, FontWeight.SemiBold).X + 28f * scale + 6f * scale;
```
XeldarAlz: Non-blocking. `0.9f`/`SemiBold`, `+ 28f` (the button height) and `+ 6f` are `AppSkin.HeaderAction`'s internal geometry (`AppSkin.cs:391-393`) re-derived here. Change the button's height or padding and this reserve silently drifts, sliding the title back under it.

Expose `AppSkin.HeaderActionWidth(label)` and call it from both.

src/Aetherphone/Apps/Music/MusicApp.Home.cs
```diff
@@ -142,23 +142,28 @@ private void DrawRecentChips(float scale, float available)
                 ? Palette.WithAlpha(ui.Accent, 0.16f)
                 : Palette.WithAlpha(ui.TitleInk, hovered ? 0.13f : 0.07f);
             Squircle.Fill(drawList, min, max, rounding, ImGui.GetColorU32(fill));
-            var artMax = new Vector2(min.X + chipHeight, max.Y);
+            var showText = configuration.PhoneScale >= PhoneSizeCatalog.DefaultScale;
```
XeldarAlz: Non-blocking, product call. Below M the chip loses title *and* author, so Recently Played is unlabelled cover art on XS and S - two of the six sizes. Hiding the Aethergram page dots is fine, they're decoration; here the text is the chip's only content.

Worth trying title-only on one line at `Caption1` before hiding both.

src/Aetherphone/Apps/Games/Breakout/BreakoutApp.cs
```diff
@@ -240,13 +248,22 @@ private void DrawLives(Rect body, float rowY, float scale)
     {
         var drawList = ImGui.GetWindowDrawList();
         var lastLife = board.Lives == 1;
-        for (var index = 0; index < board.Lives; index++)
+        var radius = 4f * scale;
+        var origin = new Vector2(body.Min.X + 22f * scale, rowY);
+        for (var life = 0; life < 3; life++)
```
XeldarAlz: Non-blocking. `3` is `BreakoutBoard.cs:82`'s starting `Lives` duplicated - a balance change there silently loses a slot here. `const int StartingLives = 3` on the board, read by both.

src/Aetherphone/Apps/Velvet/VelvetShell.Profile.cs
```diff
@@ -81,7 +81,10 @@ private void DrawProfileBody(Rect body, VelvetProfileDto user)
             var displayName = Typography.FitText(name, textWidth - badgeSpace, TextStyles.Title1);
             var nameSize = Typography.Measure(displayName, TextStyles.Title1);
             var nameX = centerX - (nameSize.X + badgeSpace) * 0.5f;
-            Typography.Draw(new Vector2(nameX, lineTop), displayName, VelvetTheme.TitleInk, TextStyles.Title1);
+            var nameHovering = ImGui.IsMouseHoveringRect(new Vector2(nameX, lineTop),
```
XeldarAlz: Non-blocking. `displayName` is now only used to measure, and `nameSize.X` - the marquee box width - is the width of the *ellipsised* string, so the box carries the `…` glyph's width.

`MathF.Min(Typography.Measure(name, TextStyles.Title1).X, textWidth - badgeSpace)` states it directly and drops the throwaway `FitText`.

src/Aetherphone/Apps/YellowPages/YellowPagesApp.Browse.cs
```diff
@@ -405,6 +418,42 @@ private bool DrawIntentTile(ImDrawListPtr drawList, int intent, Vector2 min, Vec
         return UiInteract.Click(min, max, hovered);
     }
 
+    private static readonly Dictionary<string, string[]> IntentLabelWords = new();
```
XeldarAlz: Nit: static field wedged between two methods - move it up with the type's other statics.

Also worth asking whether a never-cleared cache keyed by localized text earns its keep against three two-word `Split`s per frame. `FitLabelStyle`'s per-word `Typography.Measure` calls dominate the cost either way, and those aren't cached.

src/Aetherphone/Windows/Widgets/WidgetChrome.cs
BluntEXE: Fixed: EyebrowMarquee now uses EyebrowWidth for fullWidth, matching the formula the callers pass as maxWidth. Deleted MeasureTrackedWidth.

src/Aetherphone/Windows/Widgets/WidgetChrome.cs
BluntEXE: Fixed: braced.

src/Aetherphone/Apps/Notes/NotesApp.cs
BluntEXE: Fixed: pointed Marquee's DrawLeftAuto/DrawRightAuto/DrawCenteredAuto at UiInteract.Hover, fixes every *Auto call site (Notes included) at once.

src/Aetherphone/Apps/Photos/PhotosApp.Grid.cs
BluntEXE: Fixed: extracted AppSkin.HeaderActionWidth(label), both call sites now share it.

src/Aetherphone/Apps/Music/MusicApp.Home.cs
BluntEXE: Went with title-only at Caption1 below M, matching your suggestion. Author still hides below M.

src/Aetherphone/Apps/Games/Breakout/BreakoutApp.cs
BluntEXE: Fixed: added BreakoutBoard.StartingLives, both sites read it.

src/Aetherphone/Apps/Velvet/VelvetShell.Profile.cs
BluntEXE: Fixed: nameSize.X now MathF.Min(Typography.Measure(name, TextStyles.Title1).X, textWidth - badgeSpace) per your suggestion, dropped the FitText call.

src/Aetherphone/Apps/YellowPages/YellowPagesApp.Browse.cs
BluntEXE: Moved IntentLabelWords up with the other statics. Left the cache as-is, the per-word Measure calls dominate cost either way.

XeldarAlz:
Pushed a merge of master onto this branch (`5a116ac7`) since merging #73 put it in conflict. Verified: Release build clean (0 errors), `dotnet test` 158/158.

Conflict resolutions, all "keep both":
- `AppHeader.DrawTitleWithReserve` carries both new optional parameters: `style` from here, `leftReserve` from #73. Order is `style` then `leftReserve`, so this branch's positional callers are unaffected.
- `LinkshellRow.cs:75` keeps the `- timeReserve` subtraction on top of master's `UiInteract.Hover`.
- `DailiesApp.cs`, `TimersApp.cs`, `MessageApp.Thread.cs`: dropped master's row hover locals. Those rows moved to `Marquee.DrawLeftAuto`, which hit-tests internally, so the locals were dead, and Message's outer one shadowed an inner declaration.
- `TetrisApp.cs:117` keeps the `outerMargin` extraction with master's gated hover.
- `MusicApp.Home.cs:151` keeps the `showAuthor` branch with master's gated hover.

One change beyond the conflicts: this branch predates #74, so it reintroduced 15 raw `ImGui.IsMouseHoveringRect` hit tests (Aethergram/Message thread headers, Music chips, Velvet profile, WidgetChrome, LinkshellRow, Tetris). Master now holds the invariant that the call only appears inside `UiInteract.cs`; leaving them raw would have regressed the occlusion fix for every text row this branch touches, so they are all `UiInteract.Hover` now.

Threads are all resolved and it is mergeable again. Pull before you push anything further.

PR #72, opened 2026-07-28
fix(media): fix stray pixels baked into uploaded photos

YozoraCho:
## What

`ImageProcessor.DecodeToTextureAsync` used to rent its pixel buffer from `ArrayPool<byte>.Shared`. Now it just allocates a plain array. That also lets the `try`/`finally` and the separate `length` variable go away.

## Why

This is the "weird pixels appearing on upload" report that's been open since 0.9.8.7. Uploaded photos come out with small coloured specks in them. The source file on the poster's PC is clean, and the Photos app shows it fine, so it only ever showed up after posting.

First thing worth ruling out: it isn't the CDN and it isn't JPEG quality. I pulled the served bytes straight out of the plugin's own image cache and the specks are already in there, so the damage happens on the poster's machine before anything gets uploaded.

Then I measured every speck across two separate corrupted images, a 1080x1080 gram and a 512x512 avatar:

| Property | Value | Runs matching |
| --- | --- | --- |
| Run width | exactly 4 px | 38 / 38 |
| Run height | 1 px | 38 / 38 |
| `x mod 4` | 0 | 38 / 38 |
| `y mod 8` | 0 | 38 / 38 |
| Byte offset in RGBA buffer `mod 16` | 0 | 38 / 38 |
| Byte offset `mod 32` | 16 | 38 / 38 |

That's what gives it away. 4 px at 4 bytes each is 16 bytes, 16-byte aligned, which is one `Vector128<byte>` store. Photos don't line up on SIMD boundaries. Something writing to the wrong address does, every time. The 8-row spacing fits too, since that's the JPEG encoder's strip height.

So here's what I think is happening. `ArrayPool<byte>.Shared` is shared across the whole process, and ImageSharp pulls its JPEG encoder scratch buffers from that same pool. Our decode path handed its buffer back the moment `CreateFromRawAsync`'s task finished, but that doesn't actually tell us the pixels have been consumed, and Dalamud doesn't give us anything that does. If the buffer goes back even slightly early, `CopyPixelDataTo` can be writing one image into it while an unrelated photo is being encoded out of it. You end up with 16-byte pieces of one picture baked into another picture's upload.

That lines up with how people described it, too: fine on a quiet session, broke after relogging when the whole feed was loading at once, survived a clean reinstall, and didn't happen to everyone.

The cost is one allocation per decode. Callers cache the decoded texture so this doesn't run often, and the array is cheap next to the decode itself. There's no safe way to keep pooling here without a completion signal we don't have, so I left a comment on the method explaining that, otherwise someone will reasonably try to optimise it back.

## How to test

It's a race, so posting one photo on an idle client and seeing it come out clean doesn't really tell you anything. You want load.

1. `dotnet build -c Release`
2. Open Aethergram and scroll a busy feed so a batch of images is decoding.
3. While that's still loading, post a photo.
4. Check the uploaded file, not what's on screen. Either open the post's image
   URL in a browser, or grab the newest files out of
   `pluginConfigs/Aetherphone/cache/images/`. Those are the served bytes. They
   have no file extension, so copy one and rename it to `.jpg` to open it.
5. Zoom in and look for 4 px wide, 1 px tall horizontal dashes sitting on
   smooth areas, in colours that have nothing to do with the photo. If you see
   several at the same x but different y, that's the one.

Worth doing a few times, since before this change it only showed up now and then.

Also worth a quick regression pass, because every image in the app goes through this method: feed photos, avatars, stories, DM images, Yellow Pages ads and the Photos gallery should all still load normally.

## Checklist

- [x] If this changes user-visible behavior, README is updated (n/a, bug fix only)


XeldarAlz (changes requested):
`dotnet build -c Release` passes on aeaf536, no new warnings. Not run in-game.

The change itself is safe, but the diagnosis doesn't hold and merging as-is would close a bug that is still live. Findings inline.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -68,26 +67,31 @@ public static BakedImage BakeJpeg(string sourcePath, int maxDimension)
         return new BakedImage(stream.ToArray(), width, height);
```
XeldarAlz, line 96: Blocking: these are the bytes that get uploaded (`BakeCroppedJpeg` -> `SaveAsJpeg` -> `ToArray`), and they never pass through `DecodeToTextureAsync`. You confirmed the specks are already in the cached served bytes, so the corruption happens on this path, not the one this PR changes.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -68,26 +67,31 @@ public static BakedImage BakeJpeg(string sourcePath, int maxDimension)
...
     }
```
XeldarAlz, line 97: Next step, deterministic and offline: hash `BakeCroppedJpeg(source, crop, 1080, 1080)` output in a threaded loop over the poster's file. Differing hashes localise it to bake/ImageSharp; identical and clean clears bake and points at `UploadImageAsync`.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -68,26 +67,31 @@ public static BakedImage BakeJpeg(string sourcePath, int maxDimension)
...
 
+    /// Decodes <paramref name="bytes"/> and uploads the pixels as a texture.
```
XeldarAlz: Non-blocking: keep the code change, drop this block. The stated rationale isn't accurate and it tells the next maintainer not to touch it. Style too: the repo's other 232 `///` comments use `<summary>` (see `Core/Muster/MusterShare.cs:3`).

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -68,26 +67,31 @@ public static BakedImage BakeJpeg(string sourcePath, int maxDimension)
...
+    ///
+    /// The pixel buffer is a dedicated array on purpose. Do not switch it to a
+    /// pooled one: the shared array pool is process-wide, and the JPEG encoder
```
XeldarAlz: Blocking: ImageSharp 3.x doesn't use `ArrayPool<byte>.Shared`. Its default allocator is `UniformUnmanagedMemoryPoolMemoryAllocator`, unmanaged blocks. The shipped 3.1.12 XML docs have 5 mentions of `UniformUnmanagedMemoryPool` and 0 of `ArrayPool`. No shared pool to race on.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -68,26 +67,31 @@ public static BakedImage BakeJpeg(string sourcePath, int maxDimension)
...
     {
-        var (pixels, length, width, height) = await Task.Run(() =>
+        var (pixels, width, height) = await Task.Run(() =>
         {
             using var image = Image.Load<Rgba32>(bytes);
-            var length = image.Width * image.Height * 4;
-            var buffer = ArrayPool<byte>.Shared.Rent(length);
-            image.CopyPixelDataTo(buffer.AsSpan(0, length));
-            return (buffer, length, image.Width, image.Height);
+            var buffer = new byte[image.Width * image.Height * 4];
```
XeldarAlz: Blocking: `ArrayPool<byte>.Shared` doesn't pool above 1 MB. A 1080x1080 RGBA buffer is 4,665,600 bytes, so `Rent` already allocated a fresh array and `Return` dropped it. This diff is a no-op for grams, stories and 1280px DMs, which rules the pool out for your 1080x1080 sample.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -68,26 +67,31 @@ public static BakedImage BakeJpeg(string sourcePath, int maxDimension)
...
+            image.CopyPixelDataTo(buffer);
```
XeldarAlz: Blocking: the old code could never stomp another renter. `CopyPixelDataTo` was the only write and ran before the array left the method; `Return` ran after the await. The reverse hazard, Dalamud reading after the task completes, corrupts the on-screen texture, not the uploaded file.

src/Aetherphone/Core/Media/ImageProcessor.cs
```diff
@@ -68,26 +67,31 @@ public static BakedImage BakeJpeg(string sourcePath, int maxDimension)
...
+            return (buffer, image.Width, image.Height);
```
XeldarAlz: Non-blocking: the two paths that were genuinely pooled do regress. `PhotosApp.cs:660` (256px thumbs) and `AvatarUpload.cs:40` (512x512, exactly 1 MB, the last poolable bucket). Gallery scrolling now allocates 256 KB per thumbnail decode. Survivable given caching, just naming it.

Syrilai:
oh yeah lgtm 👍

YozoraCho:
I hate github and the way it handles shit

Syrilai:
i mean next time might as well just commit every single existing commit and not just a small subset 👍

PR #73, opened 2026-07-28
fix(jobs): reserve header space and marquee row text

BluntEXE:
## Summary

Jobs was the one app left off #71's text-truncation/marquee sweep. Two spots still used static clipping instead of the shared reserve/marquee pattern:

- **Header title**: used plain `Typography.DrawCentered` across the full header width, ignoring the palette + categories icon buttons in the corner. Long localized titles (e.g. Spanish "Trabajos", Turkish "Sınıflar") could run into the buttons. Switched to `AppHeader.DrawTitleWithReserve`, same helper used by Feedback, Market, Venues, Chirper, Aethergram, and Velvet.
- **Gearset rows**: name and level/ilvl subtitle used static `Typography.FitText`, silently clipping to an ellipsis with no way to read the full text. Switched each line to its own `Marquee.DrawLeftAuto` call (separate ids/hover checks per line, following the pattern in #71's `TimersApp.DrawTimerRow` and `NotesApp.DrawReminderRow`), so long names now scroll on hover instead of clipping.

## Test plan

- `dotnet build` clean
- Verified in-game via full plugin reload: header title no longer overlaps the button cluster, row name/subtitle scroll independently on hover

### DIFF
--- a/src/Aetherphone/Apps/Jobs/JobsApp.cs
+++ b/src/Aetherphone/Apps/Jobs/JobsApp.cs
@@ -217,7 +217,8 @@ internal sealed partial class JobsApp : IPhoneApp
         var categoriesCenter = new Vector2(buttonCenter.X - radius * 2f - 10f * scale, rowCenterY);
         var clusterLeftEdge = (hasCategories ? categoriesCenter.X : buttonCenter.X) - radius;
         var rightReserve = content.Max.X - clusterLeftEdge + 8f * scale;
-        AppHeader.DrawTitleWithReserve(content, "jobs.header.title", DisplayName, rightReserve, ui.TitleInk, scale);
+        AppHeader.DrawTitleWithReserve(content, "jobs.header.title", DisplayName, rightReserve, ui.TitleInk, scale,
+            leftReserve: 0f);
 
         colorButtonRect = new Rect(buttonCenter - new Vector2(radius, radius), buttonCenter + new Vector2(radius, radius));
         UiAnchors.Report("jobs.color", colorButtonRect);
@@ -386,13 +387,12 @@ internal sealed partial class JobsApp : IPhoneApp
         var textRight = noteRight -
                         (note.Length == 0 ? 0f : Typography.Measure(note, TextStyles.Caption2).X + 28f * scale);
         var maxTextWidth = MathF.Max(1f, textRight - textLeft);
-        var rowId = job.Kind == JobEntryKind.Gearset ? "jobs.row.gearset." + job.GearsetId : "jobs.row.class." + job.ClassJobId;
-        Marquee.DrawLeftAuto(drawList, rowId + ".name", job.Name, textLeft, contentRect.Min.Y + 12f * scale,
+        Marquee.DrawLeftAuto(drawList, job.NameRowId, job.Name, textLeft, contentRect.Min.Y + 12f * scale,
             maxTextWidth, TextStyles.Headline, ui.TitleInk);
         var subtitle = job.ItemLevel >= 0
             ? Loc.T(L.Jobs.LevelItemLevel, job.Abbreviation, job.Level, job.ItemLevel)
             : Loc.T(L.Jobs.LevelOnly, job.Abbreviation, job.Level);
-        Marquee.DrawLeftAuto(drawList, rowId + ".sub", subtitle, textLeft, contentRect.Min.Y + 36f * scale,
+        Marquee.DrawLeftAuto(drawList, job.SubRowId, subtitle, textLeft, contentRect.Min.Y + 36f * scale,
             maxTextWidth, TextStyles.Footnote, ui.MutedInk);
 
         var noteCenter = new Vector2(noteRight, contentRect.Center.Y);
--- a/src/Aetherphone/Apps/Jobs/JobsApp.cs
+++ b/src/Aetherphone/Apps/Jobs/JobsApp.cs
@@ -211,10 +211,14 @@ internal sealed partial class JobsApp : IPhoneApp
     private void DrawHeader(Rect content, float scale)
     {
         var rowCenterY = content.Min.Y + AppHeader.Height * scale * 0.5f;
-        Typography.DrawCentered(new Vector2(content.Center.X, rowCenterY), DisplayName, ui.TitleInk, 1.15f,
-            FontWeight.SemiBold);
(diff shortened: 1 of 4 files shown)


src/Aetherphone/Apps/Jobs/JobsApp.cs
```diff
@@ -211,10 +211,14 @@ public void Draw(in PhoneContext context)
...
         var rowCenterY = content.Min.Y + AppHeader.Height * scale * 0.5f;
-        Typography.DrawCentered(new Vector2(content.Center.X, rowCenterY), DisplayName, ui.TitleInk, 1.15f,
-            FontWeight.SemiBold);
         var radius = 15f * scale;
         var buttonCenter = new Vector2(content.Max.X - Metrics.Space.Lg * scale - radius, rowCenterY);
+        var hasCategories = gameData.LocalPlayer is not null;
+        var categoriesCenter = new Vector2(buttonCenter.X - radius * 2f - 10f * scale, rowCenterY);
+        var clusterLeftEdge = (hasCategories ? categoriesCenter.X : buttonCenter.X) - radius;
+        var rightReserve = content.Max.X - clusterLeftEdge + 8f * scale;
+        AppHeader.DrawTitleWithReserve(content, "jobs.header.title", DisplayName, rightReserve, ui.TitleInk, scale);
```
XeldarAlz: Blocking: `DrawTitleWithReserve` hardcodes a 44px left inset for the back button (`AppHeader.cs:44`), and all seven existing callers draw one via `AppHeader.Draw` first. Jobs has none, so the title centers in `[min+44, max-94]`: 25px left of the screen center at scale 1, and it jumps another 20px right when the categories button hides. Add a `leftReserve` parameter (default 44f) and pass 0 here. Same signature #71 is editing, so whichever lands second rebases.

src/Aetherphone/Apps/Jobs/JobsApp.cs
```diff
@@ -382,15 +385,15 @@ private void DrawJobRow(ImDrawListPtr drawList, Rect rowRect, Rect contentRect,
         var noteRight = hasMenu ? menuCenter.X - menuRadius - 8f * scale : contentRect.Max.X;
         var textRight = noteRight -
                         (note.Length == 0 ? 0f : Typography.Measure(note, TextStyles.Caption2).X + 28f * scale);
-        var name = Typography.FitText(job.Name, textRight - textLeft, TextStyles.Headline);
-        Typography.Draw(drawList, new Vector2(textLeft, contentRect.Min.Y + 12f * scale), name, ui.TitleInk,
-            TextStyles.Headline);
+        var maxTextWidth = MathF.Max(1f, textRight - textLeft);
+        var rowId = job.Kind == JobEntryKind.Gearset ? "jobs.row.gearset." + job.GearsetId : "jobs.row.class." + job.ClassJobId;
+        Marquee.DrawLeftAuto(drawList, rowId + ".name", job.Name, textLeft, contentRect.Min.Y + 12f * scale,
```
XeldarAlz: Non-blocking: same gating gap as my note on #71 `NotesApp.cs:269`. `DrawLeftAuto` hover-tests with a raw `ImGui.IsMouseHoveringRect`, so rows keep scrolling under the color picker and the category editor even though `Draw` calls `UiInteract.BlockThisFrame` (line 142), and under the three dropdowns. Jobs stacks more overlays over its rows than any other site, so it is the worst case for it. Cleanest fix is still pointing the three `*Auto` helpers at `UiInteract.Hover`.

src/Aetherphone/Apps/Jobs/JobsApp.cs
```diff
@@ -382,15 +385,15 @@ private void DrawJobRow(ImDrawListPtr drawList, Rect rowRect, Rect contentRect,
         var noteRight = hasMenu ? menuCenter.X - menuRadius - 8f * scale : contentRect.Max.X;
         var textRight = noteRight -
                         (note.Length == 0 ? 0f : Typography.Measure(note, TextStyles.Caption2).X + 28f * scale);
-        var name = Typography.FitText(job.Name, textRight - textLeft, TextStyles.Headline);
-        Typography.Draw(drawList, new Vector2(textLeft, contentRect.Min.Y + 12f * scale), name, ui.TitleInk,
-            TextStyles.Headline);
+        var maxTextWidth = MathF.Max(1f, textRight - textLeft);
+        var rowId = job.Kind == JobEntryKind.Gearset ? "jobs.row.gearset." + job.GearsetId : "jobs.row.class." + job.ClassJobId;
```
XeldarAlz: Nit: `rowId` plus the two concatenations allocate three strings per row per frame, and every row draws regardless of scroll, so a full gearset list churns a few hundred strings a frame. Building the id once in `JobsReader` and hanging it on `JobEntry` (rebuilt every 2s) drops that to zero.

XeldarAlz (changes requested):
Verified on `a4cc69c` in a throwaway worktree: Release build clean (0 errors, only the pre-existing AngleSharp NU1902 warnings), `dotnet test` 143/143 passing. Not run in-game here, so the visual check is taken on trust.

Routing Jobs through the shared helpers instead of hand-rolling a reserve is the right call, and the row ids are correctly unique (gearset ids and no-gearset class ids never collide across sections). One blocking note on where the title now lands, two smaller ones inline.

src/Aetherphone/Apps/Jobs/JobsApp.cs
BluntEXE: Fixed: DrawTitleWithReserve now takes an optional leftReserve (default 44f, keeps the 7 existing back-button callers unchanged). Jobs passes leftReserve: 0f.

src/Aetherphone/Apps/Jobs/JobsApp.cs
BluntEXE: Fixed: pointed Marquee's *Auto helpers at UiInteract.Hover, same root cause as your #71 NotesApp note. This branch predates that fix so applied it here directly.

src/Aetherphone/Apps/Jobs/JobsApp.cs
BluntEXE: Fixed: JobEntry now builds NameRowId/SubRowId once in its constructor (rebuilt every 2s refresh), JobsApp just reads them.

PR #74, opened 2026-07-28
fix(input): stop clicks passing through occluding windows

BluntEXE:
## Summary

The phone accepted clicks and hovers even when a different Dalamud window sat on top of it. Every row, button, and gesture used raw ImGui rect checks that never asked which window ImGui considers hovered this frame, so clicks passed straight through a plugin window drawn on top.

## Root cause

`UiInteract.Hover()` already existed as the shared hit-test primitive most of the app calls through, but it never checked window ownership. `PhoneWindow.Draw()` now records `ImGui.IsWindowHovered(ImGuiHoveredFlags.ChildWindows)` once per frame; `Hover()` and `Click()` both gate on that. `Click()` also re-checks it directly, so a call site with its own hover test still can't fire an action while another window occludes the phone.

The same gap showed up outside `UiInteract` in a few shared components that predate it or duplicate its logic locally: `Marquee`'s auto-hover variants, `DragTracker.Begin` (Control Center's swipe, Notification Center's swipe-to-dismiss), `HomeInteractionController`'s tap and magnify hover, and the Games framework's `RestartButton`/`Button`. Fixed once in each, which covers every caller.

## Per-area

- `UiInteract`: window-hover gate, a 3-arg `Hover` overload for the existing `clip` parameter, and `Click()` hardened as a backstop for callers with a raw hover test.
- `PhoneWindow`: records window-hover state once per frame before any hit test runs.
- Home screen: tile taps, magnify hover, page dots/arrows, edit-mode Add/Done, folder contents, widget gallery and size menu.
- Control Center: tile press/drag-start, swipe-open/dismiss, top-band cursor.
- Dynamic Island, camera controls, notification center and banners, chat bubbles/transcript/menu, roughly twenty mini-games' board input, and every remaining `Marquee`/hover call site in the app. Same pattern each time.
- `Games/Flow`: kept a plain bounds check for the in-progress drag path. Routing it through `UiInteract.Hover` let `DragScrollHost`'s scroll-vs-tap heuristic set `InputBlocked` mid-drag and freeze the pipe being drawn. Only the initiating press is gated now, matching the pattern already used for the home screen's drag continuation and Solitaire's card grab.

## Test plan

- `dotnet build` clean after each change
- Verified in-game via full plugin reload across Home, Control Center, Dynamic Island, Camera, Notifications, Message/social apps, and the affected mini-games, with a Dalamud window held on top

### DIFF
--- a/src/Aetherphone/Apps/Collections/CollectionsApp.Browse.cs
+++ b/src/Aetherphone/Apps/Collections/CollectionsApp.Browse.cs
@@ -443,13 +443,13 @@ internal sealed partial class CollectionsApp
         {
             var nameY = row.Center.Y - 16f * scale;
             var nameSize = Typography.Measure(item.Name, TextStyles.BodyEmphasized);
-            var nameHovering = ImGui.IsMouseHoveringRect(new Vector2(textLeft, nameY),
+            var nameHovering = UiInteract.Hover(new Vector2(textLeft, nameY),
                 new Vector2(textLeft + textWidth, nameY + nameSize.Y));
             Marquee.DrawLeft("collections.item." + item.Id, item.Name, textLeft, nameY,
                 textWidth, TextStyles.BodyEmphasized, ui.TitleInk, nameHovering);
             var subY = row.Center.Y + 4f * scale;
             var subSize = Typography.Measure(subtitle, TextStyles.Footnote);
-            var subHovering = ImGui.IsMouseHoveringRect(new Vector2(textLeft, subY),
+            var subHovering = UiInteract.Hover(new Vector2(textLeft, subY),
                 new Vector2(textLeft + textWidth, subY + subSize.Y));
             Marquee.DrawLeft("collections.item.sub." + item.Id, subtitle, textLeft, subY, textWidth,
                 TextStyles.Footnote, ui.MutedInk, subHovering);
@@ -458,7 +458,7 @@ internal sealed partial class CollectionsApp
         {
             var nameSize = Typography.Measure(item.Name, TextStyles.BodyEmphasized);
             var nameY = row.Center.Y - nameSize.Y * 0.5f;
-            var nameHovering = ImGui.IsMouseHoveringRect(new Vector2(textLeft, nameY),
+            var nameHovering = UiInteract.Hover(new Vector2(textLeft, nameY),
                 new Vector2(textLeft + textWidth, nameY + nameSize.Y));
             Marquee.DrawLeft("collections.item." + item.Id, item.Name, textLeft, nameY,
                 textWidth, TextStyles.BodyEmphasized, ui.TitleInk, nameHovering);
@@ -638,7 +638,7 @@ internal sealed partial class CollectionsApp
             var centerY = (rowMin.Y + rowMax.Y) * 0.5f;
             var label = index == 0 ? Loc.T(L.Collections.AllSources) : sourceList[index - 1];
             var selected = index == sourceIndex;
-            var hovered = ImGui.IsMouseHoveringRect(rowMin, rowMax);
+            var hovered = UiInteract.Hover(rowMin, rowMax);
             if (hovered)
             {
                 Squircle.Fill(drawList, rowMin, rowMax, 9f * scale, ImGui.GetColorU32(ui.HoverTint));
@@ -670,8 +670,8 @@ internal sealed partial class CollectionsApp
             return;
         }
 
-        if (ImGui.IsMouseClicked(ImGuiMouseButton.Left) && !ImGui.IsMouseHoveringRect(min, max, false) &&
-            !ImGui.IsMouseHoveringRect(sourceMenuAnchor.Min, sourceMenuAnchor.Max, false))
+        if (ImGui.IsMouseClicked(ImGuiMouseButton.Left) && !UiInteract.Hover(min, max, false) &&
+            !UiInteract.Hover(sourceMenuAnchor.Min, sourceMenuAnchor.Max, false))
         {
             sourceMenuOpen = false;
         }
--- a/src/Aetherphone/Apps/Collections/CollectionsApp.Browse.cs
+++ b/src/Aetherphone/Apps/Collections/CollectionsApp.Browse.cs
@@ -638,7 +638,7 @@ internal sealed partial class CollectionsApp
             var centerY = (rowMin.Y + rowMax.Y) * 0.5f;
             var label = index == 0 ? Loc.T(L.Collections.AllSources) : sourceList[index - 1];
             var selected = index == sourceIndex;
-            var hovered = UiInteract.Hover(rowMin, rowMax);
+            var hovered = UiInteract.HoverWindowOnly(rowMin, rowMax);
             if (hovered)
             {
                 Squircle.Fill(drawList, rowMin, rowMax, 9f * scale, ImGui.GetColorU32(ui.HoverTint));
@@ -670,8 +670,8 @@ internal sealed partial class CollectionsApp
             return;
         }
 
-        if (ImGui.IsMouseClicked(ImGuiMouseButton.Left) && !UiInteract.Hover(min, max, false) &&
-            !UiInteract.Hover(sourceMenuAnchor.Min, sourceMenuAnchor.Max, false))
+        if (ImGui.IsMouseClicked(ImGuiMouseButton.Left) && !UiInteract.HoverWindowOnly(min, max, false) &&
+            !UiInteract.HoverWindowOnly(sourceMenuAnchor.Min, sourceMenuAnchor.Max, false))
         {
             sourceMenuOpen = false;
         }
--- a/src/Aetherphone/Apps/Games/Breakout/BreakoutApp.cs
+++ b/src/Aetherphone/Apps/Games/Breakout/BreakoutApp.cs
@@ -1,3 +1,4 @@
+using Aetherphone.Windows.Components;
 using Aetherphone.Apps.Games.Framework;
 using Aetherphone.Core.Animation;
 using Aetherphone.Core;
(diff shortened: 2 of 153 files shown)


src/Aetherphone/Windows/Components/DropdownMenu.cs
```diff
@@ -156,8 +156,8 @@ public int Draw(Rect screen, PhoneTheme theme, ReadOnlySpan<Item> items, out Row
                 cursorRight -= actionSlot;
             }
 
-            var rowHovered = ImGui.IsMouseHoveringRect(rowMin, rowMax);
-            var editHovered = item.CanEdit && editRect is { } er && ImGui.IsMouseHoveringRect(er.Min, er.Max);
+            var rowHovered = UiInteract.Hover(rowMin, rowMax);
```
XeldarAlz: Blocking: `Gate()` calls `BlockThisFrame()` at the top of every host app's Draw (AethergramApp.cs:274, JobsApp.cs:139, PhotosApp.cs:117), so `InputBlocked` is already true when this runs later in the same frame. `Hover` returns false for every row, so no menu item is hoverable or clickable, and line 213 inverts and closes the menu on the first click. Same at ChatMenuController.cs:228.

src/Aetherphone/Apps/Collections/CollectionsApp.Browse.cs
```diff
@@ -638,7 +638,7 @@ private void DrawSourceMenuOverlay()
             var centerY = (rowMin.Y + rowMax.Y) * 0.5f;
             var label = index == 0 ? Loc.T(L.Collections.AllSources) : sourceList[index - 1];
             var selected = index == sourceIndex;
-            var hovered = ImGui.IsMouseHoveringRect(rowMin, rowMax);
+            var hovered = UiInteract.Hover(rowMin, rowMax);
```
XeldarAlz: Blocking: line 630 reserves this exact rect via `HoverOverlay`, so `MouseOverOverlay` is true whenever the pointer is inside the menu. `hovered` is now always false and the `UiInteract.Click` at 658 never fires, so no source is selectable. Line 673 inverts too and dismisses the menu on the first click.

src/Aetherphone/Apps/Jobs/JobsApp.ColorPicker.cs
```diff
@@ -243,7 +243,7 @@ private int DrawPresets(ImDrawListPtr drawList, Rect area, PhoneTheme theme, flo
         {
             var center = new Vector2(area.Min.X + cellWidth * (index % PresetColumns + 0.5f),
                 area.Min.Y + rowHeight * (index / PresetColumns + 0.5f));
-            var hovered = ImGui.IsMouseHoveringRect(center - new Vector2(radius), center + new Vector2(radius));
+            var hovered = UiInteract.Hover(center - new Vector2(radius), center + new Vector2(radius));
```
XeldarAlz: Blocking: JobsApp.cs:140 calls `BlockThisFrame()` while `pickerOpen`, and the picker is drawn at JobsApp.cs:202 in the same frame. `hovered` is always false here and the click sits behind `if (!hovered) continue;`, so no preset can be picked. Save at line 275 and JobsApp.Categories.cs:315 fail identically.

src/Aetherphone/Windows/PhoneWindow.cs
```diff
@@ -202,6 +202,7 @@ public override void Draw()
     {
         LastPosition = ImGui.GetWindowPos();
         LastSize = ImGui.GetWindowSize();
+        Components.UiInteract.SetWindowHovered(ImGui.IsWindowHovered(ImGuiHoveredFlags.ChildWindows));
```
XeldarAlz: Blocking: `IsWindowHovered` without `AllowWhenBlockedByActiveItem` returns false whenever `g.ActiveId != 0`, including for the phone's own widgets. Focus any InputText (chat composer, SearchField) and the whole UiInteract layer goes dead until focus is lost: send, emoji, and field-clear stop responding. Add `ImGuiHoveredFlags.AllowWhenBlockedByActiveItem`.

src/Aetherphone/Core/Input/DragTracker.cs
```diff
@@ -21,7 +22,7 @@ public bool Begin(Rect startZone)
         }
 
         var position = ImGui.GetMousePos();
-        if (!startZone.Contains(position))
+        if (!UiInteract.Hover(startZone.Min, startZone.Max))
```
XeldarAlz: Non-blocking: `Hover` layers `InputBlocked` and `MouseOverOverlay` on top of the window check, which is wider than what this needs. That is the same coupling that froze the Flow drag. A window-gate-only helper would keep Control Center and notification swipes independent of DragScrollHost state, and would also fix the four blocking spots above.

src/Aetherphone/Windows/Components/ConfirmOverlay.cs
```diff
@@ -74,7 +74,7 @@ public void Draw(Rect screen, PhoneTheme theme)
                 service.CancelActive();
             }
             else if (!service.Busy && ImGui.IsMouseClicked(ImGuiMouseButton.Left) &&
-                     !ImGui.IsMouseHoveringRect(cardRect.Min, cardRect.Max))
+                     !UiInteract.Hover(cardRect.Min, cardRect.Max))
```
XeldarAlz: Non-blocking: this is a negated hover, so gating it flips the failure direction. With another window over the phone, `Hover` is false and a click inside that window now cancels the dialog. Same shape at ShareSheet.cs:72, ReportOverlay.cs:94, MusicApp.Playlists.cs:301, PersonPicker.cs:163.

src/Aetherphone/Apps/Games/Breakout/BreakoutApp.cs
```diff
@@ -1,3 +1,4 @@
+using Aetherphone.Windows.Components;
```
XeldarAlz: Nit: the new using lands above the existing sorted block. Same in Pairs, Reversi, Simon, Solitaire, Tetris, Twenty48, WaterSort, Whack and CameraApp.

src/Aetherphone/Apps/Games/Flow/FlowApp.cs
```diff
@@ -155,13 +155,12 @@ private bool DrawDifficultyRow(in GameContext context, float scale)
 
     private int ResolveHover(GameGrid grid)
     {
-        var mouse = ImGui.GetMousePos();
-        if (!grid.Bounds.Contains(mouse))
+        if (!grid.Bounds.Contains(ImGui.GetMousePos()))
```
XeldarAlz: Nit: dropping the cached `mouse` local means `GetMousePos()` is now called twice per frame here, and the same in Chess, Sudoku, Reversi, Whack and CrystalDrop. Sweeper kept its local; worth matching that.

XeldarAlz (commented):
Built clean on bb1a4de (`dotnet build -c Release`, 0 errors, 4 pre-existing NU1902 warnings) and 143/143 tests pass. Not verified in-game, so the runtime findings below come from reading the code.

The window-hover gate itself is the right fix. The problem is routing it through `UiInteract.Hover`, which also drags in `InputBlocked` and `MouseOverOverlay`: the overlays that raise those flags for everyone else now block themselves. Four blocking spots inline.

src/Aetherphone/Windows/Components/DropdownMenu.cs
BluntEXE: Fixed: added UiInteract.HoverWindowOnly (window-gate only, skips InputBlocked/MouseOverOverlay) and pointed the row/edit/dismiss hover tests at it. Same fix applied to ChatMenuController.cs:228.

src/Aetherphone/Apps/Collections/CollectionsApp.Browse.cs
BluntEXE: Fixed: row hover now uses UiInteract.HoverWindowOnly, and the outside-click dismiss checks too.

src/Aetherphone/Apps/Jobs/JobsApp.ColorPicker.cs
BluntEXE: Fixed: preset hover and Save button hover now use UiInteract.HoverWindowOnly. Same fix applied to JobsApp.Categories.cs:315.

src/Aetherphone/Windows/PhoneWindow.cs
BluntEXE: Fixed: added ImGuiHoveredFlags.AllowWhenBlockedByActiveItem.

src/Aetherphone/Core/Input/DragTracker.cs
BluntEXE: Fixed: added the window-gate-only helper (UiInteract.HoverWindowOnly) and pointed DragTracker.Begin at it, which also fixes the Flow drag freeze.

src/Aetherphone/Windows/Components/ConfirmOverlay.cs
BluntEXE: Fixed: added UiInteract.ClickedOutside (requires the click to land inside this window before counting as outside, rather than negating Hover). Applied here and at ShareSheet.cs, ReportOverlay.cs, MusicApp.Playlists.cs, and PersonPicker.cs.

src/Aetherphone/Apps/Games/Breakout/BreakoutApp.cs
BluntEXE: Fixed: sorted using directives in Breakout and the other games/CameraApp you listed.

src/Aetherphone/Apps/Games/Flow/FlowApp.cs
BluntEXE: Fixed: cached mouse in Flow. Chess/Sudoku/Reversi/Whack/CrystalDrop already only call GetMousePos() once, matching Sweeper.

PR #76, opened 2026-08-01
feat(aetherstream): join approval, party mode, and sync fixes

Kiro-XVI:
## Summary
- Fixes the actual root cause of "nothing works": the backend was returning HTTP 426 client_update_required because this fork was 184 commits behind upstream (merged in already, see the earlier merge commits on this branch).
- Fixes the real bugs found in live two-client testing: `WatchAlongSession.OnJoined`/`OnState`/`OnEnded` were firing on the background WebSocket thread instead of the main/Framework thread that `video.Play`/`Seek`/`Pause`/`Stop` and the D3D11 screen painter require - this is what gets video/audio actually playing for a viewer. Also fixes a screen-transform placement timing bug, a `VideoShareWatchPresence` toggle that was silently killing hosting when turned off, an indistinguishable failed-vs-empty search state on the Join screen, and surfaces mpv init failures via `VideoEngine.LastError` instead of silently showing "Playing" forever.
- Adds "party mode" (a host can open a room before anything's queued) and a "waiting for the next video" state instead of the in-world screen vanishing the moment the queue naturally empties.
- Adds a viewer Resync button, a Now Playing card that actually reflects what a viewer is watching (previously always showed "Nothing playing"), and queue persistence across a plugin reload or relog.
- Adds host approve/deny per viewer, shared queue suggestions, and host-initiated kick - all wired and tested-safe client-side, gated behind backend support that doesn't exist yet (safe no-ops until the server ships the matching `stream.joinRequest`/`stream.queueSuggest`/`stream.kick` signals, see the doc comments on `SignalType`).
- Backfills all 85 `aetherstream.*` locale keys across all nine languages (previously only in `L.cs`, never in the JSON catalogs) and adds an AetherStream mention to the README highlights, in every language.
- Drops dead dependencies and bundled binaries flagged in the port/alphachannel-engine review (#68, superseded by this branch): `Native/libmpv-2.dll` and `Native/yt-dlp.exe` (111 MB) were bundled but never actually loaded from there - the runtime updater already downloads its own copies into the Dalamud config directory. Also removes `NoireLib` (unreferenced, was pulling in `System.Data.SQLite`/`Castle.Core` and causing 650 analyzer warnings) and `Penumbra.Api`/`ZXing.Net` (only used by `TextureTranslate.cs`, the QR-code-into-Penumbra-.tex path from the old VFX screen-mounting approach this branch already replaced with `ScreenPainter`), plus the now-dead `APIHelper.cs`/`APIProvider.cs`. `latest.zip`: 72.3 MB -> 14.5 MB.

## Known follow-ups
- A Player-tab scroll bug (both kinetic drag and native wheel/scrollbar unresponsive while Lock Position is on) is still under investigation - temporary `[PlayerScroll]` diagnostic logging is in place but the root cause isn't confirmed yet.
- The three backend-gated features above need the corresponding server-side signals before they do anything.
- Remaining findings from the #68 review not yet addressed here: `WatchAlongSession.awaitingHostAck` join-timing bug, `NativeFileDialog` bypassing the `FilePicker`/ReShade guard, `VideoEngine` texture/cancellation-token leaks, blocking `Thread.Sleep`/un-awaited async in the render loop, the dead SNES path in `Plugin.cs`, zip-slip/no-hash-check on downloaded native binaries, style normalization on the ported video files, and a per-frame allocation in `GetScreenPresets()`.

## Test plan
- [x] Build succeeds (`dotnet build Aetherphone.sln -c Release`), 0 errors
- [x] Full test suite passes (293/293)
- [ ] Host can go live, a viewer can join, roster shows both correctly
- [ ] Video and audio play for the viewer
- [ ] Host's screen position/yaw/scale sync to the viewer's own screen
- [ ] Playback pause/seek sync between host and viewer
- [ ] Queue auto-advances to the next video, and stays up in a waiting state instead of closing once it empties
- [ ] Reload the plugin mid-queue and confirm the upcoming list survives


XeldarAlz (commented):
Backend for the three gated features is live on `dev` (`95f3acd`) — join approval, queue suggestions and kick all work now, plus `upcomingQueue` relay. Two answers to the open questions: the screen transform was genuinely broken (the fields weren't on the server envelope at all, and the state broadcast rebuilt the message from stored room state, so they were accepted and dropped — fixed, no retest needed), and empty/omitted `url` already worked and now has a test. Kick bars rejoin for the room's lifetime and replies `reason:"denied"`; deny is one-shot and doesn't bar a retry.

Signalling architecture and the threading fix both look right, and the licence diligence is thorough. One item I'd treat as blocking (unvalidated remote URLs reaching mpv) plus some smaller notes inline. Branch also needs a rebase — it's currently conflicting against `master`.

src/Aetherphone/Core/Video/WatchAlongSession.cs
```diff
@@ -0,0 +1,802 @@
...
+    // Runs on the main thread via OnFrameworkUpdate. Only ever called once OnState has already
+    // confirmed Mode == Viewing (or, via ResyncNow(), a re-apply of the last message under the
+    // same guarantee).
+    private void ApplyStateSync(CallControl message)
+    {
+        if (message.Url is { Length: > 0 } url && url != viewingUrl)
+        {
+            viewingUrl = url;
+            ViewingEntry = queue.CreateDisplayEntry(url);
+            video.Play(url);
```
XeldarAlz, line 589: Blocking: `url` here is whatever the host published, and it reaches `mpv_command(["loadfile", url, ...])` with no scheme check. A malicious host can push `\\\\attacker.host\\share\\x.mp4` and Windows will attempt SMB auth, leaking every viewer's NetNTLMv2 hash. `file:///` and mpv's other protocol handlers are reachable the same way, and any http URL becomes a "who is watching, from which IP" beacon.

`VideoEngine.ValidateURL` already does exactly the right check — gate this call on it. It has to be here rather than inside `Play`, since local-file playback is legitimate; only network-sourced URLs need the restriction.

src/Aetherphone/Core/Video/VideoEngine.cs
```diff
@@ -0,0 +1,474 @@
...
+        {
+            return _mpvRenderer?.GetMediaTitle() ?? string.Empty;
+        }
+
+        return string.Empty;
+    }
+
+    internal string? GetCurrentUrl() => _mpvRenderer?.GetCurrentUrl();
+
+    internal bool ValidateURL(string inputUrl, out Uri? url)
```
XeldarAlz: This has zero callers in the PR. It's the correct http/https + DNS-host check and it's the fix for the unvalidated remote URL path in `ApplyStateSync` — just needs wiring up. Worth applying to the incoming `stream.queueSuggestion` URL too.

src/Aetherphone/Core/Video/Resources.cs
```diff
@@ -0,0 +1,398 @@
...
+		}
+		
+		return null;
+	}
+
+	internal async Task CheckMPVAsync()
+	{
+		string filenameStart = "mpv-dev-lgpl-x86_64-";
+		string filenameEnd = ".7z";
+		string url = "https://api.github.com/repos/zhongfly/mpv-winbuild/releases/latest";
```
XeldarAlz, line 98: Downloads an executable and runs it, with no integrity check and no version pin — `/latest` means the binary can change between runs, and this is a third-party mpv build rather than official mpv. Suggest pinning a release tag and verifying a SHA-256. HTTPS to GitHub only proves you reached GitHub, not that the artifact is the one you reviewed.

THIRD-PARTY-NOTICES.md
```diff
@@ -26,6 +26,37 @@ The application icons under `src/Aetherphone/Icons/` are derived from
 - License: MIT (Copyright (c) 2020-2026 Paweł Kuna); full text reproduced in
   the MIT section below.
 
+## mpv
+
+`Native/libmpv-2.dll` is a redistributed binary build of libmpv, used for
```
XeldarAlz: This describes `Native/libmpv-2.dll` and `Native/yt-dlp.exe` as redistributed binaries, but `Resources.cs` downloads both into the Dalamud config dir at runtime "unlike the old bundled `Native/libmpv-2.dll` this replaces" — while the csproj still globs `Native\*.dll;Native\*.exe`. One of the three is stale; worth reconciling since this is a licence-facing document.

Same stale claim in the `VideoUrlResolver` header comment, which says the port "removes an entire runtime-downloaded-binary dependency" — no longer true now that yt-dlp is downloaded for `ytdl_hook`.

src/Aetherphone/Aetherphone.csproj
```diff
@@ -62,6 +68,14 @@
     <PackageReference Include="SixLabors.ImageSharp" Version="3.1.12" />
     <PackageReference Include="System.Security.Cryptography.ProtectedData" Version="10.0.0" />
     <PackageReference Include="Vortice.Direct3D11" Version="3.8.3" />
+    <PackageReference Include="Vortice.D3DCompiler" Version="3.8.3" />
+    <PackageReference Include="Penumbra.Api" Version="5.15.1" />
```
XeldarAlz: `Resources.cs` says the port has no "Penumbra dependency at all anymore" — looks like this reference is a leftover. Worth dropping if nothing consumes it; declaring it can affect plugin load expectations.

src/Aetherphone/Core/Video/APIHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+        string? url = _core.GetCurrentUrl();
+        if (string.IsNullOrEmpty(url))
+        {
+            return null;
+        }
+
+        int pos = (int)_core.GetInfo()[0];
+        string stateStr = _core.GetPaused() ? "paused" : "playing";
+
+        return JsonConvert.SerializeObject(new IPCVideoState(stateStr, Uri.EscapeDataString(url), pos, _resources.CurrentTimeNTPNormalizedMilliseconds));
```
XeldarAlz: Pulls in Newtonsoft.Json for one 4-field serialization while the rest of telephony uses source-generated System.Text.Json (`TelephonyJsonContext`). The ported file also keeps upstream style — `_core`/`_resources`, non-`sealed`, `pos`/`stateStr` — against the convention in `StreamSignalRouter` next door. ~15 lines to normalize both.

src/Aetherphone/Core/Video/WatchAlongSession.cs
```diff
@@ -0,0 +1,802 @@
...
+        // confirmed Hosting, later updates are just that, updates, not another "go live" moment.
+        if (Mode != WatchAlongMode.Hosting)
+        {
+            awaitingHostAck = true;
+        }
+
+        // Temporary diagnostic for the screen-transform sync path - remove once confirmed working
+        // end to end. IsActive being false here (screenPosition null) is the single most likely
+        // reason a viewer's screen stays parked at its own spawn fallback instead of following.
+        AepLog.Info(screenPosition is { } sp
```
XeldarAlz: This fires an `Info` line on every publish, and position drift makes that frequent during playback. `OnRoster` similarly builds a `string.Join`/`Select` allocation per roster update purely for diagnostics. Both are marked temporary — please strip them or drop to `Debug` before merge; this repo has had a log-flood problem before.

src/Aetherphone/Core/Video/WatchAlongSession.cs
```diff
@@ -0,0 +1,802 @@
...
+        var updated = new List<PendingJoinRequest>(PendingRequests.Count + 1);
+        foreach (var existing in PendingRequests)
+        {
+            if (existing.UserId != from.UserId)
+            {
+                updated.Add(existing);
+            }
+        }
+
+        updated.Add(new PendingJoinRequest(from.UserId, from.Name, from.World, from.DisplayName));
```
XeldarAlz, line 482: `from.Name` and `from.World` will always be empty strings. The server has never carried real name/world in `ParticipantInfo` — `displayName` only, everywhere, including `call.incoming` — and I kept `stream.joinRequest` and `stream.queueSuggestion` consistent with that rather than carving an exception. Render `DisplayName` or the UI shows blanks. Same applies to `OnQueueSuggested`.

src/Aetherphone/Core/Video/WatchAlongSession.cs
```diff
@@ -0,0 +1,802 @@
...
+
+        // Defensive tidy-up for the speculative host-approval extension: if someone we still
+        // thought was pending shows up in the authoritative roster (approved through some path
+        // other than our own ApproveRequest call), drop them from PendingRequests too rather than
+        // leaving a stale approve/deny row for someone already in.
+        if (PendingRequests.Count > 0)
+        {
+            foreach (var participant in Roster)
+            {
+                RemovePendingRequest(participant.UserId);
```
XeldarAlz, line 450: Nothing does the equivalent on `stream.left`. A pending requester who disconnects or withdraws now produces a `stream.left` to the host, so without clearing here too, a withdrawn request leaves a stale approve/deny row that silently no-ops when clicked.

src/Aetherphone/Apps/AetherStream/AetherStreamApp.Join.cs
```diff
@@ -0,0 +1,208 @@
...
+    private void TickNearbyRefresh()
+    {
+        nearbyRefreshTimer += ImGui.GetIO().DeltaTime;
+        if (nearbyRefreshTimer < NearbyRefreshIntervalSeconds)
+        {
+            return;
+        }
+
+        nearbyRefreshTimer = 0f;
+        watchAlong.RequestNearbyStreams();
```
XeldarAlz, line 126: `stream.nearby` has no server handler, so this 5s poll falls on the floor and the Nearby section stays permanently empty. Cheap (a WS frame, not HTTP) so it's dead UI rather than a load problem.

Worth noting it's also blocked on a design decision: approval currently layers on top of the mutual-contact + block gate rather than replacing it, so even with nearby implemented a non-contact still couldn't knock. Say the word if you want the permissive version — it's a one-line gate change server-side, but it means strangers can ping a host.

Kiro-XVI:
Update: pushed two more fixes after testing against the dev backend, which now ships join approval, queue suggestions, and kick for real (commit 95f3acd). These are no longer speculative/backend-gated as described above.

- `fix(aetherstream)`: wires up `stream.left` (was dispatched but never subscribed), which left stale approve/deny rows for requesters who dropped or withdrew before a decision. Also drops a caption line on pending join requests that always rendered blank, since the server never populates `Name`/`World` on `stream.joinRequest`, only `DisplayName`. Stale "speculative extension, never sent by the server today" comments across the protocol and UI code are updated to match.
- `fix(video)`: unrelated bug found in the same test session, `DxHandler.Dispose()` was defined but never called anywhere, so its D3D11 Present hook leaked on every plugin unload (`Aetherphone is leaking 1 hooks!` in the log). Now disposed in both `Plugin.cs` teardown paths.

src/Aetherphone/Aetherphone.csproj
```diff
@@ -39,6 +39,12 @@
     <Content Include="Sounds\**\*.mp3;Sounds\**\*.wav">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="Native\*.dll;Native\*.exe">
```
XeldarAlz: Blocking: this adds Native/libmpv-2.dll (94 MiB) and Native/yt-dlp.exe (17 MiB) to git permanently, while Resources.cs downloads its own copies into the config dir at runtime. Every clone pays 116 MB forever and the release zip ships a second copy. Keep one path, not both.

src/Aetherphone/Core/Video/VideoEngine.cs
```diff
@@ -0,0 +1,474 @@
...
+            formattedUrl = "https://" + formattedUrl;
+        }
+
+        return Uri.TryCreate(formattedUrl, UriKind.Absolute, out url)
+            && (url?.Scheme == Uri.UriSchemeHttp || url?.Scheme == Uri.UriSchemeHttps)
+            && url.Host.Contains('.') && !url.Host.EndsWith('.')
+            && Uri.CheckHostName(url.Host) == UriHostNameType.Dns;
+    }
+
+    internal bool PlaySnes(string path)
```
XeldarAlz: Blocking: PlaySnes has no caller anywhere in the plugin. That strands Snes9xRenderer.cs (840 lines), InputManager.cs, two persisted Configuration keys, and DownloadSNES9XAsync pulling an unpinned libretro nightly (Resources.cs:300). Deleting it also drops the emulator/ROM question from this PR.

src/Aetherphone/Core/Video/WatchAlongSession.cs
```diff
@@ -0,0 +1,831 @@
...
+            pendingViewerStop = true;
+        }
+
+        Mode = WatchAlongMode.None;
+        IsAwaitingApproval = false;
+        Roster = Array.Empty<WatchAlongParticipant>();
+        Interlocked.Exchange(ref pendingJoinSync, null);
+        Interlocked.Exchange(ref pendingStateSync, null);
+
+        confirm.Alert(Loc.T(L.AetherStream.KickedTitle), Loc.T(L.AetherStream.KickedBody),
```
XeldarAlz: Blocking: OnKicked, OnDeclined and OnQueueSuggestionResult all run on the WebSocket receive thread, and ConfirmService.Ask does a check-then-set on Active plus an Enqueue into a plain Queue<T>. Same thread-affinity bug this PR fixes for video.Play, just missed on the alert path.

src/Aetherphone/Core/Video/WatchAlongSession.cs
```diff
@@ -0,0 +1,831 @@
...
+    // sync already carries whatever the host plays to every viewer, no extra plumbing needed).
+    public void ApproveQueueSuggestion(string suggestionId)
+    {
+        var suggestion = FindQueueSuggestion(suggestionId);
+        if (suggestion is null)
+        {
+            return;
+        }
+
+        queue.Add(new VideoQueueEntry(suggestion.Url, suggestion.Url, string.Empty, null, null));
```
XeldarAlz, line 194: Blocking, same root as the ValidateURL thread: an approved suggestion goes straight into queue.Add, so a viewer-supplied UNC or file:// path reaches mpv the moment the host plays it. Host approval is not a filter here, since the title shown is the raw URL.

src/Aetherphone/Core/Video/TextureTranslate.cs
```diff
@@ -0,0 +1,130 @@
...
+using Penumbra.Api.Enums;
+using Penumbra.Api.IpcSubscribers;
+using ZXing;
+using ZXing.Common;
+using ZXing.QrCode;
+using ZXing.QrCode.Internal;
+
+namespace Aetherphone.Core.Video;
+
+internal sealed class TextureTranslate
```
XeldarAlz: Non-blocking, and it corrects my earlier Penumbra note: this class is never instantiated, and it is the only consumer of both Penumbra.Api and ZXing.Net. Deleting the file is what makes those two references droppable.

src/Aetherphone/Aetherphone.csproj
```diff
@@ -65,6 +71,14 @@
     <PackageReference Include="SixLabors.ImageSharp" Version="3.1.12" />
     <PackageReference Include="System.Security.Cryptography.ProtectedData" Version="10.0.0" />
     <PackageReference Include="Vortice.Direct3D11" Version="3.8.3" />
+    <PackageReference Include="Vortice.D3DCompiler" Version="3.8.3" />
+    <PackageReference Include="Penumbra.Api" Version="5.15.1" />
     <PackageReference Include="YoutubeExplode" Version="6.6.0" />
+    <PackageReference Include="SharpDX.Direct3D11" Version="4.2.0" />
+    <PackageReference Include="SharpDX.D3DCompiler" Version="4.2.0" />
+    <PackageReference Include="SharpDX.XAudio2" Version="4.2.0" />
+    <PackageReference Include="NoireLib" Version="1.8.0.3" />
```
XeldarAlz: Non-blocking: NoireLib has no code usage anywhere, and its analyzer is the source of every warning in the build (NoireLib_002, telling us to abandon System.Text.Json, against the source-generated STJ convention here). Worth dropping the reference.

src/Aetherphone/Apps/AetherStream/AetherStreamApp.Player.cs
```diff
@@ -0,0 +1,592 @@
...
+        // would otherwise just get clipped off the bottom with no way to reach it.
+        using (AppSurface.Begin(body))
+        {
+            // Temporary diagnostic for the Player-tab scroll bug (wheel + kinetic drag both
+            // reported dead while Lock Position is on) - remove once confirmed fixed. Fires only
+            // on actual wheel/click activity so it doesn't spam the log every frame.
+            var scrollIo = ImGui.GetIO();
+            if (scrollIo.MouseWheel != 0f || (ImGui.IsMouseClicked(ImGuiMouseButton.Left) && ImGui.IsWindowHovered()))
+            {
+                AepLog.Info($"[PlayerScroll] wheel={scrollIo.MouseWheel:F2} scrollY={ImGui.GetScrollY():F1} " +
```
XeldarAlz: Non-blocking: a third temporary diagnostic, and this one sits in draw code, firing on every wheel tick and every click while hovered. Same request as the WatchAlongSession logs: strip it or drop it to Debug before merge.

src/Aetherphone/Apps/AetherStream/AetherStreamApp.Join.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        var avatarRadius = 18f * scale;
+        var avatarCenter = new Vector2(rect.Min.X + 8f * scale + avatarRadius, rect.Center.Y);
+        AvatarView.DrawRemote(drawList, avatarCenter, avatarRadius, theme, row.Name, row.World, null, remoteImages,
+            lodestone, 0.8f, 28);
+
+        var textLeft = avatarCenter.X + avatarRadius + 12f * scale;
+        Typography.Draw(new Vector2(textLeft, rect.Center.Y - 16f * scale), row.DisplayName, theme.TextStrong,
+            TextStyles.Body);
+        Typography.Draw(new Vector2(textLeft, rect.Center.Y + 2f * scale), $"{row.Name}  ·  {row.World}",
```
XeldarAlz: Non-blocking: the join-request caption got fixed, but this row still renders row.Name and row.World, and feeds them to AvatarView.DrawRemote. If NearbyStreamInfo is DisplayName-only like ParticipantInfo, this shows a blank separator and a fallback avatar.

src/Aetherphone/Core/Video/Resources.cs
```diff
@@ -0,0 +1,398 @@
...
+					_ = DownloadYTDLPAsync();
+				}
+			});
+		});
+	}
+
+	internal string? GetLocationMPV()
+	{
+		string filenameStart = "mpv-dev-lgpl-x86_64-";
+		string? dir = Directory.GetDirectories(_configDir, $"{filenameStart}*").FirstOrDefault();
```
XeldarAlz: Nit: the ported files keep upstream formatting, tabs instead of 4-space, across Resources.cs, MpvRenderer.cs, ScreenPainter.cs, Snes9xRenderer.cs, InputManager.cs and DxHandler.cs, against .editorconfig. One reformat pass settles this and the APIHelper style thread together.

src/Aetherphone/Windows/Components/ImageFit.cs
```diff
@@ -1,9 +1,30 @@
 using Aetherphone.Core;
+using Dalamud.Bindings.ImGui;
+using Dalamud.Interface.Textures.TextureWraps;
 
 namespace Aetherphone.Windows.Components;
 
 internal static class ImageFit
 {
+    public static readonly Vector4 LetterboxFill = new(0f, 0f, 0f, 1f);
```
XeldarAlz: Non-blocking: the last four commits add 334 lines of photo aspect and letterbox work across Aethergram, Velvet, Camera, ImageProcessor and WallpaperCrop, unrelated to watch-along. One concern per PR, and it makes an already large diff harder to land. Worth splitting out.

XeldarAlz (changes requested):
Built at 62bf4179: 0 errors, 293 tests pass. Not run in-game. Every warning in the log is NoireLib_002, from the NoireLib package this PR adds; master has none. Locale lockstep verified: 87 aetherstream keys in L.cs and in all nine JSONs, real translations, no em dashes.

Of the ten open threads, only stream.left is fixed and the DisplayName one is half done. Both blocking URL items are unchanged: VideoEngine.ValidateURL still has zero callers. New findings inline, four blocking.

XeldarAlz (changes requested):
Built at abedcf5f: 0 errors, 293/293 tests pass. Not run in-game. The NoireLib_002 warning flood is gone, only the pre-existing AngleSharp NU1902 is left.

abedcf5f clears the whole dependency cluster (bundled binaries, NoireLib, Penumbra/ZXing, TextureTranslate, APIHelper), and GetLocationMPV/GetLocationYTDLP do only resolve from the config dir as you say, so I have resolved those five threads plus the stream.left one.

Nothing else moved. All five blocking items are unchanged at head: ValidateURL still has zero callers, so both URL paths still reach mpv unvalidated, ConfirmService.Alert still fires from the WS receive thread in three handlers, and PlaySnes is still dead. The branch is also 122 commits behind master with 6 conflicting files now, worth rebasing before the next round.

THIRD-PARTY-NOTICES.md
XeldarAlz: abedcf5f settles this the other way: both binaries are gone from the tree, so the mpv and yt-dlp sections now describe files that do not exist. Point them at the runtime download into the config dir instead. The AlphaChannel section (line 52) also still credits the SNES9x renderer, which ties into the PlaySnes thread.

XeldarAlz:
@Kiro-XVI ping on this one. Nothing has moved since abedcf5f, and everything between here and merge is already written up in the last review: wire ValidateURL into both URL paths (state sync and approved suggestions), move the three ConfirmService alerts off the WS receive thread, drop the dead PlaySnes path, and point the mpv/yt-dlp sections of THIRD-PARTY-NOTICES.md at the runtime download instead of the deleted binaries. The branch is also conflicting against master now, so it needs a rebase. Push those and I'll do a final pass and merge.

PR #77, opened 2026-08-01
feat(jobs): drag to reorder custom job categories and their gearsets

Kiro-XVI:
## Summary
- Custom job categories had no way to reorder at all, and a naive reorder UI couldn't have worked anyway: `JobsReader.Build` always re-sorted every category's entries by game-defined priority/abbreviation on every rebuild (every 2 seconds, plus after any edit), discarding any manual order. Custom categories now display in `GearsetIds`' own list order instead.
- Drag-and-drop to reorder categories themselves, via a grip handle on each custom section's header.
- Drag-and-drop to reorder the gearsets within one category, via a grip handle on each row.
- Along the way, fixed the phone appearing to move during either drag: `DragScrollHost`'s kinetic scroll was racing the drag handle for the same press (it checks for a new scroll gesture before the handle gets a chance to claim it). Fixed by handing the gesture back via a new `AppSurface.SurfaceScope.CancelDrag()`, the same escape hatch `SkywatcherApp`'s scrubber already used for the identical problem.

## Test plan
- [ ] Build succeeds (`dotnet build Aetherphone.sln -c Release`)
- [ ] Create 2+ custom job categories, each with 2+ gearsets
- [ ] Drag a category header up/down and confirm the section order updates and persists across a Jobs tab close/reopen
- [ ] Drag a gearset row within a category and confirm the row order updates and persists
- [ ] Confirm the phone window and list scroll position stay put during both drags

### DIFF
--- a/src/Aetherphone/Apps/Jobs/JobsApp.Categories.cs
+++ b/src/Aetherphone/Apps/Jobs/JobsApp.Categories.cs
@@ -16,6 +16,13 @@ internal sealed partial class JobsApp
     private const float EditorTitleHeight = 18f;
     private const float EditorFieldHeight = 30f;
     private const int CategoryNameMaxLength = 24;
+    private const float CategoryDragHandleRadius = 9f;
+
+    // Matches AetherStreamQueue's own reorder drag threshold - the phone-appears-to-move bug
+    // turned out to be DragScrollHost racing the handle for the same press (fixed via
+    // surface.CancelDrag() in JobsApp.cs), not a timing race with the window lock, so there's no
+    // need for a long-press dead zone here after all.
+    private const float DragThreshold = 7f;
 
     private static readonly List<JobsCategory> NoCategories = new();
 
@@ -25,6 +32,27 @@ internal sealed partial class JobsApp
     private int categoryEditorOpenedFrame;
     private string categoryEditorName = string.Empty;
     private bool focusCategoryField;
+
+    // Rebuilt every frame from the on-screen custom section headers (see JobsApp.cs's main draw
+    // loop) - used both to place each header's drag handle and, once a drag is active, to find
+    // the nearest header to the cursor as the drop target. Header heights are fixed but section
+    // card heights vary with gearset count, so target detection compares against real header
+    // positions rather than assuming a uniform row height (contrast AetherStreamQueue.Reorder's
+    // drag, where every row is the same height).
+    private readonly List<(int CategoryIndex, Rect Rect)> categoryHeaderRects = new();
+    private int categoryDragIndex = -1;
+    private Vector2 categoryDragPressPos;
+    private bool categoryDragActive;
+
+    // Reordering the jobs/gearsets *within* one custom category - separate from the header drag
+    // above, which reorders the categories themselves. Rows inside a section are uniform height
+    // (RowHeight), so this can use AetherStreamQueue's simpler displaced-row drag math instead of
+    // the nearest-header search the variable-height category headers need.
+    private int jobDragCategoryIndex = -1;
+    private int jobDragIndex = -1;
+    private Vector2 jobDragStart;
+    private float jobDragY;
+    private bool jobDragActive;
 
     private List<JobsCategory> CurrentCategories()
     {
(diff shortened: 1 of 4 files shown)


src/Aetherphone/Apps/Jobs/JobsApp.cs
```diff
@@ -170,28 +170,52 @@ public void Draw(in PhoneContext context)
...
                 }
+
+                // Fixes the phone appearing to move during a category/row drag: it was never the
+                // window's own position, it was DragScrollHost's kinetic scroll racing the drag
+                // handle for the exact same press. DragScrollHost.Begin runs before any row/handle
+                // this frame, so on the press's first frame it sees "no widget active yet" and
+                // happily starts its own scroll gesture on the same touch. Handing it back once a
+                // Jobs drag has been picked up (same pattern as SkywatcherApp's scrubber) stops
+                // that race outright.
+                if (categoryDragIndex >= 0 || jobDragIndex >= 0)
```
XeldarAlz, line 210: Blocking: `OnClosed` (JobsApp.cs:66) does not reset `categoryDragIndex`, `jobDragIndex`, `jobDragActive` or `jobDragY`. Close the phone mid-drag and the release is never observed; on reopen the first drawn frame takes the mouse-up branch and commits a reorder wherever the cursor happens to sit. Reset the drag fields in `OnClosed`.

src/Aetherphone/Apps/Jobs/JobsApp.Categories.cs
```diff
@@ -89,6 +117,244 @@ private void DrawCategoriesMenu(Rect content, PhoneTheme theme)
...
+    // starting a reorder.
+    private void DrawCategoryDragHandle(Rect headerRect, int categoryIndex, float scale)
+    {
+        categoryHeaderRects.Add((categoryIndex, headerRect));
+
+        var drawList = ImGui.GetWindowDrawList();
+        var radius = CategoryDragHandleRadius * scale;
+        var center = new Vector2(headerRect.Max.X - radius - 2f * scale, headerRect.Center.Y);
+        ImGui.SetCursorScreenPos(center - new Vector2(radius));
+        ImGui.InvisibleButton($"##categoryDragHandle{categoryIndex}", new Vector2(radius * 2f));
```
XeldarAlz, line 138: Blocking: this is the only raw `InvisibleButton` in the repo and it bypasses `UiInteract`, so the handle stays hoverable and pressable while `menu.Gate()` / `BlockThisFrame()` has input blocked. Clicking a row menu item or the category editor that overlaps a handle starts a drag underneath it. Gate on `!UiInteract.InputBlocked` before accepting `IsItemActivated`.

src/Aetherphone/Apps/Jobs/JobsApp.Categories.cs
```diff
@@ -89,6 +117,244 @@ private void DrawCategoriesMenu(Rect content, PhoneTheme theme)
...
+    //
+    // A real InvisibleButton for the same reason as DrawCategoryDragHandle - without an active
+    // ImGui item under the press, Dear ImGui's own window-drag takes over instead, dragging the
+    // whole phone rather than starting a reorder. Returns whether the handle is hovered, so
+    // DrawJobRow can exclude that area from the row's own hover/equip-click handling.
+    private bool DrawJobDragHandle(ImDrawListPtr drawList, Vector2 center, float radius, int categoryIndex,
+        int rowIndex, float scale)
+    {
+        ImGui.SetCursorScreenPos(center - new Vector2(radius));
+        ImGui.InvisibleButton($"##jobDragHandle{categoryIndex}_{rowIndex}", new Vector2(radius * 2f));
```
XeldarAlz, line 269: Blocking: the interpolated label allocates two strings per handle per frame (here and line 138), which the no-allocation-in-Draw rule rules out. `ImGui.PushID(rowIndex)` around a constant literal id gives the same unique ImGui id for free.

src/Aetherphone/Apps/Jobs/JobsApp.Categories.cs
```diff
@@ -89,6 +117,244 @@ private void DrawCategoriesMenu(Rect content, PhoneTheme theme)
...
+    {
+        for (var index = 0; index < categoryHeaderRects.Count; index++)
+        {
+            if (categoryHeaderRects[index].CategoryIndex != categoryIndex)
+            {
+                continue;
+            }
+
+            var rect = categoryHeaderRects[index].Rect;
+            var drawList = ImGui.GetForegroundDrawList();
```
XeldarAlz, line 232: Non-blocking: the foreground draw list is not clipped to the phone, and `categoryHeaderRects` also holds headers scrolled outside the surface, so this highlight can paint over the game window. `ClosestCategoryHeader` has no distance cap either, so an off-screen header is a valid drop target. ControlCenter.cs:104 pushes a clip rect for exactly this.

src/Aetherphone/Apps/Jobs/JobsApp.cs
```diff
@@ -170,28 +170,52 @@ public void Draw(in PhoneContext context)
...
                         var section = sections[index];
                         var title = section.IsCustom ? section.CustomTitle : Loc.T(section.RoleTitle);
+                        var headerTop = ImGui.GetCursorScreenPos();
+                        var headerWidth = ImGui.GetContentRegionAvail().X;
                         ui.SectionHeading(title, index == 0 ? 8f : 4f);
+                        if (section.IsCustom)
+                        {
+                            var headerRect = new Rect(headerTop,
+                                new Vector2(headerTop.X + headerWidth, ImGui.GetCursorScreenPos().Y));
+                            DrawCategoryDragHandle(headerRect, section.CategoryIndex, scale);
```
XeldarAlz, line 193: Non-blocking: `InvisibleButton` advances the ImGui cursor, so it lands roughly 2 to 4 scaled px above where `SectionHeading` left it, and every custom section card sits tighter under its heading than the role sections do. Restore the cursor to `headerRect.Max.Y` after drawing the handle.

src/Aetherphone/Apps/Jobs/JobsApp.Categories.cs
```diff
@@ -16,6 +16,13 @@ internal sealed partial class JobsApp
     private const float EditorTitleHeight = 18f;
     private const float EditorFieldHeight = 30f;
     private const int CategoryNameMaxLength = 24;
+    private const float CategoryDragHandleRadius = 9f;
+
+    // Matches AetherStreamQueue's own reorder drag threshold - the phone-appears-to-move bug
```
XeldarAlz, line 21: Non-blocking: `AetherStreamQueue` does not exist in this repo (4 references). The real precedent is ControlCenter.cs:28, which already declares `DragThreshold = 7f`. Line 240's "old up/down-arrow version this replaced" also contradicts the PR body, which says there was no reorder at all. Comments here are why-only, so these blocks should go.

src/Aetherphone/Apps/Jobs/JobsApp.Categories.cs
```diff
@@ -26,6 +33,27 @@ internal sealed partial class JobsApp
...
+    private readonly List<(int CategoryIndex, Rect Rect)> categoryHeaderRects = new();
+    private int categoryDragIndex = -1;
+    private Vector2 categoryDragPressPos;
+    private bool categoryDragActive;
+
+    // Reordering the jobs/gearsets *within* one custom category - separate from the header drag
+    // above, which reorders the categories themselves. Rows inside a section are uniform height
+    // (RowHeight), so this can use AetherStreamQueue's simpler displaced-row drag math instead of
+    // the nearest-header search the variable-height category headers need.
+    private int jobDragCategoryIndex = -1;
```
XeldarAlz, line 51: Nit: `Core/Input/DragTracker.cs` already tracks press origin, threshold, delta and release, and ControlCenter uses it for tile reordering. Both state machines here reimplement it, which is five extra fields each to keep in sync.

XeldarAlz (changes requested):
Built f6ff29a8 in a clean worktree (`dotnet build Aetherphone.sln -c Release`): succeeded, 0 errors. Test suite: 278/278 pass. Not verified in-game.

Three blocking items inline: drag state survives `OnClosed` and later commits a reorder the user never made, the new `InvisibleButton` handles ignore the `UiInteract` input gate, and their labels allocate every frame in Draw. Four non-blocking notes follow.

XeldarAlz (changes requested):
Good catch on the root cause. `JobsReader.Build` re-sorting custom buckets on every rebuild is exactly why a naive reorder UI would have looked broken, and finding the `DragScrollHost` race rather than blaming `LockPosition` is the right diagnosis. The feature is worth having; the notes below are about the seams around it.

I have pushed a rebase plus fixes to `pr-77-fixes` on this repo (your commit is preserved as-is with you as author, my changes sit in a second commit on top) so nothing here needs redoing by hand. Verified: `dotnet build -c Release` clean, 517/517 tests pass.

## Blocking

**1. The branch no longer merges.** `master` grew its own `SurfaceScope.CancelDrag()` in `d07438b0`, so the `AppSurface.cs` hunk here is a duplicate and is the conflict. Dropped it; `surface.CancelDrag()` works against master unchanged.

**2. `DrawCategoryDragHandle` corrupts the layout cursor.** It does `SetCursorScreenPos(center - radius)` + `InvisibleButton` and never restores, and `DrawSectionCard` reads `GetCursorScreenPos()` as the card origin. `SectionHeading` leaves the cursor at `headerTop.Y + 30*scale + spacing`; the handle leaves it at `headerTop.Y + 24*scale + 1.5*spacing`. Net result: every custom section's card is drawn **4*scale px too high on the first section and 2*scale on the rest** (the `topPad` differs, 8f vs 4f), the error grows with UI scale, and it compounds down the stack. `HousingApp.Map.cs:401-406` has the pattern: save the cursor, position, submit, restore.

**3. The handles bypass `UiInteract`'s input gate.** `ImGui.IsItemHovered()` / `IsItemActivated()` ignore `InputBlocked`, the overlay reservation and `WindowHovered`, all of which `UiInteract.Hover` honours (`UiInteract.cs:75-76`). `JobsApp.Draw` calls `UiInteract.BlockThisFrame()` when the color picker or category editor is open, but the handles underneath stayed live, so a press *through* the overlay started a reorder. Same for an open dropdown. Both handles now confirm through `UiInteract.Hover` before claiming a press.

**4. The row drag can dead-lock permanently.** `UpdateJobDrag` was the only thing clearing `jobDragIndex`, and it only ran from `DrawSectionCard` for the one matching custom section. If that section stopped drawing while the press was held (logout, since `Draw` skips the whole branch when `LocalPlayer is null`; closing the app, since `OnClosed` did not reset the drag fields; or the category being deleted), `jobDragIndex` stayed set and `DrawJobDragHandle`'s `if (jobDragIndex < 0 && activated)` guard then refused **every** row drag for the rest of the session. Both gestures now release from `Draw` itself (`FinishCategoryDrag` / `FinishJobDrag`), and `OnClosed` resets them.

## Should fix

**5. Nearest-header-center drop targeting misfires with variable-height cards.** With three categories whose headers sit at Y 100 / 700 / 800, dragging the third up to Y 400 ties and lands it in slot **0**, not slot 1: the midpoint between two header centers says nothing about the boundary between their slots when one section is eight gearsets tall and the next is one. There was also no distance cap, so releasing anywhere, even off the phone, committed a reorder with no way to abort. Categories now claim the vertical span of their whole section, and a release outside every span cancels.

**6. `GetForegroundDrawList()` for the drop highlight is unclipped.** The foreground list has no clip rect for the scrolling surface, so a header scrolled partway under the app header bar still painted its highlight over the header and the bezel. The other foreground-list uses in `Apps/` are full-screen overlays, not in-surface decoration. Moved to the window draw list, drawn inside the surface.

**7. `categoryHeaderRects.Clear()` sat inside the non-empty branch**, so with no sections the list kept the previous frame's rects while the update still ran. Cleared once per frame now.

## Non-blocking, also applied

- **Comments.** About 55 of the 352 added lines were comments, and most were PR-description narrative rather than a why the code cannot express ("the phone-appears-to-move bug turned out to be...", "so there's no need for a long-press dead zone here after all"). Per `CLAUDE.md` these belong in the PR body; kept the `InvisibleButton` justification, which is a genuine why.
- **Four comments cite `AetherStreamQueue`, which does not exist in this repo** (`grep -r AetherStreamQueue src/` is empty; that is the unmerged AlphaChannel port). `DragThreshold = 7f` was a magic constant whose only justification pointed at a file that is not there.
- **A comment contradicted the PR body**: `ReorderCategory` said "unlike the old up/down-arrow version this replaced", but the body correctly says there was no way to reorder at all.
- **Both reorders mutated through `CurrentCategories()`**, the read path, which can return the shared static `NoCategories`. The convention here is `CategoriesForWrite()` behind a `CurrentContentId == 0` guard (`AssignGearsetToCategory`, `RemoveGearsetFromCategory`). The count guards kept it safe, but it is one refactor from writing into a process-wide static.
- **`CustomOrderedEntries` allocated a `Dictionary` + `List` + `ToArray()` per custom category per rebuild**, and `Rebuild()` runs every 2s (every 0.1s while an equip is pending). Now a single right-sized array. The behaviour was correct as written: `BuildGearsetEntries` only buckets via `categoryByGearsetId`, so no entry could be silently dropped.
- **Lifted-row polish**: travel is clamped to its own card so a dragged row cannot paint over a neighbouring section, the clamp matches the drop math so the preview agrees with the result, and the dragged row draws last so it reads as lifted rather than sliding under its neighbours.
- **`DrawJobDragHandle`'s `scale` parameter was unused**, and `jobDragY` became write-only state; both removed.

## One thing I got wrong

I initially flagged the interpolated `ImGui.InvisibleButton($"##jobDragHandle{...}")` labels as per-frame allocations. They are not: `ImU8String` is an `[InterpolatedStringHandler]` ref struct with a 512-byte inline buffer and an `ArrayPool` fallback, so short labels never allocate. Left as written.

## Still needs an in-game pass

The build and unit tests are green, but the four behavioural boxes in the test plan are untested by me: the two drags persisting across a tab close/reopen, and the phone staying put during each. Worth running before this lands.

XeldarAlz:
Merged to master as 400e594d, with your commit intact and you as its author.

GitHub cannot mark this PR merged because it landed via `pr-77-fixes` (your branch was rebased off the `AppSurface.cs` conflict and carried a review-fix commit on top), so closing it manually. The work itself is in.

Thanks for this one, the diagnosis was the hard part: `JobsReader.Build` re-sorting custom buckets on every rebuild is exactly why a reorder UI would have looked broken no matter how it was built, and tracing the phone-appears-to-move symptom to the `DragScrollHost` gesture race rather than to `LockPosition` is not an obvious call.

PR #79, opened 2026-08-02
feat(notifications): toggle notification toast

Radvo:
## What

Adds toggle for notification toast banners. Users can now disable visual toasts while keeping sounds/vibration, or disable sounds/vibration while keeping toasts.

Also adds a mouseover hint icon support to SettingsRow.Bool() for cleaner inline help text ("Vibration" and "Show Toasts") and collapsing options when notifications are toggled off.

Detached vibration event from banner showing.

## Why

There is currently no way to disable only toasts, just disabling everything or only sounds. This adds a toggle for more customization on how users want to receive notifications.

## How to test

1. Open /phone → Settings → Notifications
2. Toggle "Show Notification Toasts" off
3. Trigger a test notification (/phone test)
4. Verify: No toast appears, but sound plays and phone vibrates
5. Toggle "Show Notification Toasts" back on
6. Verify: Toast appears again (/phone test)
7. Toggle "Do Not Disturb" on
8. Verify: ShowNotificationToasts, Vibration, and Notification Sound rows are hidden
9. Toggle "Do Not Disturb" back off
10. Open /phone → Settings → Notifications → Linkpearl and repeat steps 2-6
11. Open /phone → Settings → Notifications and hover over new hint labels, validate hint text fades in/out.

Also tested on all phone and text size options.

## Checklist


src/Aetherphone/Core/Shell/PhoneShell.cs
```diff
@@ -157,7 +158,7 @@ public bool ConsumeCloseRequest()
 
     public void ForceMinimized() => minimize.SnapMinimized();
 
-    private void OnBannerShown()
+    private void OnVibration(PhoneNotification _)
```
XeldarAlz: Blocking: this loses the banner's gates. `NotificationBanner.OnPresented` bailed out on `!phoneVisible()` and on the notification's app already being open, so `Shown` never fired there. Now every notification shakes. And `shake.Advance` only runs in `Draw` (PhoneShell.cs:202), so a shake fired while the phone is closed stays queued and plays on next open.

src/Aetherphone/Core/Notifications/NotificationService.cs
```diff
@@ -150,7 +151,17 @@ private void Present(PhoneNotification notification)
         Added?.Invoke(stamped);
         if (!configuration.DoNotDisturb)
         {
-            Presented?.Invoke(stamped);
+            var appSetting = configuration.NotificationSettingFor(notification.SettingsKey);
```
XeldarAlz: Non-blocking: `NotificationSettingFor` inserts a default entry when the key is missing, so every app that notifies now writes itself into the persisted `NotificationSettings` dictionary from the present path. Add a read-only `ShouldShowToast(appId)` next to `IsAppNotificationEnabled` (TryGetValue, no insert).

src/Aetherphone/Apps/Settings/Pages/NotificationsPage.cs
```diff
@@ -41,7 +41,7 @@ public void Draw(in PhoneContext context, Rect body)
         using (AppSurface.Begin(body))
         {
             SettingsSection.Header(Loc.T(L.Common.Alerts), theme);
-            var alerts = GroupCard.Begin(theme, 3);
+            var alerts = GroupCard.Begin(theme, configuration.DoNotDisturb ? 1 : 4);
```
XeldarAlz: Non-blocking: the card height comes from the pre-toggle value but the rows below come from the post-toggle one, so on the frame DND flips, 4 rows draw over a 1-row card and `End()` reserves the wrong height, overlapping the next section. Read `DoNotDisturb` into a local before `Begin` and branch on that.

src/Aetherphone/Apps/Settings/Pages/AppNotificationPage.cs
```diff
@@ -35,13 +35,25 @@ public void Draw(in PhoneContext context, Rect body)
         using (AppSurface.Begin(body))
         {
             SettingsSection.Header(Loc.T(L.Common.Alerts), theme);
-            var card = GroupCard.Begin(theme, 1);
+            var appSetting = configuration.NotificationSettingFor(channel.AppId);
             var wasEnabled = configuration.IsAppNotificationEnabled(channel.AppId);
+            var card = GroupCard.Begin(theme, wasEnabled ? 2 : 1);
```
XeldarAlz, line 40: Non-blocking: same one-frame mismatch as NotificationsPage.cs:44. The card is sized from `wasEnabled` but the Show Toasts row is drawn from `enabled`, so on the frame you flip the toggle the second row paints outside the card. Branch on `wasEnabled` and let the new value apply next frame.

src/Aetherphone/Core/Shell/PhoneShell.cs
```diff
@@ -80,7 +81,7 @@ public PhoneShell(PhoneServices services, AppBundle bundle)
         MusterChatBridge.Bind(services.Musters, services.MusterLauncher, navigation);
         AdChatBridge.Bind(services.YellowPages, services.YellowPagesLauncher, navigation);
         banner = new NotificationBanner(notifications, VisibleAppId, PhoneVisible, router);
-        banner.Shown += OnBannerShown;
+        notifications.Vibration += OnVibration;
```
XeldarAlz, line 84: Nit: PhoneShell was the only subscriber to `NotificationBanner.Shown`. With this line replacing it the event is dead, so drop `Shown` and both `Shown?.Invoke()` sites at NotificationBanner.cs:316 and :327.

src/Aetherphone/Windows/Components/MinimizedPhone.cs
```diff
@@ -110,13 +110,13 @@ private static void DrawBadge(ImDrawListPtr dl, Rect body, float scale, float al
         Typography.Draw(dl, center - labelSize * 0.5f, label, new Vector4(1f, 1f, 1f, alpha), 0.66f, FontWeight.Bold);
     }
 
-    private void OnPresented(PhoneNotification _)
+    private void OnVibration(PhoneNotification _)
```
XeldarAlz, line 113: Nit: `configuration.Vibration` is now checked three times for one shake, at NotificationService.cs:160 which gates the invoke, then here and in `PhoneShell.OnVibration`. Keep the gate in one place, the service is the natural one.

src/Aetherphone/Windows/Components/SettingsRow.cs
```diff
@@ -10,18 +10,37 @@ internal static class SettingsRow
...
+    public static bool Bool(Rect row, string label, bool value, PhoneTheme theme, string? id = null, string? hint = null)
     {
         var scale = ImGuiHelpers.GlobalScale;
         var width = Metrics.Size.ToggleWidth * scale;
         var height = Metrics.Size.ToggleHeight * scale;
-        var min = new Vector2(row.Max.X - width, row.Center.Y - height * 0.5f);
-        var labelMaxWidth = MathF.Max(1f, min.X - 10f * scale - row.Min.X);
+        var toggleMin = new Vector2(row.Max.X - width, row.Center.Y - height * 0.5f);
+        
+        var iconHeight = 22f * scale;
```
XeldarAlz: Nit: `22f`, `8f` and the `2f` below are raw pixels, spacing and sizing go through `Metrics` tokens times `GlobalScale` per the style rules. The blank lines added in this method (and in AppNotificationPage) also carry trailing whitespace, which `.editorconfig` trims.

src/Aetherphone/Core/Localization/L.cs
```diff
@@ -922,6 +922,9 @@ internal static class Settings
         public static readonly LocString DoNotDisturb = new("settings.doNotDisturb", "Do Not Disturb");
         public static readonly LocString Vibration = new("settings.vibration", "Vibration");
         public static readonly LocString VibrationHint = new("settings.vibrationHint", "The phone shakes briefly when a notification arrives.");
+        public static readonly LocString ShowNotificationToasts = new("settings.showNotificationToasts", "Show Notification Toasts");
+        public static readonly LocString ShowNotificationToastsHint = new("settings.showNotificationToastsHint", "Display banner notifications at the top of the screen.");
+        public static readonly LocString ShowToasts = new("settings.showToasts", "Show Toasts");
```
XeldarAlz: Nit: `showToasts` and `showNotificationToasts` are two keys for the same idea, one would do. English also says Toasts while all nine locales translate it as banner and the component is `NotificationBanner`, so banner wording reads more consistent in English too.

src/Aetherphone/Apps/Settings/Pages/NotificationsPage.cs
```diff
@@ -50,25 +50,33 @@ public void Draw(in PhoneContext context, Rect body)
                 configuration.Save();
             }
 
-            var vibration = SettingsRow.Bool(alerts.NextRow(), Loc.T(L.Settings.Vibration),
-                configuration.Vibration, theme);
-            if (vibration != configuration.Vibration)
+            if (!configuration.DoNotDisturb)
```
XeldarAlz: Nit: Do Not Disturb is a temporary mode, but this hides Vibration, Notification Sound and the only route to the sound page while it is on, so you cannot set alert preferences until you turn it off. Consider leaving the rows visible, or dimmed.

XeldarAlz (changes requested):
Release build and all 304 tests pass on e564e43, not verified in game. One blocking item on the vibration refactor, everything else is inline and non-blocking.

src/Aetherphone/Apps/Settings/Pages/AppNotificationPage.cs
Radvo: While testing saw this leads to a barely noticeable stagger on how drawing the sound list happens frame 1 because of checking on `enabled` on line 48 (now 60), are you okay with me also switching that to `wasEnabled`  ?

Radvo:
Pushed new version addressing comments :
1. Added gates to vibration firing
2. Validated json no longer gets default values added
3. Opted for dimming when do not disturb is on rather than hiding as per last comment
4. Fixed card sizing based on wrong variable for app notification settings
5. Removed unused event
6. No longer re-checking whether vibration should fire outside the service
7. Reworked the addition of hints 
8. Removed duplicate key and renamed everything to `NotificationBanner` instead of `Toast`

XeldarAlz (approved):
All review points addressed. I pushed 6ada0981 on top for two things the fix commit introduced: the new settings keys were declared in L.cs with a capitalised prefix while the catalogs store them lowercase (StringCatalog compares ordinally, so both strings fell back to English in all eight translated languages), and HintIcon drew through the cursor moving CenterIcon overload while gating HoverTooltip.Show behind its own hover test, which stopped the tooltip spring from stepping on exit and killed the fade out.

Release build clean on 6ada0981, CI running, not verified in game yet. Thanks for the thorough second pass.

XeldarAlz:
Please send me a message from discord, so I can give you a contributor role. If you want to keep developing for the project, let me know.

PR #80, opened 2026-08-02
Fix Wine encryption and imported audio playback

BluntEXE:
## What this fixes

Two failures reproduced on Linux with XLCore, wine-xiv, and .NET 10:

- Platform P-256 ECDH fails with `0x80090029`, leaving encryption unsupported.
- `MediaFoundationReader` fails for common imported WAV and MP3 files, making those sounds silent.

## Changes

- Use BouncyCastle for managed P-256 ECDH.
- Preserve interoperability with the existing SPKI public-key format, PKCS#8 private-key format, and `EC1` wrapped-key format.
- Handle malformed server-provided public, private, and ephemeral keys without interrupting processing of other keys or wraps.
- Decode PCM and IEEE-float WAV files with `WaveFileReader`, including WAVE_FORMAT_EXTENSIBLE PCM/float (ffmpeg, Audacity).
- Decode MP3 files with the managed NLayer decoder.
- Continue using `MediaFoundationReader` for compressed WAV and other audio formats it still handles, including as a fallback when `WaveFileReader` can't parse a malformed or exotic WAV header.
- Add native/managed ECDH interoperability tests and managed WAV/MP3 reader tests.
- Add the required BouncyCastle and NLayer third-party notices.
- Rebased onto current master (v0.9.9.7) to pick up everything merged since this was first opened.
- Removed the per-step `AepLog.Info` calls added during development once the fix was confirmed working. The warning on a failed key unwrap stays, that's a real failure signal, not routine confirmation, and is now cached per (scope, generation) so a rotation-orphaned wrap warns once instead of on every thread open and send.

### PKCS#8 encoding note

BouncyCastle produces a standards-compliant 150-byte PKCS#8 representation for newly exported P-256 private keys. .NET's own `ExportPkcs8PrivateKey` produced 138 bytes. Existing 138-byte private keys still import correctly, but keys exported after this change use BouncyCastle's 150-byte encoding.

### Key-clearing note

`KeyVault.ClearKey` now only drops the managed reference to the private key. BouncyCastle's `D` value is an immutable `BigInteger` on the managed heap, so the private scalar survives until GC, where `ECDiffieHellman.Dispose` previously released the CNG handle immediately. This is a reasonable trade-off for Wine support (there is no managed equivalent of a CNG handle to release deterministically), but it's a real change in how long key material can remain resident in memory after logout or vault lock.

### Linux / Wine

- Encryption reached the active state and survived a restart under Wine.
- PCM WAV, IEEE-float WAV, and MP3 files played through XLCore using the managed readers.
- Radio and WebM/Opus song playback continued to work.
- Formats still routed through Media Foundation remain dependent on codecs available in the Wine prefix.

### Build

- Clean Release build after the master merge.
- 478/478 tests passed.

## Test plan

- [x] Reviewer build on Windows to confirm Media Foundation paths are unaffected there. Confirmed by XeldarAlz: clean Release build, 478/478 tests pass.
- [x] Confirm existing (pre-fix) private keys still import under BouncyCastle's PKCS#8 handling. Covered by `ExistingPlatformPrivateKeyImportsWithSamePublicKey`, which imports a real 138-byte .NET-native-exported key.
- [x] Accept WAVE_FORMAT_EXTENSIBLE PCM/float WAV files instead of falling through to Media Foundation and staying silent on Wine.
- [x] Fall back to Media Foundation when `WaveFileReader`'s constructor itself throws on a malformed or exotic WAV, instead of failing past the fallback.


src/Aetherphone/Core/Notifications/SoundEffectPlayer.cs
```diff
@@ -160,6 +161,29 @@ private void RunOnce(string path, float volume)
...
+            ".mp3" => new Mp3FileReaderBase(path,
+                waveFormat => new Mp3FrameDecompressor(waveFormat)),
+            _ => new MediaFoundationReader(path),
+        };
+    }
+
+    private static WaveStream OpenWaveReader(string path)
+    {
+        var reader = new WaveFileReader(path);
+        if (reader.WaveFormat.Encoding is WaveFormatEncoding.Pcm or WaveFormatEncoding.IeeeFloat)
```
XeldarAlz: Blocking: WAVE_FORMAT_EXTENSIBLE PCM fails this check. NAudio reports Encoding=Extensible (runtime type WaveFormatExtraData) for a 40 byte fmt chunk, so those files fall through to MediaFoundation and stay silent on Wine, the exact case this PR fixes. ffmpeg and Audacity emit extensible routinely. Read the sub-format GUID from ExtraData[6..22] and accept PCM (00000001) and float (00000003).

src/Aetherphone/Core/Notifications/SoundEffectPlayer.cs
```diff
@@ -160,6 +161,29 @@ private void RunOnce(string path, float volume)
...
+            ".wav" => OpenWaveReader(path),
+            ".mp3" => new Mp3FileReaderBase(path,
+                waveFormat => new Mp3FrameDecompressor(waveFormat)),
+            _ => new MediaFoundationReader(path),
+        };
+    }
+
+    private static WaveStream OpenWaveReader(string path)
+    {
+        var reader = new WaveFileReader(path);
```
XeldarAlz: Blocking: the MediaFoundation fallback below is unreachable when the constructor itself throws. Verified that new WaveFileReader on a malformed or exotic .wav throws FormatException right here, so playback just logs and gives up where master played the file through Media Foundation on Windows. Wrap the constructor in try/catch and fall back.

src/Aetherphone/Core/Crypto/ConversationKeyStore.cs
```diff
@@ -721,6 +721,10 @@ private void CacheWraps(string scopeId, int currentGeneration, KeyWrapDto[] wrap
             {
                 generations[wrap.Generation] = cek;
             }
+            else
+            {
+                AepLog.Warning($"[Encryption] failed to unwrap key for {scopeId} generation {wrap.Generation}.");
```
XeldarAlz, line 735: Non-blocking: failed unwraps are never cached, only successes reach the ContainsKey skip above, so every EnsureChatKeys / EnsureVelvetKeys / EnsureGramKeys call retries and re-logs. A rotation-orphaned wrap is a permanent failure, so this warns on every thread open and every send for the whole session. Track failed (scope, generation) pairs and log once.

src/Aetherphone/Core/Notifications/SoundEffectPlayer.cs
```diff
@@ -160,6 +161,29 @@ private void RunOnce(string path, float volume)
         }
     }
 
+    internal static WaveStream OpenReader(string path)
+    {
+        return Path.GetExtension(path).ToLowerInvariant() switch
```
XeldarAlz, line 166: Non-blocking: src/Aetherphone/Sounds/README.md:23 still says playback uses Windows Media Foundation, which is now wrong for exactly .wav and .mp3. Worth noting there too that dispatch is by extension now, so a misnamed file (mp3 bytes named .wav) no longer plays where content sniffing used to cover it.

src/Aetherphone/Core/Crypto/KeyVault.cs
```diff
@@ -388,7 +384,6 @@ private bool HasStoredKeyForCurrentUser()
 
     private void ClearKey()
     {
```
XeldarAlz, line 386: Non-blocking: ClearKey now only drops the reference. The BouncyCastle D value is an immutable BigInteger on the managed heap, so the private scalar survives until GC, where ECDiffieHellman.Dispose released the CNG handle at once. Reasonable trade for Wine support, but worth stating in the PR body rather than losing silently.

src/Aetherphone/Core/Crypto/CryptoBox.cs
```diff
@@ -1,149 +1,194 @@
...
+}
+
 internal static class CryptoBox
 {
     public const int CekBytes = 32;
     private const int NonceBytes = 12;
     private const int TagBytes = 16;
     private const string WrapPrefix = "EC1.";
     private static readonly byte[] WrapInfo = Encoding.UTF8.GetBytes("aethernet-cek-v1");
+    private static readonly SecureRandom Random = new();
```
XeldarAlz: Nit: naming this Random shadows System.Random inside CryptoBox, so Random.Shared and friends stop resolving in this file. SecureRandomSource or KeyRandom reads clearer.

src/Aetherphone/Core/Crypto/CryptoBox.cs
```diff
@@ -1,149 +1,194 @@
...
+            var pair = generator.GenerateKeyPair();
+            return new EcPrivateKey((ECPrivateKeyParameters)pair.Private,
+                (ECPublicKeyParameters)pair.Public);
         }
-        catch (CryptographicException)
-        {
-            return null;
-        }
-        catch (PlatformNotSupportedException)
+        catch (ArgumentException)
```
XeldarAlz: Nit: this catch is narrower than ImportPublicKey and ImportPrivateKey, which both use catch (Exception). Since the point of the method is degrading gracefully on hostile platforms, matching them beats letting an unexpected BouncyCastle exception type escape into ProvisionAsync.

XeldarAlz (changes requested):
Verified on Windows: clean Release build of the PR head (a33abdec), 478/478 tests pass. Not run in-game.

The crypto migration holds up under scrutiny. I confirmed separately that BouncyCastle rejects off-curve public keys (ArgumentException, caught in both import paths, so no invalid curve exposure), that the shared static SecureRandom is thread-safe (4000 concurrent generations, 4000 distinct scalars, zero errors), and that the left-padded agreement matches .NET's fixed-length x-coordinate. Both blocking items are in the WAV path, details inline.

Merge hygiene: commit 94b8d22 carries a Co-Authored-By: Claude trailer, which CLAUDE.md forbids, so this needs a squash with a clean message. Directory.Build.props is still 0.9.9.7, which is already released, so it needs a bump or the release tag will not fire.

src/Aetherphone/Core/Notifications/SoundEffectPlayer.cs
BluntEXE: Fixed in `e5bfd78a`. `OpenWaveReader` now reads the sub-format GUID from `WaveFormatExtraData.ExtraData[6..22]` when `Encoding == Extensible` and accepts PCM (`00000001-...`) and IEEE float (`00000003-...`), falling back to Media Foundation for anything else.

src/Aetherphone/Core/Notifications/SoundEffectPlayer.cs
BluntEXE: Fixed in `e5bfd78a`. `OpenWaveReader` now wraps the `new WaveFileReader(path)` constructor in try/catch and falls back to `MediaFoundationReader` on `FormatException`, instead of letting it throw past the fallback.

src/Aetherphone/Core/Crypto/ConversationKeyStore.cs
BluntEXE: Fixed in `e5bfd78a`. `ConversationKeyStore` now tracks failed `(scopeId, generation)` unwraps in a `ConcurrentDictionary` and skips both the unwrap attempt and the warning once a pair has failed once. Cleared in `Clear()` alongside the other caches so a vault relock/unlock retries instead of staying permanently suppressed.

src/Aetherphone/Core/Notifications/SoundEffectPlayer.cs
BluntEXE: Fixed in `e5bfd78a`. Sounds/README.md now says playback dispatches by extension (managed decoders for .mp3/.wav, Media Foundation fallback otherwise), and notes that a misnamed file plays through the wrong decoder where content sniffing used to catch it.

src/Aetherphone/Core/Crypto/KeyVault.cs
BluntEXE: Not changed in code, agreed it's a reasonable trade-off. Added a "Key-clearing note" section to the PR body explaining that the BouncyCastle `D` BigInteger survives on the managed heap until GC, unlike the old CNG handle release.

src/Aetherphone/Core/Crypto/CryptoBox.cs
BluntEXE: Fixed in `e5bfd78a`. Renamed to `SecureRandomSource`.

src/Aetherphone/Core/Crypto/CryptoBox.cs
BluntEXE: Fixed in `e5bfd78a`. Widened `TryGenerateIdentity`, `TryExportPublicKey`, and `TryExportPrivateKey` to `catch (Exception)`, matching `ImportPublicKey`/`ImportPrivateKey`.

XeldarAlz (approved):
All seven review threads are addressed and verified against the code, not just the replies.

Blocking items: WAVE_FORMAT_EXTENSIBLE PCM and float are now accepted via the sub-format GUID, and the WaveFileReader constructor falls back to Media Foundation instead of throwing past it.

Non-blocking items: failed unwraps are cached per (scope, generation) and cleared when the vault leaves Unlocked, Sounds/README.md is corrected, the ClearKey trade-off is documented in the PR body, SecureRandomSource no longer shadows System.Random, and the Try* catches match the Import* ones.

Verified separately on this head: BouncyCastle and .NET PKCS#8 import each other in both directions (150 byte and 138 byte encodings, matching public keys), and SPKI stays 91 bytes on both, so old and new clients interoperate.

Two commits added on top: the 0.9.9.8 version bump was reverted so this ships without firing a release tag, and the .mp3 arm got the same Media Foundation fallback the .wav arm has, since all 13 bundled sounds are .mp3 and NLayer rejecting one would have gone silent on Windows.

CI green: 478/478 tests pass on Windows.

PR #82, opened 2026-08-03
fix(emote): stop the idle scroll firing while an event window is open

YozoraCho:
## What

`PhoneEmoteController.BlockingConditions` now covers the occupied and event condition flags. It previously held eight flags, all of them combat, zoning, or cutscene states. It now holds thirty eight, adding the full `Occupied` family, gathering, crafting, fishing, the retainer summoning bell, trade, materia melding, housing functions, and the various mini-game and transport states.

No other behaviour changes. The gate is still the same linear check in `IsBlocked()`.

## Why

Bug report: a user opened a gathering node, the idle scroll animation fired, and they ended up in a broken state. They stayed in the gather window but could run around. Closing the window by hand left them unable to move or log out, and only a force close of the game recovered it.

`PhoneEmoteController` sends an emote chat command whenever the phone is visible and the player has held still for 400ms. The only things that suppressed it were `InCombat`, the two `BetweenAreas` flags, the cutscene flags, `OccupiedInQuestEvent`, and `Casting`. Nothing stopped it firing into a live gathering event, so the plugin issued a command while the client was mid event and the client's event state came out desynced. The stuck movement and the refused logout are both consistent with the client still believing it is in that event.

Two things make this worth fixing ahead of other open reports:

- `ScrollWhileIdle` defaults to `true` in `Configuration.cs`, so this is on for every user who has not turned it off.
- "Phone open while standing still" is the plugin's main use case, not an edge case.

The reporter guessed that retainers and the market board would have the same problem, and they were right. Those are the same class of modal event window, so this fixes the class rather than just the gathering node.

Known tradeoff: this is a denylist that has to enumerate every occupied state the game has. A future patch that adds a new event window will reopen the same hole. The alternative, gating on whether any game addon holds focus, is harder to leave a gap in but a much larger change, and not worth bundling into a bug fix.

## How to test

The emote fires after 400ms of stillness with the phone visible, so the test is to stand still rather than to race anything.

1. `dotnet build Aetherphone.sln -c Release`
2. Confirm `ScrollWhileIdle` is on (Settings, Immersion, or the idle tile in Control Center).
3. Open the phone, walk to a gathering node, open it, and stand still for several seconds. No emote should fire. Close the window and confirm you can move and log out normally.
4. Repeat at a retainer summoning bell, a market board, a crafting log, and while fishing.
5. Regression, since this gate covers every idle scroll: stand still in a town with nothing open and confirm the scroll still plays, both the loop and the one-shot variant.
6. Regression on the states deliberately left unblocked: sitting on furniture and riding a mount should both still scroll, since neither is an event window and both are normal phone use.

## Checklist

- [ ] Condition flags confirmed live via `/xldata` at a gathering node, retainer bell, and market board
- [x] If this changes user-visible behavior, README is updated (n/a, bug fix only)
- [x] Matches existing style: no `what` comments, no new user-visible strings

### DIFF
--- a/src/Aetherphone/Core/Emote/PhoneEmoteController.cs
+++ b/src/Aetherphone/Core/Emote/PhoneEmoteController.cs
@@ -20,7 +20,16 @@ internal sealed class PhoneEmoteController : IDisposable
     {
         ConditionFlag.InCombat, ConditionFlag.BetweenAreas, ConditionFlag.BetweenAreas51,
         ConditionFlag.OccupiedInCutSceneEvent, ConditionFlag.WatchingCutscene, ConditionFlag.WatchingCutscene78,
-        ConditionFlag.OccupiedInQuestEvent, ConditionFlag.Casting,
+        ConditionFlag.OccupiedInQuestEvent, ConditionFlag.Casting, ConditionFlag.OccupiedInEvent,
+        ConditionFlag.Gathering, ConditionFlag.Crafting, ConditionFlag.TradeOpen, ConditionFlag.ExecutingCraftingAction,
+        ConditionFlag.Unconscious, ConditionFlag.MeldingMateria, ConditionFlag.OperatingSiegeMachine,
+        ConditionFlag.CarryingItem, ConditionFlag.CarryingObject, ConditionFlag.Mounting, ConditionFlag.Mounting71,
+        ConditionFlag.ParticipatingInCustomMatch, ConditionFlag.PlayingLordOfVerminion, ConditionFlag.ChocoboRacing,
+        ConditionFlag.PlayingMiniGame, ConditionFlag.Performing, ConditionFlag.Transformed,
+        ConditionFlag.UsingHousingFunctions, ConditionFlag.Occupied, ConditionFlag.Occupied30, ConditionFlag.Occupied33,
+        ConditionFlag.Occupied38, ConditionFlag.Occupied39, ConditionFlag.OccupiedSummoningBell,
+        ConditionFlag.ExecutingGatheringAction, ConditionFlag.PreparingToCraft, ConditionFlag.BeingMoved,
+        ConditionFlag.LoggingOut, ConditionFlag.Fishing,
     };
 
     private readonly Configuration configuration;
--- a/src/Aetherphone/Core/Emote/PhoneEmoteController.cs
+++ b/src/Aetherphone/Core/Emote/PhoneEmoteController.cs
@@ -23,7 +23,7 @@ internal sealed class PhoneEmoteController : IDisposable
         ConditionFlag.OccupiedInQuestEvent, ConditionFlag.Casting, ConditionFlag.OccupiedInEvent,
         ConditionFlag.Gathering, ConditionFlag.Crafting, ConditionFlag.TradeOpen, ConditionFlag.ExecutingCraftingAction,
         ConditionFlag.Unconscious, ConditionFlag.MeldingMateria, ConditionFlag.OperatingSiegeMachine,
-        ConditionFlag.CarryingItem, ConditionFlag.CarryingObject, ConditionFlag.Mounting, ConditionFlag.Mounting71,
+        ConditionFlag.CarryingItem, ConditionFlag.CarryingObject, ConditionFlag.EditingPortrait,
         ConditionFlag.ParticipatingInCustomMatch, ConditionFlag.PlayingLordOfVerminion, ConditionFlag.ChocoboRacing,
         ConditionFlag.PlayingMiniGame, ConditionFlag.Performing, ConditionFlag.Transformed,
         ConditionFlag.UsingHousingFunctions, ConditionFlag.Occupied, ConditionFlag.Occupied30, ConditionFlag.Occupied33,
@@ -72,14 +72,15 @@ internal sealed class PhoneEmoteController : IDisposable
             return;
         }
 
+        var now = Environment.TickCount64;
         var player = objectTable.LocalPlayer;
         if (player is null || IsBlocked())
         {
             hasSample = false;
+            lastCastMilliseconds = now;
             return;
         }
 
-        var now = Environment.TickCount64;
         if (IsBusy(player.Address))
         {
             lastCastMilliseconds = now;


src/Aetherphone/Core/Emote/PhoneEmoteController.cs
```diff
@@ -20,7 +20,16 @@ internal sealed class PhoneEmoteController : IDisposable
     {
         ConditionFlag.InCombat, ConditionFlag.BetweenAreas, ConditionFlag.BetweenAreas51,
         ConditionFlag.OccupiedInCutSceneEvent, ConditionFlag.WatchingCutscene, ConditionFlag.WatchingCutscene78,
-        ConditionFlag.OccupiedInQuestEvent, ConditionFlag.Casting,
+        ConditionFlag.OccupiedInQuestEvent, ConditionFlag.Casting, ConditionFlag.OccupiedInEvent,
+        ConditionFlag.Gathering, ConditionFlag.Crafting, ConditionFlag.TradeOpen, ConditionFlag.ExecutingCraftingAction,
```
XeldarAlz, line 24: Blocking: IsBlocked() returns at :67, before `lastCastMilliseconds = now` at :76. Crafting, Performing and Carrying used to exit through IsBusy and got a 2500ms grace once the state cleared; they now exit here and need only 400ms of stillness. That makes the trailing edge of an event window tighter, not safer. Push lastCastMilliseconds in the blocked branch too.

src/Aetherphone/Core/Emote/PhoneEmoteController.cs
```diff
@@ -20,7 +20,16 @@ internal sealed class PhoneEmoteController : IDisposable
...
+        ConditionFlag.Unconscious, ConditionFlag.MeldingMateria, ConditionFlag.OperatingSiegeMachine,
+        ConditionFlag.CarryingItem, ConditionFlag.CarryingObject, ConditionFlag.Mounting, ConditionFlag.Mounting71,
```
XeldarAlz: Nit: Mounting and Mounting71 add nothing. Mount-up is a cast, so Casting already blocks it, and once mounted CharacterModes.Mounted makes IsBusy true at :193. Same for InThatPosition via InPositionLoop, so description step 6 cannot pass: neither mounts nor furniture scroll today.

src/Aetherphone/Core/Emote/PhoneEmoteController.cs
```diff
@@ -20,7 +20,16 @@ internal sealed class PhoneEmoteController : IDisposable
...
+        ConditionFlag.ParticipatingInCustomMatch, ConditionFlag.PlayingLordOfVerminion, ConditionFlag.ChocoboRacing,
+        ConditionFlag.PlayingMiniGame, ConditionFlag.Performing, ConditionFlag.Transformed,
+        ConditionFlag.UsingHousingFunctions, ConditionFlag.Occupied, ConditionFlag.Occupied30, ConditionFlag.Occupied33,
+        ConditionFlag.Occupied38, ConditionFlag.Occupied39, ConditionFlag.OccupiedSummoningBell,
+        ConditionFlag.ExecutingGatheringAction, ConditionFlag.PreparingToCraft, ConditionFlag.BeingMoved,
```
XeldarAlz, line 31: Non-blocking: EditingPortrait (100) is the same class of modal you are closing off, and an emote firing into the portrait editor changes the pose being edited. Worth adding while the list is open.

src/Aetherphone/Core/Emote/PhoneEmoteController.cs
```diff
@@ -20,7 +20,16 @@ internal sealed class PhoneEmoteController : IDisposable
...
+        ConditionFlag.LoggingOut, ConditionFlag.Fishing,
     };
```
XeldarAlz, line 33: Non-blocking: ShortcutRunner.cs:49 keeps a second ConditionFlag denylist guarding the same ChatSender, still at five flags. After this PR a shortcut macro can push a command into a gathering node exactly as the idle scroll did. One shared table beats two that drift.

XeldarAlz (changes requested):
Built Release clean, 470/470 tests pass. Not verified in-game, so the flag semantics below come from the Dalamud ConditionFlag enum and FFXIVClientStructs CharacterModes, not from /xldata.

The denylist itself reads right and the gathering fix is real. One blocking issue: the new early return skips the recast grace these states used to get.

YozoraCho:
Pushed. Fixed the recast regression, dropped the two Mounting flags, added `EditingPortrait`, and verified the main flags in-game with `/xldata`.

src/Aetherphone/Core/Emote/PhoneEmoteController.cs
YozoraCho: Hoisted `now` above the player check and set `lastCastMilliseconds` in the blocked branch, so anything leaving a blocked state needs the full 2500ms again instead of dropping to 400ms.

src/Aetherphone/Core/Emote/PhoneEmoteController.cs
YozoraCho: Dropped both, agreed. Mount-up is a cast, and `CharacterModes.Mounted` covers the rest. Test step 6 was wrong for the same reason.

src/Aetherphone/Core/Emote/PhoneEmoteController.cs
YozoraCho: Added. Had not considered the pose being edited, that is the same failure mode as the gathering node.

src/Aetherphone/Core/Emote/PhoneEmoteController.cs
YozoraCho: Want to take this in a follow-up, and I think the shared table needs some design first. `ChatSwallowedConditions` feeds `hold.Advance()` in `TryClearGate`. so it's a wait-and-retry gate rather than a denylist. Moving these flags into it would make a shortcut hang for `MaxWaitSeconds` at a gathering node and then fail with `GameBusy`, instead of simply not firing. The two lists mean different things today: "do not fire" vs "wait until chat is accepted."

Happy to unify them once the shortcut side has a skip path, but that felt like more than this fix should carry.

YozoraCho:
Verified the flags live rather than from the enum.

`OccupiedSummoningBell` (50) at the bell, `Gathering` (6) and `ExecutingGatheringAction` (42) at a node, with 42 appearing only for a split second, consistent with the swing animations.

One thing worth recording: ID 85 (`Unknown85`) is also set at a node, and `Gathering` drops before it does, so 85 covers the tail after the window closes. That tail is the same window your recast comment was about. I measured it at under 2.5s, so the restored recast grace already covers it and I've left 85 out. I would rather not denylist a flag I can't pin the meaning of when the cooldown handles it. Noting it here in case it comes up again, and if you know what sets 85, I'd like to know as well.

XeldarAlz (approved):
Recast grace restored and the flags verified live, so this is good to go. Thanks for the /xldata pass and for writing up the Unknown85 tail, that's useful to have on the record.

Agreed on leaving ShortcutRunner for a follow-up. ChatSwallowedConditions is a wait-and-retry gate, not a denylist, so folding these flags in would stall a shortcut for MaxWaitSeconds and then fail with GameBusy. It needs a skip path first.

PR #83, opened 2026-08-03
feat(housing): add the Housing app, a map-first plot browser and lottery tracker

YozoraCho:
## What

Adds `Housing`, an opt-in app that shows open housing plots on the real in-game district maps and tracks lottery phases, with a watchlist and reminders.

## What it does

- Map-first browser over Mist, The Lavender Beds, The Goblet, Shirogane and
  Empyreum, on any world.
- Plot markers carry size, lottery phase, price, eligibility and reported
  entries. Shape encodes size and the inner mark encodes phase, so state is
  legible without relying on color.
- Watchlist plus reminders ahead of a phase deadline, delivered through the
  existing notification pipeline.
- Filter drawer over size, phase, eligibility, division, freshness and entry
  count, plus a list view sorted by entries, scan age, size, price or location.

## Data

Housing reads Aetherphone's housing service, which polls and caches the public PaissaDB API once for all users rather than each client polling it. The PaissaHouse plugin is not required, and nothing is read from another plugin's configuration.

Polling is deliberately conservative: 20 minutes by default with a 15 minute floor, only while the app is on screen, and hourly in the background only when a watchlist or a pending reminder makes it useful. Reminders are local time arithmetic and need no network. On failure the app falls back to its disk cache and labels the data as cached.

## Maps

District maps and plot positions are read from the player's own game installation via `TerritoryType`, `Map` and `HousingMapMarkerInfo`. Textures are referenced by game path, so nothing is extracted or redistributed. Marker subrows span two maps per district, so the main division and subdivision are kept as separate plans with a switch between them. When local game data cannot supply a plan the map says so and points at the list rather than drawing invented geometry.

## Shared code

One change outside the app: `ChipRail` gains a `Draw(Rect row, ...)` overload beside the existing cursor-based one, with an `overlay` flag for chips inside absolutely positioned overlay surfaces. The original signature delegates to it, so no existing caller changes. This let Housing drop a duplicate chip implementation and use the shared rail for all seven of its chip rows.

`SettingsRow.Info` now gives the label a width floor. Previously a long value consumed the whole row budget and the label marquee-scrolled to gibberish ("Endpoint" rendering as "dpoin"). Rows with short values lay out unchanged.

## Localization

185 new keys in `L.cs` and all nine catalogs, added in this commit. The eight non-English sets are machine translated and would benefit from a native speaker pass on the FFXIV terminology, particularly ward, plot, division and lottery phase.

## Notes for review

- Input hit testing is the riskiest area and is not unit testable. Worth a
  click-through pass in game: with the filter drawer open, the map top bar,
  zoom controls and plot sheet underneath should all be inert.
- The endpoint is a compile-time constant with no config override, so it can be
  moved without stranding saved configs.
- Map diagnostics are behind `#if DEBUG`.

## Checklist

- [x] Verified against live API responses: district ids, 0-indexed ward and plot normalisation, and `purchase_system` bit flags all decode correctly.


src/Aetherphone/Core/Housing/HousingEndpoints.cs
```diff
@@ -0,0 +1,16 @@
+using System.Globalization;
+
+namespace Aetherphone.Core.Housing;
+
+internal static class HousingEndpoints
+{
+    public const string BaseUrl = "https://housing-api.yozoracho.dev/api/v1";
```
XeldarAlz, line 7: Blocking: every install would poll a contributor-owned domain, while `housing.dataSourceNotice` tells users this is "Aetherphone's housing service" and the Diagnostics row prints this URL back to them. Please either host it under a project-owned domain or change the copy to name the real operator.

src/Aetherphone/Core/Housing/HousingFormat.cs
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        return Loc.T(L.Housing.AgeDays, (int)age.TotalDays);
+    }
+
+    public static string ExactLocalTime(DateTime utcMoment) =>
+        utcMoment == default
+            ? Loc.T(L.Housing.TimeUnknown)
+            : DateTime.SpecifyKind(utcMoment, DateTimeKind.Utc).ToLocalTime()
+                .ToString("d MMM yyyy HH:mm", Loc.Culture);
```
XeldarAlz: Blocking: hand-formatted `HH:mm` ignores the user's 12/24 hour preference, and CLAUDE.md requires all clock text through `TimeText.Clock`. Follow `ModerationNoticeText.cs:170`: `local.ToString("d MMM yyyy", Loc.Culture) + " " + TimeText.Clock(local)`.

src/Aetherphone/Apps/AppStore/AppStoreApp.Detail.cs
```diff
@@ -167,6 +167,7 @@ private float DrawInformation(Vector2 origin, float width, float top, IPhoneApp
     private static string DeveloperName(IPhoneApp app) => app.Id switch
     {
         "health" => "Yozora",
+        "housing" => "Yozora",
```
XeldarAlz, line 170: Blocking: `AppStoreCatalog.cs` has no `["housing"]` entry, so `For("housing")` returns the generic fallback copy and files the app under Tools instead of Adventure. The new `L.StoreCopy.HousingSub`/`HousingBody` strings (plus their nine catalog entries) have no reader at all.

src/Aetherphone/Apps/Housing/HousingApp.Details.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            var size = plot?.Size ?? (HousingPlotSize)(watchRecord?.Size ?? 0);
+            var phase = plot?.Phase ?? (HousingLotteryPhase)(watchRecord?.Phase ?? 0);
+            var title = Loc.T(L.Housing.PlotTitle, key.Plot, HousingFormat.SizeLabel(size));
+            Typography.Draw(drawList, new Vector2(left, y), Typography.FitText(title, width, TextStyles.Title2),
+                ui.TitleInk, TextStyles.Title2);
+            y += Typography.LineHeight(TextStyles.Title2) + 2f * scale;
+            var place = HousingFormat.Place(HousingDistricts.Name(key.DistrictId), key.Ward);
+            Typography.Draw(drawList, new Vector2(left, y),
+                Typography.FitText(string.Concat(housing.WorldName, " · ", place), width, TextStyles.Subheadline),
```
XeldarAlz: Blocking: this prints the currently selected world for a plot keyed to `key.WorldId`, so a watchlist entry from another world is labelled with the wrong world name. The Region and Data Center rows further down already resolve from `key.WorldId`. Line 138 has the same problem: it stores `housing.WorldName` on the watch record.

src/Aetherphone/Apps/Housing/HousingApp.List.cs
```diff
@@ -0,0 +1,138 @@
...
+            card.End();
+            ImGui.Dummy(new Vector2(0f, 12f * scale));
+        }
+    }
+
+    private void OpenFromList(HousingPlot plot)
+    {
+        if (plot.Key.Ward != housing.Ward)
+        {
+            housing.SelectWard(plot.Key.Ward);
```
XeldarAlz, line 93: Blocking: `OpenFromList` only corrects the ward. Reached from the watchlist with a plot from another world or district (its snapshot is still in `snapshots`), it changes the current district's ward and pops to a map that does not contain the plot. Switch world and district too, or push Details when the key does not match the current pair.

src/Aetherphone/Apps/Housing/HousingChrome.cs
```diff
@@ -0,0 +1,439 @@
...
+internal static class HousingChrome
+{
+    public const float ChipHeight = 22f;
+    public const float StatRowHeight = 26f;
+
+    private static readonly Vector4 White = new(1f, 1f, 1f, 1f);
+    private static readonly Vector4 Transparent = new(0f, 0f, 0f, 0f);
+
+    public static bool Hover(Vector2 min, Vector2 max, bool overlay) =>
+        overlay ? ImGui.IsMouseHoveringRect(min, max) : UiInteract.Hover(min, max);
```
XeldarAlz: Non-blocking: the overlay path drops the window gate. `UiInteract.HoverWindowOnly` is the established equivalent, and it is what this PR's own `ChipRail` overlay path uses. As written, overlay chrome hover-highlights and swaps the cursor while another window sits on top; only the click stays gated.

src/Aetherphone/Apps/Housing/HousingApp.cs
```diff
@@ -0,0 +1,563 @@
...
+        drawView = DrawView;
+    }
+
+    public string Id => HousingService.AppId;
+
+    public string DisplayName => Loc.T(L.Apps.Housing);
+
+    public string Glyph => "Ho";
+
+    public int BadgeCount => housing.Watch.PendingReminderCount;
```
XeldarAlz: Non-blocking: `PendingReminderCount` counts reminders that have not fired yet, so the dot shows while waiting and clears the moment the notification actually arrives. Every other app badges unread or ready work (`AnnouncementsApp` unread, `ActivityApp` ventures ready).

src/Aetherphone/Core/Housing/HousingWatchStore.cs
```diff
@@ -0,0 +1,351 @@
...
+
+        reminder.Phase = (byte)plot.Phase;
+        reminder.PhaseEndUnix = phaseEnd;
+        reminder.Entries = plot.Entries;
+        reminder.LastSeenUnix = ToUnix(plot.LastSeenUtc);
+        reminder.Notified = false;
+        return true;
+    }
+
+    public void MarkNotified(HousingReminderRecord reminder)
```
XeldarAlz, line 324: Non-blocking: nothing ever removes a fired reminder. `Notified` flips and the record stays in `configuration.HousingReminders` for the life of the install, re-serialized on every `Save()`. Either drop the record here or prune entries whose `PhaseEndUnix` is already past.

src/Aetherphone/Core/Housing/HousingRestProvider.cs
```diff
@@ -0,0 +1,234 @@
...
+    {
+        if (worldId == 0 || districtId == 0)
+        {
+            return null;
+        }
+
+        using (await throttle.EnterAsync(token).ConfigureAwait(false))
+        {
+            var status = 0;
+            var payload = await http.GetJsonAsync(HousingEndpoints.World(BaseUrl, worldId),
```
XeldarAlz, line 80: Non-blocking: the description says requests are conditional and repeats collapse to 304s, but `HttpService.GetJsonAsync` sends a plain GET and nothing here keeps an ETag. A 304 would also deserialize to null and be counted as a failure, which trips the backoff. Worth implementing or dropping from the description.

src/Aetherphone/Apps/Housing/HousingGlyphs.cs
```diff
@@ -0,0 +1,160 @@
...
+        Fill(drawList, packedInk, roof);
+        var bodyMin = new Vector2(center.X - radius * 0.70f, center.Y - radius * 0.12f);
+        var bodyMax = new Vector2(center.X + radius * 0.70f, center.Y + radius * 0.86f);
+        drawList.AddRectFilled(bodyMin, bodyMax, packedInk, radius * 0.16f);
+        var doorMin = new Vector2(center.X - radius * 0.22f, center.Y + radius * 0.24f);
+        var doorMax = new Vector2(center.X + radius * 0.22f, center.Y + radius * 0.86f);
+        drawList.AddRectFilled(doorMin, doorMax, packedHole, radius * 0.10f);
+    }
+
+    public static void Placard(ImDrawListPtr drawList, Rect bounds, Vector4 board, Vector4 edge, float scale)
```
XeldarAlz: Non-blocking: `Placard`, `HexagonOutline` and `DiamondOutline` have no call sites. Same for `HousingPlan.Covers` and `ShowsSubdivision`, `HousingDistricts.IndexOf`, `HousingRegions.OrderOf`, and `HousingCache`'s `IHousingDataProvider` implementation (the service calls `Read`/`ReadWorlds` directly). YAGNI.

src/Aetherphone/Apps/Housing/HousingApp.Sheet.cs
```diff
@@ -0,0 +1,335 @@
...
+        if (HousingChrome.PillButton(confirmRect, confirmLabel, true, ui, true))
+        {
+            var minutes = choices[Math.Clamp(reminderChoice, 0, choices.Length - 1)];
+            if (housing.Watch.SetReminder(plot, housing.WorldName, minutes))
+            {
+                configuration.HousingReminderMinutes = minutes;
+                configuration.Save();
+                reminderPickerOpen = false;
+                ShowToast(Loc.T(L.Housing.ReminderConfirmed, HousingFormat.LeadTime(minutes),
+                    HousingFormat.PhaseLabel(plot.Phase).ToLower(Loc.Culture),
```
XeldarAlz: Nit: lowercasing a translated label breaks languages that capitalise nouns. German turns "Bewerbungsphase" into "bewerbungsphase" in this toast. Better to keep the label as translated, or give the confirmation its own sentence-cased key per phase.

XeldarAlz (changes requested):
Release build clean and 470/470 tests pass on 95feff09; not exercised in game. Localization is in lockstep: 185 keys in L.cs and all nine catalogs, no missing keys and no placeholder drift. Findings are inline, five blocking and the rest not.

XeldarAlz (approved):
lgtm

src/Aetherphone/Core/Housing/HousingEndpoints.cs
XeldarAlz: Accepted as is for now: we will run on YozoraCho's API for the time being, so the endpoint stays. Moving it to a project-owned domain is tracked as follow-up, and `BaseUrl` is a compile-time constant, so the swap will not strand saved configs.

src/Aetherphone/Core/Housing/HousingRestProvider.cs
XeldarAlz: Dropped the claim from the description rather than implementing conditional requests, since the 20 minute floor plus the on-screen-only gate already keeps the request rate low. ETag handling can come with the next `HttpService` pass.

src/Aetherphone/Apps/Housing/HousingChrome.cs
XeldarAlz: Fixed in 73d708d: the overlay path now uses `UiInteract.HoverWindowOnly`, matching the `ChipRail` overlay path.

XeldarAlz (approved):
All eleven findings are settled. The four blocking code issues were fixed in a209d78 (TimeText.Clock, the missing AppStoreCatalog entry, world resolution via WorldNameOf(key.WorldId), and the cross-world/district route in OpenFromList), along with the badge count, reminder pruning, the dead code and the lowercased phase label. 73d708d gates overlay chrome hover through UiInteract.HoverWindowOnly.

Two are accepted rather than changed: the endpoint stays on YozoraCho's API for now, and the conditional-request claim was removed from the description instead of implementing ETags. Both are noted inline. Build and guards are green.

PR #85, opened 2026-08-04
fix(housing): stop chip labels wrapping and tidy the filter drawer rows

YozoraCho:
## What

Fixes chip labels wrapping inside their own pill in the Housing filter drawer, lifts the chip rows off the section rule they were sitting on, and shortens "Other phases" to "Other".

## Why

`Typography.DrawCentered` applies `AutoWrapWidth`, which measures from the text centre to the nearer window content edge and wraps if the string is wider than double that. For the last chip in a rail that budget collapses, so a single line label wrapped to two lines inside a pill that `ChipWidth` had already sized for one line. "Other phases" rendered as two lines spilling out of its own chip. `ChipRail` knows its own chip bounds, so the window relative heuristic is wrong for it.

Separately, `SectionLabel` draws its hairline at `LineHeight(Caption1) + 3f * scale` and `DrawChipSection` started the chip row at `+ 4f * scale`, one pixel below it, so every row looked welded to the rule above it.

With the wrapping fixed the phase row still overflowed, so the third chip is now "Other". The section header already reads LOTTERY PHASE, so the noun is redundant. The shortened forms keep agreement with the noun they drop, so `Otras` and `Outras` stay feminine plural.

## How to test

1. `/phone`, open Housing, tap the filter button to open the drawer.
2. Check the Lottery phase row: all three chips should sit on one line each, fully inside their pills, with no text spilling past a chip edge.
3. Check every section: the chips should sit clear of the rule under the section label, not touching it.
4. Switch to German in Settings and reopen the drawer. Long phase labels may still pan the row, which is the intended `ChipRail` behaviour, but nothing should render as a two line chip.

## Checklist

## Note on shared code

One change outside the app: `ChipRail.DrawChip` swaps `Typography.DrawCentered` for `Typography.Measure` plus `Typography.Draw`. No signature changes. This also affects the YellowPages, Muster and Housing reminder rails, which had the same latent wrapping bug and none of which relied on wrapping.

Localization is in lockstep: `housing.filterOtherPhases` is updated in `L.cs` and all nine catalogs in this commit.

### DIFF
--- a/src/Aetherphone/Apps/Housing/HousingApp.Filters.cs
+++ b/src/Aetherphone/Apps/Housing/HousingApp.Filters.cs
@@ -28,7 +28,7 @@ internal sealed partial class HousingApp
     private readonly bool[] reminderActive = new bool[HousingDefaults.ReminderChoices.Length];
 
     private static float FilterSectionHeight(float scale) =>
-        Typography.LineHeight(TextStyles.Caption1) + (ChipRail.RowHeight + 12f) * scale;
+        Typography.LineHeight(TextStyles.Caption1) + (ChipRail.RowHeight + Metrics.Space.Sm * 2f) * scale;
 
     private static float FilterDrawerHeightFor(float scale)
     {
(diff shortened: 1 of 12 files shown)


XeldarAlz (approved):
lgtm

PR #86, opened 2026-08-05
fix(social): stop feeds jumping to the wrong scroll position

Kiro-XVI:
## Summary
- After posting in Aethergram or Chirper, the feed refresh prepended new items without resetting scroll, landing you mid feed instead of at the top where the new post is.
- Aethergram's bottom nav tabs (Home, Search, Activity, Profile) all drew through the same AppSurface scroll region ID, so ImGui carried one tab's scroll offset into another; switching to Activity after scrolling down Home opened it mid list instead of at the top.

## Test plan
- [ ] Build succeeds, full test suite passes (293/293 locally)
- [ ] Scroll partway down Aethergram Home, post something, confirm the feed lands at the top on return
- [ ] Scroll partway down Aethergram Home, switch to Activity, confirm it opens at the top; check Search and Profile too
- [ ] Repeat the compose scroll check in Chirper


src/Aetherphone/Apps/Aethergram/AethergramApp.cs
```diff
@@ -295,20 +295,23 @@ private void DrawRoot(Rect area)
 
         var navRect = new Rect(new Vector2(area.Min.X, area.Max.Y - BottomNavHeight * scale), area.Max);
         var tabArea = new Rect(contentArea.Min, new Vector2(area.Max.X, navRect.Min.Y));
-        switch (activeTab)
+        using (ImRaii.PushId((int)activeTab))
```
XeldarAlz, line 387: Non-blocking: the same shared surface sits one level down. ForYou and Following both draw through DrawFeedList (AethergramApp.cs:510) under one `##appSurface` id, so switching scope keeps the other scope's offset. Pushing `(int)scope` there closes the pair.

src/Aetherphone/Apps/Chirper/ChirperApp.Compose.cs
```diff
@@ -25,6 +25,7 @@ private void DrawCompose(Rect area)
             quoteTargetId = null;
             store.RefreshFeed(SocialFeedScope.ForYou);
             store.RefreshFeed(SocialFeedScope.Following);
+            feedScrollTopPending = true;
```
XeldarAlz, line 25: Non-blocking: both scopes refresh, but the single flag is consumed by whichever list draws, so the scope you were not on still lands mid feed with a new post above it. A pending flag per scope covers both. Same in Aethergram.

src/Aetherphone/Apps/Aethergram/AethergramApp.Compose.cs
```diff
@@ -58,6 +58,7 @@ private void DrawCompose(Rect area)
                 caption = string.Empty;
                 store.RefreshFeed(SocialFeedScope.ForYou);
                 store.RefreshFeed(SocialFeedScope.Following);
+                feedScrollTopPending = true;
```
XeldarAlz, line 97: Nit: story compose reaches this branch too (`!composeAvatarMode` is true in story mode), so posting a story now yanks the feed to top. The tray is up there, so it may be intended, just calling it out since the description does not.

XeldarAlz (commented):
Built Release clean at 0de6cacb, 67/67 tests pass, not verified in game. Both fixes still apply to current master and are still needed there. Note the branch is 358 commits behind master and only carries 67 tests, so the 293/293 in the test plan is not from this checkout. Three notes inline, none blocking.

src/Aetherphone/Apps/Aethergram/AethergramApp.cs
```diff
@@ -384,20 +384,23 @@ private void DrawRoot(Rect area)
...
```
XeldarAlz, line 387: Nit: `ImRaii.PushId` returns a class, so this allocates one disposable per frame in draw code. Keeping it because SceneCompositor.DrawLayer already does the same on this exact path, so a raw PushID/PopID pair here would be inconsistent for 24 bytes.

src/Aetherphone/Apps/Chirper/ChirperApp.cs
```diff
@@ -262,6 +262,7 @@ private void DrawHome(Rect area)
         {
             activeScope = (SocialFeedScope)selected;
             actions.Reset();
+            feedScrollTopPending = true;
```
XeldarAlz, line 265: Non-blocking, pushed as a follow-up commit: For You and Following also share one scroll region, so switching scopes dropped you at the offset from the list you left. Same flag, same treatment Velvet already gives its feed scopes (VelvetShell.Feed.cs:53). Mirrored in AethergramApp.cs:686.

XeldarAlz (approved):
Verified on the merged branch: `dotnet build -c Release` clean, 517/517 tests pass. Not run in game.

The diagnosis is right and the fix is the same one SceneCompositor already uses per router layer, so the tabs now get their own scroll region. I merged master in (the branch was 404 commits behind) and pushed one follow-up commit for the scope switch, see the inline note.

PR #91, opened 2026-08-06
fix(venues): correct ffxiv venues fallback link

RayaSerahill:
## What

Changed FFXIV Venues fallback links in the Venues app to use /venue/{id} instead of /{id}.

## Why

Previous base link sends users to a 404 page on the FFXIV Venues website

## Checklist


XeldarAlz (approved):
lgtm

PR #92, opened 2026-08-06
fix(chrome): correct camera landscape sizing and case rotation

RayaSerahill:
## What

  - Fixed camera landscape mode so the phone frame, case art, and screen interior keep consistent sizing after rotation.
  - Made the usable screen area stay the same size when the phone turns sideways.
  - Rotated hardware button placement and drawing for landscape, so shell controls stay aligned with the rails.
  - Kept case-art overflow margins based on the short side to avoid oversized landscape frame art.
  - Rotated case art on landscape mode to match the rotation of the hardware buttons

## Why

  Camera landscape swaps the phone window dimensions, but the chrome geometry still treated the X axis as the narrow phone side. That made the frame/case art expand and made the interior lose rail-width pixels horizontally in landscape. Aligning the rail and case-art math with the rotated chassis keeps the phone visually the same size and prevents the camera frame from shrinking or ballooning.

## How to test

 - Open the phone
 - Go to camera
 - Flip to landscape

## Checklist

### DIFF
--- a/src/Aetherphone/Core/Theme/ChassisGeometry.cs
+++ b/src/Aetherphone/Core/Theme/ChassisGeometry.cs
@@ -31,11 +31,22 @@ internal readonly struct ChassisGeometry
 
     public static ChassisGeometry Device(Rect window, PhoneTheme theme, float scale)
     {
-        var rail = theme.RailWidth * scale;
-        var body = new Rect(new Vector2(window.Min.X + rail, window.Min.Y),
-            new Vector2(window.Max.X - rail, window.Max.Y));
+        var body = BodyRect(window, theme, scale);
         return new ChassisGeometry(body, theme.DeviceRounding * scale, theme.MetalWidth * scale,
             theme.GlassWidth * scale);
+    }
+
+    public static Rect BodyRect(Rect window, PhoneTheme theme, float scale)
+    {
+        var rail = theme.RailWidth * scale;
+        if (window.Width > window.Height)
+        {
+            return new Rect(new Vector2(window.Min.X, window.Min.Y + rail),
+                new Vector2(window.Max.X, window.Max.Y - rail));
+        }
+
+        return new Rect(new Vector2(window.Min.X + rail, window.Min.Y),
+            new Vector2(window.Max.X - rail, window.Max.Y));
     }
 
     public static ChassisGeometry Puck(Rect body) =>
--- a/src/Aetherphone/Core/Theme/ChassisGeometry.cs
+++ b/src/Aetherphone/Core/Theme/ChassisGeometry.cs
@@ -39,7 +39,7 @@ internal readonly struct ChassisGeometry
     public static Rect BodyRect(Rect window, PhoneTheme theme, float scale)
     {
         var rail = theme.RailWidth * scale;
-        if (window.Width > window.Height)
+        if (window.IsLandscape())
         {
             return new Rect(new Vector2(window.Min.X, window.Min.Y + rail),
                 new Vector2(window.Max.X, window.Max.Y - rail));
--- a/src/Aetherphone/Windows/Components/DeviceChrome.cs
+++ b/src/Aetherphone/Windows/Components/DeviceChrome.cs
@@ -7,6 +7,11 @@ namespace Aetherphone.Windows.Components;
 
 internal static class DeviceChrome
 {
+    private const float SideButtonStartFraction = 0.250f;
+    private const float SideButtonLengthFraction = 0.108f;
+    private const float MuteButtonStartFraction = 0.205f;
+    private const float LockButtonStartFraction = 0.315f;
+    private const float ShortButtonLengthFraction = 0.082f;
     private const float ChamferFraction = 0.4f;
 
     private const float MaskGrow = 0.5f;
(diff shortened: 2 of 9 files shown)


XeldarAlz (commented):
Math checks out. Landscape screen is now the exact transpose of portrait, and the case-art UVs rotate CCW to match the new button rotation. `QuadClipped` is shared with `PhoneCasePreview`, but that passes `landscape: false`, so thumbnails are safe. Release build clean.

Main note: orientation is already known state (`PhoneShell.LandscapeActive`, `CaseArt.IsLandscape`), and this PR re-derives it from geometry seven times. Three of those compare button bounds instead of the chassis. Comments inline.

Follow-up ticket, not this PR: `PhoneWindow.OrientedSize` animates the rotation, so every hard `>` compare makes the buttons teleport on the crossover frame.

src/Aetherphone/Core/Theme/ChassisGeometry.cs
```diff
@@ -31,13 +31,24 @@ private ChassisGeometry(Rect body, float bodyRadius, float metalWidth, float gla
...
-            new Vector2(window.Max.X - rail, window.Max.Y));
+        var body = BodyRect(window, theme, scale);
         return new ChassisGeometry(body, theme.DeviceRounding * scale, theme.MetalWidth * scale,
             theme.GlassWidth * scale);
     }
 
+    public static Rect BodyRect(Rect window, PhoneTheme theme, float scale)
+    {
+        var rail = theme.RailWidth * scale;
+        if (window.Width > window.Height)
```
XeldarAlz: `CaseArt.IsLandscape(Rect)` already exists. This PR adds seven more copies of the same comparison. Promote it to `Rect.IsLandscape` and call that everywhere.

src/Aetherphone/Windows/Components/DeviceChrome.cs
```diff
@@ -24,6 +21,13 @@ public static Rect SideButtonRect(Rect window, in ChassisGeometry chassis)
     {
         var scale = UiScale.Current;
         var device = chassis.Body;
+        if (device.Width > device.Height)
```
XeldarAlz: The rail fractions are duplicated per branch across all three methods: `0.250f`/`0.108f`, `0.205f`/`0.082f`, `0.315f`/`0.082f`. Hoist to consts so the two branches cannot drift.

src/Aetherphone/Windows/Components/HardwareButton.cs
```diff
@@ -40,8 +48,13 @@ public static void Draw(ImDrawListPtr drawList, Rect bounds, PhoneTheme theme, R
     private static void Boss(ImDrawListPtr drawList, Rect bounds, PhoneTheme theme, float scale)
     {
         var pad = 2.4f * scale;
-        var min = new Vector2(bounds.Min.X, bounds.Min.Y - pad);
-        var max = new Vector2(bounds.Max.X, bounds.Max.Y + pad);
+        var horizontal = bounds.Width > bounds.Height;
```
XeldarAlz: `Draw` already has an authoritative `side`. `Boss` re-guesses it. Pass it in: `var horizontal = side is RailSide.Top or RailSide.Bottom;`

Not reachable today (84px button vs 9px rail), so maintainability only.

src/Aetherphone/Windows/Components/SideButton.cs
```diff
@@ -68,6 +68,7 @@ public SideButtonAction Update(Rect bounds, PhoneTheme theme, float delta)
     private static void DrawButton(Rect bounds, PhoneTheme theme, bool hovered, float progress, bool pressing)
     {
         var press = pressing ? 0.35f + 0.65f * progress : 0f;
-        HardwareButton.Draw(ImGui.GetWindowDrawList(), bounds, theme, RailSide.Right, hovered, press, 0f);
+        var side = bounds.Width > bounds.Height ? RailSide.Top : RailSide.Right;
```
XeldarAlz: `DeviceChrome.SideButtonRect` knew the side when it built this rect. Take `RailSide` as a parameter instead of inferring it.

Also lines 25-26: the hit padding stayed on portrait axes, so the 8px inward grab slop now lands outside the window and gets clipped.

src/Aetherphone/Windows/Components/SideToggle.cs
```diff
@@ -13,7 +13,8 @@ public static bool Update(Rect bounds, PhoneTheme theme, bool active, string hin
         var hitMax = new Vector2(bounds.Max.X + 4f * scale, bounds.Max.Y + 6f * scale);
         var hovered = UiInteract.Hover(hitMin, hitMax);
         var press = hovered && ImGui.IsMouseDown(ImGuiMouseButton.Left) ? 1f : 0f;
-        HardwareButton.Draw(ImGui.GetWindowDrawList(), bounds, theme, RailSide.Left, hovered, press, active ? 1f : 0f);
+        var side = bounds.Width > bounds.Height ? RailSide.Bottom : RailSide.Left;
```
XeldarAlz: Same: take `RailSide` from `PhoneShell` rather than inferring it.

src/Aetherphone/Core/Theme/ChassisGeometry.cs
RayaSerahill: Alright, I centralized the calls to Rect.IsLandscape 33d8c13

src/Aetherphone/Windows/Components/DeviceChrome.cs
RayaSerahill: Hoisted the fractions to consts and using them instead 6bd03b1

src/Aetherphone/Windows/Components/HardwareButton.cs
RayaSerahill: Adressed in bf8c619: Boss now uses the existing RailSide rather than inferring the orientation

src/Aetherphone/Windows/Components/SideButton.cs
RayaSerahill: Addressed in 88e1e62: `SideButton` now receives the authoritative `RailSide`, and the hit padding follows the correct landscape axes.

src/Aetherphone/Windows/Components/SideToggle.cs
RayaSerahill: Addressed in 88e1e62: `SideToggle` now receives the authoritative `RailSide` instead of inferring it

RayaSerahill:
All review comments have been addressed and the threads marked as resolved. I pushed the fixes in the latest commits

XeldarAlz (approved):
Lgtm

PR #93, opened 2026-08-06
fix(media): cull off-screen thumbnails in photo pickers

BluntEXE:
## Summary

Four photo picker grids decoded every image in their source list on every frame, with no check for what was actually scrolled into view: the compose "+" picker shared by Aethergram and Velvet, the wallpaper picker, the avatar and music station crop picker, and the classified ad photo picker.

Users with large photo libraries hit crashes and freezes opening these. One report in Discord: deleting photos fixed it for them. Decode count caused it, not image size. PR #69 already caps single-image dimensions, so one large photo can't do this on its own.

## Root cause

`PhotosApp.Grid.cs` already virtualizes its grid: it checks `scrollY`/`viewHeight`/`margin` before decoding a row. These four pickers used `ImGui.Dummy` plus `SameLine` auto-layout instead, looping the full path array and calling `Get()` on every entry regardless of visibility.

## Fix

I ported the scroll-culling pattern from `PhotosApp.Grid.cs` to all four sites: compute each cell's position directly, skip decode and click handling for rows outside `scrollY ± margin`, reserve the full scroll height with a trailing `Dummy`.

Three files (`PhotoComposeSession.cs`, `ImagePickCrop.cs`, `YellowPagesApp.Compose.cs`) used `ImRaii.PushStyle(ItemSpacing)` to space the old auto-layout grid. The new manual layout computes spacing itself, so that call and its `using Dalamud.Interface.Utility.Raii;` import are gone from those three. `WallpaperPage.cs` still uses `ImRaii` elsewhere, so its import stays.

## Testing

Local dev build, 2880 dummy photos (1280x720) dropped into the Photos library, opened the compose picker before and after.

- Before: memory spiked roughly 4GB above idle, UI froze.
- After: negligible RAM impact, no freeze.

This dataset didn't produce a full crash, only a freeze, but the mechanism matches what caused the reported crashes on larger or higher resolution libraries. I confirmed the other three pickers build and follow the identical code path; I didn't load-test each one individually.

### DIFF
--- a/src/Aetherphone/Apps/Settings/Pages/WallpaperPage.cs
+++ b/src/Aetherphone/Apps/Settings/Pages/WallpaperPage.cs
@@ -365,29 +365,38 @@ internal sealed class WallpaperPage : ISettingsPage
 
             DragScrollHost.Begin(photosKey);
 
-            var cell = (ScrollLayout.StableContentWidth() - gap * (PhotoColumns - 1)) / PhotoColumns;
+            var avail = ScrollLayout.StableContentWidth();
+            var cell = (avail - gap * (PhotoColumns - 1)) / PhotoColumns;
+            var origin = ImGui.GetCursorScreenPos();
+            var scrollY = ImGui.GetScrollY();
+            var viewHeight = ImGui.GetWindowSize().Y;
+            var margin = cell + 60f * scale;
             for (var index = 0; index < photoPaths.Length; index++)
             {
-                using (ImRaii.PushId(index))
+                var column = index % PhotoColumns;
+                var rowIndex = index / PhotoColumns;
+                var top = rowIndex * (cell + gap);
+                if (top + cell < scrollY - margin || top > scrollY + viewHeight + margin)
                 {
-                    ImGui.Dummy(new Vector2(cell, cell));
-                    var min = ImGui.GetItemRectMin();
-                    var max = ImGui.GetItemRectMax();
-                    DrawPhotoThumb(photoPaths[index], min, max, theme);
-                    if (UiInteract.Click(min, max, UiInteract.Hover(min, max)))
-                    {
-                        overlay = Overlay.None;
-                        navigator.Open(new WallpaperCropPage(photoPaths[index], navigator, assign, wallpapers,
-                            wallpaperImages));
-                        return;
-                    }
+                    continue;
                 }
 
-                if (index % PhotoColumns != PhotoColumns - 1)
+                var min = new Vector2(origin.X + column * (cell + gap), origin.Y + top);
+                var max = new Vector2(min.X + cell, min.Y + cell);
+                DrawPhotoThumb(photoPaths[index], min, max, theme);
+                if (UiInteract.Click(min, max, UiInteract.Hover(min, max)))
                 {
-                    ImGui.SameLine();
+                    overlay = Overlay.None;
+                    navigator.Open(new WallpaperCropPage(photoPaths[index], navigator, assign, wallpapers,
+                        wallpaperImages));
+                    return;
                 }
             }
+
+            var rows = (photoPaths.Length + PhotoColumns - 1) / PhotoColumns;
+            var totalHeight = rows * (cell + gap);
+            ImGui.SetCursorScreenPos(origin);
+            ImGui.Dummy(new Vector2(avail, totalHeight));
         }
     }
 
--- a/src/Aetherphone/Apps/Settings/Pages/WallpaperPage.cs
+++ b/src/Aetherphone/Apps/Settings/Pages/WallpaperPage.cs
@@ -354,7 +354,6 @@ internal sealed class WallpaperPage : ISettingsPage
         var photosKey = ImGui.GetID("##wallpaperPhotos");
         ImGui.SetCursorScreenPos(grid.Min);
         var gap = 6f * scale;
-        using (ImRaii.PushStyle(ImGuiStyleVar.ItemSpacing, new Vector2(gap, gap)))
         using (var child = ImRaii.Child("##wallpaperPhotos", grid.Size, false,
                    DragScrollHost.ScrollFlags(ImGuiWindowFlags.NoBackground)))
         {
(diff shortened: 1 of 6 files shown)


src/Aetherphone/Apps/Settings/Pages/WallpaperPage.cs
```diff
@@ -365,29 +365,38 @@ private void DrawPhotoPicker(Rect body, PhoneTheme theme)
...
-                            wallpaperImages));
-                        return;
-                    }
+                    continue;
                 }
 
-                if (index % PhotoColumns != PhotoColumns - 1)
+                var min = new Vector2(origin.X + column * (cell + gap), origin.Y + top);
+                var max = new Vector2(min.X + cell, min.Y + cell);
+                DrawPhotoThumb(photoPaths[index], min, max, theme);
```
XeldarAlz: `DrawPhotoThumb` reads `ImGui.IsItemHovered()` (line 407), which was the per-cell `Dummy`. With no item it falls back to the picker child, so every visible thumbnail tints at once. Pass `UiInteract.Hover(min, max)` in and reuse it for the `Click` below.

src/Aetherphone/Apps/YellowPages/YellowPagesApp.Compose.cs
```diff
@@ -349,26 +348,35 @@ private void DrawPhotoPicker(Rect area)
...
-                    if (index % PickerColumns != PickerColumns - 1)
-                    {
-                        ImGui.SameLine();
-                    }
+                    continue;
+                }
+
+                var min = new Vector2(origin.X + column * (cell + gap), origin.Y + rowTop);
+                var max = new Vector2(min.X + cell, min.Y + cell);
+                DrawPickerThumbnail(pickerPaths[index], min, max, scale);
```
XeldarAlz: Same hover break: `DrawPickerThumbnail` reads `ImGui.IsItemHovered()` (line 389), no longer bound to the cell. Pass `UiInteract.Hover(min, max)` in and reuse it for the `Click` below.

src/Aetherphone/Windows/Components/ImagePickCrop.cs
```diff
@@ -119,26 +118,35 @@ private ImagePickCropEvent DrawPick(Rect area, in PhoneContext context, in Image
...
+                    continue;
+                }
 
-                    if (index % GridColumns != GridColumns - 1)
-                    {
-                        ImGui.SameLine();
-                    }
+                var min = new Vector2(origin.X + column * (cell + gap), origin.Y + rowTop);
+                var max = new Vector2(min.X + cell, min.Y + cell);
+                DrawThumbnail(pickerPaths[index], min, max, theme, scale);
```
XeldarAlz: Same hover break: `DrawThumbnail` reads `ImGui.IsItemHovered()` (line 161). Pass `UiInteract.Hover(min, max)` in and reuse it for the `Click` below.

src/Aetherphone/Windows/Components/PhotoComposeSession.cs
```diff
@@ -211,31 +210,41 @@ private void LoadCrop(int index)
...
+            }
 
-                if (UiInteract.Click(min, max, UiInteract.Hover(min, max)))
-                {
-                    TakePicked(pickerPaths[index]);
-                }
+            var min = new Vector2(origin.X + column * (cell + gap), origin.Y + top);
+            var max = new Vector2(min.X + cell, min.Y + cell);
+            var path = pickerPaths[index];
+            DrawLocalThumbnail(path, min, max, scale, style.PlaceholderFill);
```
XeldarAlz: Same hover break: `DrawLocalThumbnail` reads `ImGui.IsItemHovered()` (line 278). `DrawCaptionStrip` (line 384) submits no item either, so a hover parameter fixes both callers.

XeldarAlz (changes requested):
Right diagnosis, right pattern, and the branch builds clean locally.

One regression to fix: the per-cell `ImGui.Dummy` was the item every thumbnail helper's `ImGui.IsItemHovered()` was reading. Four inline comments.

The sweep also misses two pickers with the identical unbounded loop: `FeedbackApp.cs:320` (Feedback, add photos) and `ChatThreadView.cs:851` (Messages picker). Same crash reachable from those screens.

Nit: `WallpaperPage.cs:357` still pushes `ItemSpacing` around the child, dead with the manual layout.

Follow-up, not this PR: the cull math is in six copies now, worth one shared struct.

src/Aetherphone/Apps/Settings/Pages/WallpaperPage.cs
BluntEXE: Fixed: WallpaperPage.cs now computes hovered = UiInteract.Hover(min, max) in the loop, passes it into DrawPhotoThumb, and reuses it for the Click check below. Pushed in 05b0b7af.

src/Aetherphone/Apps/YellowPages/YellowPagesApp.Compose.cs
BluntEXE: Fixed: computes hovered = UiInteract.Hover(min, max) in the loop, passes it into DrawPickerThumbnail, reuses it for Click. Pushed in 05b0b7af.

src/Aetherphone/Windows/Components/ImagePickCrop.cs
BluntEXE: Fixed: computes hovered = UiInteract.Hover(min, max) in the loop, passes it into DrawThumbnail, reuses it for Click. Pushed in 05b0b7af.

src/Aetherphone/Windows/Components/PhotoComposeSession.cs
BluntEXE: Fixed both callers: DrawPickGrid now computes hovered from the cell rect and passes it into DrawLocalThumbnail, reusing it for Click. DrawCaptionStrip's call site now passes UiInteract.Hover(min, max) directly since it has no click handling to reuse it for. Pushed in 05b0b7af.

XeldarAlz (approved):
Lgtm

PR #94, opened 2026-08-06
fix(platform): parent the native file dialog to the game window

Kiro-XVI:
## Summary
- `NativeFileDialog.ShowDialog` created the Win32 `GetOpenFileNameW` common dialog with `hwndOwner = IntPtr.Zero`, an unowned dialog
- On a fullscreen exclusive DirectX game this can force a display mode or focus transition, destabilizing the device for any texture work happening on Dalamud's own queue at the same time
- Reproduced as a hard game crash (native access violation, confirmed via crash dump) when opening the photo/audio picker from Aethergram's post composer and from Velvet's post composer, which both route through the same `PhotoComposeSession.LaunchImportDialog` -> `FilePicker.PickImage` -> `NativeFileDialog` path
- Fix: parent the dialog to the game's own window handle instead

## Test plan
- [x] `dotnet build src/Aetherphone/Aetherphone.csproj -c Release` succeeds
- [x] Verified in-game via dev-plugin reload: opening the photo picker from Aethergram's "+" and from Velvet's post composer no longer crashes the game across repeated attempts


src/Aetherphone/Core/Platform/NativeFileDialog.cs
```diff
@@ -65,7 +66,10 @@ private static void Complete(Task<string?> dialog, Action<string> onPicked)
             var dialog = new OpenFileName
             {
                 lStructSize = Marshal.SizeOf<OpenFileName>(),
-                hwndOwner = IntPtr.Zero,
+                // An unowned dialog on a fullscreen exclusive game can force a display mode or
+                // focus transition, destabilizing the DirectX device for any texture work
+                // happening on Dalamud's own queue around the same time.
+                hwndOwner = Process.GetCurrentProcess().MainWindowHandle,
```
XeldarAlz: Blocking: an owned common dialog is modal to its owner, so comdlg32 disables the game window for as long as the picker is open. Today you can alt-tab back and keep playing while browsing, after this you cannot, and if the STA dialog thread dies mid-call the window stays disabled. Is that trade intended?

src/Aetherphone/Core/Platform/NativeFileDialog.cs
```diff
@@ -1,3 +1,4 @@
+using System.Diagnostics;
```
XeldarAlz, line 1: Non-blocking: `Process.GetCurrentProcess()` is not disposed here, unlike the `using var process` in `DetectShaderOverlay` (NativeFileDialog.cs:127 on master). `MainWindowHandle` also walks every top-level window per open. Resolve it once into a `static readonly IntPtr` next to `RunsUnderWine`.

src/Aetherphone/Core/Platform/NativeFileDialog.cs
```diff
@@ -65,7 +66,10 @@ private static void Complete(Task<string?> dialog, Action<string> onPicked)
             var dialog = new OpenFileName
             {
                 lStructSize = Marshal.SizeOf<OpenFileName>(),
-                hwndOwner = IntPtr.Zero,
+                // An unowned dialog on a fullscreen exclusive game can force a display mode or
```
XeldarAlz: Nit: repo style keeps comments to a why the code cannot express. The DirectX queue mechanism here is a theory and already lives in the commit message. One line about the owner keeping the dialog above a fullscreen game would carry the same weight.

XeldarAlz (changes requested):
Merged onto master locally: `dotnet build Aetherphone.sln -c Release` clean, 512/512 tests pass. Not tested in-game. The branch predates f5a73660, which added the Wine/ReShade gate on this file, so please rebase and confirm the repro still stands on current master. One blocking note inline.

src/Aetherphone/Core/Platform/NativeFileDialog.cs
XeldarAlz: Blocking: `Process.GetCurrentProcess()` is IDisposable, so this leaks a native process handle on every picker open. `DetectShaderOverlay` in this same file already does `using var process = ...`. Use the game's own handle instead: `Device.Instance()->hWnd` (FFXIVClientStructs Kernel.Device) is one pointer read, no allocation.

XeldarAlz: Non-blocking: `MainWindowHandle` runs EnumWindows and takes the first visible top-level window the process owns. Dalamud and other plugins own top-level windows here, so it is not guaranteed to resolve to the game window. `Device.Instance()->hWnd` is exact.

XeldarAlz: Non-blocking: the dialog runs on the STA thread built at NativeFileDialog.cs:55, so this owner belongs to a different thread. Windows attaches both input queues for an owned modal, so a slow shell enumeration (network path, dead drive) can now stall the game window too. Fair trade for keeping the picker on top, but worth knowing.

src/Aetherphone/Core/Platform/NativeFileDialog.cs
XeldarAlz: Blocking: this branch is from Jul 24, three days before `HasShaderOverlay` landed in f5a73660, so it carries no overlay gate and every machine reaches comdlg32. The AV diagnosed on this path was ReShade's dxgi hook faulting inside the shell thumbcache worker, which parenting does not prevent. Which module faulted in your dump? Please retest on a rebased branch.

src/Aetherphone/Core/Platform/NativeFileDialog.cs
```diff
@@ -65,7 +66,10 @@ private static void Complete(Task<string?> dialog, Action<string> onPicked)
...
+                // focus transition, destabilizing the DirectX device for any texture work
```
XeldarAlz: Nit: house rule is no comments except a why the code cannot express. Three lines of mechanism exceeds that, and the real reason is simpler: an unowned picker can open behind the game window and read as a freeze. One line is enough.

XeldarAlz (commented):
Merged into current master, `dotnet build Aetherphone.sln -c Release` clean and 517/517 tests pass. Not verified in game. Parenting the picker is worth having, but the handle source and the stated cause both need work. Findings inline; I am pushing the mechanical fixes to this branch.

XeldarAlz:
Pushed the fixes onto this branch, plus a merge of current master so it is testable against the shipped code.

- owner now comes from `Device.Instance()->hWnd` instead of `Process.GetCurrentProcess().MainWindowHandle`: exact, no undisposed handle, no window enumeration
- handle resolved on the calling thread rather than inside the dialog's STA thread
- mechanism comment cut to one line

Build clean, 517/517 tests pass. Still open on your side: the branch predated `HasShaderOverlay`, so please confirm what faulted in the dump and retest the picker in game now that the overlay gate is in. If the dump points at ReShade's dxgi.dll, the real fix is widening that detection, not this.

XeldarAlz:
Modal trade accepted: being modal to the game window is correct behavior for a file picker, and an unowned picker hiding behind a fullscreen game is the complaint this fixes. That resolves the standing blocking question from the Aug 6 review.

Handle source, leak and thread affinity are fixed on the branch. Build clean, 517/517 tests pass, both checks green.

Squashing with the crash diagnosis rewritten out of the commit message: this branch predated the Wine/ReShade gate in f5a73660, so the access violation is very likely the already-gated shell thumbcache path rather than a fullscreen display-mode transition. Parenting is worth having on its own merits, so the fix lands, just described for what it actually does.

PR #95, opened 2026-08-06
fix(core): harden session, settings, and realtime lifecycle races

Kiro-XVI:
## What

Fixes several session, settings, and telephony lifecycle bugs on master: Follow Character toggle fighting itself, decrypted DM plaintext surviving sign-out, KeyVault not refreshing on account switch, Confirm/incoming-call UI mutated off the Framework thread, RealtimeConnection Stop/Start races, privacy settings refetch storms, RegionSync no-op after SignIn, and related settings/clock/call edge cases.

## Why

These bugs are present on upstream master and cause broken account follow, leftover chat plaintext after sign-out, dropped/corrupt confirm dialogs, overlapping websocket loops on account switch, and settings pages that hammer MeAsync every frame after a failed load.

Closes #

## How to test

- Open Settings → Account with more than one account: turn on Follow Character and confirm it stays on across frames.
- Sign out while a DM thread with decrypted messages is open: confirm messages/key state clear.
- Switch between signed-in accounts: confirm KeyVault refreshes for the new identity.
- Trigger an incoming call / stream decline alert and confirm the confirm overlay still works.
- Open Settings privacy / Tags & Mentions / Aethergram message privacy with the network offline briefly: confirm it does not refetch every frame forever.
- Toggle 12-hour clock and confirm Clock alarms follow `TimeText`.
- Open Appearance with HomeGridRows previously set to 8: confirm it does not silently rewrite to 6 until you change density.

## Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/Aetherphone/Apps/Settings/Pages/PrivacyPage.cs
+++ b/src/Aetherphone/Apps/Settings/Pages/PrivacyPage.cs
@@ -226,12 +226,15 @@ internal sealed class PrivacyPage : ISettingsPage, IDisposable
                 {
                     shareReadReceipts = me.ShareReadReceipts;
                     sharePresence = me.SharePresence;
-                    chatPrivacyLoaded = true;
                 }
+
+                // Latch loaded even on null/failure so a dead endpoint cannot refetch every frame.
+                chatPrivacyLoaded = true;
             }
             catch (Exception exception)
             {
                 AepLog.Warning($"Chat privacy load failed: {exception.Message}");
+                chatPrivacyLoaded = true;
             }
             finally
             {
--- a/src/Aetherphone/Apps/Settings/Pages/TagsMentionsPage.cs
+++ b/src/Aetherphone/Apps/Settings/Pages/TagsMentionsPage.cs
@@ -131,12 +131,14 @@ internal sealed class TagsMentionsPage : ISettingsPage, IDisposable
                     mentionPolicy = me.MentionPolicy;
                     tagPolicy = me.TagPolicy;
                     requireTagApproval = me.RequireTagApproval;
-                    loaded = true;
                 }
+
+                loaded = true;
             }
             catch (Exception exception)
             {
                 AepLog.Warning($"Tag privacy load failed: {exception.Message}");
+                loaded = true;
             }
             finally
             {
--- a/src/Aetherphone/Core/Aethernet/SignInFlow.cs
+++ b/src/Aetherphone/Core/Aethernet/SignInFlow.cs
@@ -101,7 +101,7 @@ internal sealed class SignInFlow : IDisposable
                 if (result.Auth is { } auth)
                 {
                     session.SignIn(auth.Token, auth.User);
-                    signedIn?.Invoke();
+                    _ = Plugin.Framework.RunOnFrameworkThread(() => signedIn?.Invoke());
                     Reset();
                     return;
                 }
@@ -188,7 +188,7 @@ internal sealed class SignInFlow : IDisposable
             if (result.Auth is { } auth)
             {
                 session.SignIn(auth.Token, auth.User);
-                signedIn?.Invoke();
+                _ = Plugin.Framework.RunOnFrameworkThread(() => signedIn?.Invoke());
                 Reset();
                 return;
             }
--- a/src/Aetherphone/Core/Aethernet/StoreWork.cs
+++ b/src/Aetherphone/Core/Aethernet/StoreWork.cs
@@ -67,6 +67,5 @@ internal sealed class StoreWork : IDisposable
     public void Dispose()
     {
         cancellation.Cancel();
-        cancellation.Dispose();
     }
 }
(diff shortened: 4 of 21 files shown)


src/Aetherphone/Core/Telephony/RealtimeConnection.cs
```diff
@@ -65,20 +72,33 @@ public void Stop()
             AepLog.Warning($"Realtime abort failed: {exception.Message}");
         }
 
-        toAbort?.Dispose();
+        // The receive loop owns the socket via `using` - only Abort here so its Dispose runs once.
+        if (toWait is not null)
+        {
+            try
+            {
+                toWait.Wait(StopJoinTimeout);
```
XeldarAlz, line 80: Blocking: `AethernetSession` raises `Changed` inside `RunOnFrameworkThread` and `CallHub.Reconcile` subscribes to it, so `router.Stop()` runs on the Framework thread. Sign-in, sign-out, account switch and 401 now stall the game up to 2s here (`CallHub.SetEnabled` reaches it from the settings toggle too). The generation guard already covers the race the join was meant to close, so drop the `Wait` or move `Stop` off the main thread.

src/Aetherphone/Apps/Settings/Pages/PrivacyPage.cs
```diff
@@ -226,12 +226,15 @@ private void EnsureLoaded()
...
-                    chatPrivacyLoaded = true;
                 }
+
+                // Latch loaded even on null/failure so a dead endpoint cannot refetch every frame.
+                chatPrivacyLoaded = true;
             }
             catch (Exception exception)
             {
                 AepLog.Warning($"Chat privacy load failed: {exception.Message}");
+                chatPrivacyLoaded = true;
```
XeldarAlz, line 237: Blocking: latching on failure kills the refetch storm but makes the failure permanent for the session. The page then renders the `shareReadReceipts`/`sharePresence` defaults (both `true`) as authoritative, and `Push` sends both booleans, so flipping one silently overwrites the other server-side. Latch only when `me is not null`, and give the failure path a retry cooldown instead.

src/Aetherphone/Apps/Settings/Pages/TagsMentionsPage.cs
```diff
@@ -131,12 +131,14 @@ private void EnsureLoaded()
...
                     requireTagApproval = me.RequireTagApproval;
-                    loaded = true;
                 }
+
+                loaded = true;
             }
             catch (Exception exception)
             {
                 AepLog.Warning($"Tag privacy load failed: {exception.Message}");
+                loaded = true;
```
XeldarAlz, line 141: Blocking: same shape as PrivacyPage. After a failed load the page shows default policies as real values, and `PushTags` sends `tagPolicy` and `requireTagApproval` together, so touching one overwrites the other with a default. Same fix: cooldown rather than a permanent latch.

src/Aetherphone/Core/Telephony/CallAudioController.cs
```diff
@@ -44,13 +44,15 @@ public bool EnsureStartedLocked(Guid callId, int localSlot)
             return false;
         }
 
+        if (localSlot < 0)
```
XeldarAlz, line 47: Non-blocking: `CallHub.HandleRoster` still sets `state = CallState.Active` when this returns false, so a roster that omits the local slot leaves the call Active with no audio session and no elapsed timer, and nothing retries until the next roster frame.

src/Aetherphone/Core/Aethernet/SignInFlow.cs
```diff
@@ -101,7 +101,7 @@ public void VerifyLodestone()
                 if (result.Auth is { } auth)
                 {
                     session.SignIn(auth.Token, auth.User);
-                    signedIn?.Invoke();
+                    _ = Plugin.Framework.RunOnFrameworkThread(() => signedIn?.Invoke());
```
XeldarAlz, line 104: Nit: `Reset()` now runs before `signedIn` fires, so handlers observe the cleared flow state instead of the post-verify state. Also `Plugin.Framework` is unguarded here while `ConfirmService` gained a null guard in ab98f8f; worth keeping the two consistent.

src/Aetherphone/Core/Aethernet/StoreWork.cs
```diff
@@ -67,6 +67,5 @@ public void Run(
     public void Dispose()
     {
         cancellation.Cancel();
```
XeldarAlz, line 69: Nit: dropping `cancellation.Dispose()` is the right call since in-flight tasks still hold the token, but it reads as an accidental leak. A one-line why comment would stop someone re-adding it.

XeldarAlz (changes requested):
Build: `dotnet build -c Release` clean, 512/512 tests pass on ab98f8f. Not verified in-game. Two blockers inline (realtime `Stop` joins on the Framework thread, privacy pages latch `loaded` on failure); the rest are non-blocking.

PR #96, opened 2026-08-06
feat(home): let Shortcuts plugin launchers live in home folders

Kiro-XVI:
## What

Home folders can now contain Shortcuts tiles (including plugin launchers), not only apps. Drag a shortcut onto an app or into a folder to group them; the folder overlay draws, opens, and removes those shortcuts.

## Why

Pinning a plugin launcher to the Home screen worked, but folders rejected shortcuts, so users could not organize plugin icons the same way as apps.

Closes #

## How to test

1. Open Shortcuts, create a plugin launcher, enable Show on Home, save.
2. On the Home screen, enter edit mode and drag the shortcut onto another app icon: a folder should form.
3. Drag another shortcut into that folder.
4. Open the folder: tap a shortcut to run it; in edit mode, remove it with the badge.
5. Reload the plugin and confirm the folder still contains the shortcuts.

## Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/Aetherphone/Core/Home/HomeLayoutService.cs
+++ b/src/Aetherphone/Core/Home/HomeLayoutService.cs
@@ -1117,7 +1117,11 @@ internal sealed class HomeLayoutService
                     continue;
                 }
 
-                item.Members.Add(new HomeItem { Kind = "app", AppId = member.App!.Id });
+                var appId = member.App!.Id;
+                item.Members.Add(new HomeItem { Kind = "app", AppId = appId });
+
+                // Mirrored into AppIds so rolling back to a build that only reads AppIds keeps the app grouping.
+                item.AppIds.Add(appId);
             }
 
             return item;
--- a/src/Aetherphone/Core/Home/HomeLayoutService.cs
+++ b/src/Aetherphone/Core/Home/HomeLayoutService.cs
@@ -223,8 +223,8 @@ internal sealed class HomeLayoutService
 
     public void MakeFolder(HomeTile target, HomeTile dragged)
     {
-        if (ReferenceEquals(target, dragged) || target.IsWidget || dragged.IsWidget || target.IsShortcut ||
-            dragged.IsShortcut || target.IsFolder && dragged.IsFolder)
+        if (ReferenceEquals(target, dragged) || target.IsWidget || dragged.IsWidget
+            || target.IsFolder && dragged.IsFolder)
         {
             return;
         }
@@ -237,29 +237,29 @@ internal sealed class HomeLayoutService
 
         if (target.IsFolder)
         {
-            AddApps(target, dragged);
+            AddMember(target, dragged);
         }
         else
         {
             (targetPage, targetIndex) = Locate(target);
-            var folder = HomeTile.ForFolder(NextFolderKey(), string.Empty, new[] { target.App! });
+            var folder = HomeTile.ForFolder(NextFolderKey(), string.Empty, new[] { HomeTile.AsLeaf(target) });
             folder.Cell = target.Cell;
-            AddApps(folder, dragged);
+            AddMember(folder, dragged);
             pages[targetPage][targetIndex] = folder;
         }
 
         Commit();
     }
 
-    public void RemoveFromFolder(HomeTile folder, IPhoneApp app, int targetPage)
+    public void RemoveFromFolder(HomeTile folder, HomeTile member, int targetPage)
     {
-        if (!folder.IsFolder || !folder.Apps.Remove(app))
+        if (!folder.IsFolder || !folder.Members.Remove(member))
         {
             return;
         }
 
         targetPage = Math.Clamp(targetPage, 0, Math.Max(0, pages.Count - 1));
-        pages[targetPage].Add(HomeTile.ForApp(app));
+        pages[targetPage].Add(HomeTile.AsLeaf(member));
         Commit();
     }
 
@@ -356,11 +356,11 @@ internal sealed class HomeLayoutService
                     return;
                 }
 
-                for (var appIndex = tile.Apps.Count - 1; appIndex >= 0; appIndex--)
+                for (var memberIndex = tile.Members.Count - 1; memberIndex >= 0; memberIndex--)
                 {
-                    if (string.Equals(tile.Apps[appIndex].Id, appId, StringComparison.Ordinal))
+                    if (string.Equals(tile.Members[memberIndex].App?.Id, appId, StringComparison.Ordinal))
                     {
-                        tile.Apps.RemoveAt(appIndex);
+                        tile.Members.RemoveAt(memberIndex);
                         return;
                     }
                 }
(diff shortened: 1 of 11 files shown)


src/Aetherphone/Core/Home/HomeTile.cs
```diff
@@ -13,7 +13,7 @@ internal sealed class HomeTile : IGridTile
     public WidgetSize Size { get; set; } = WidgetSize.Medium;
     public string FolderName { get; set; } = string.Empty;
     public string FolderTint { get; set; } = string.Empty;
-    public List<IPhoneApp> Apps { get; } = new();
+    public List<HomeTile> Members { get; } = new();
```
XeldarAlz, line 16: Blocking: dropping `Apps` breaks the test project. `src/Aetherphone.Tests/HomeLayoutServiceInstallTests.cs:293` and `:295` still read `tile.Apps`, so the solution build fails with CS1061 and CI is red. Update `PageIndexOf` to walk `tile.Members[index].App?.Id`.

src/Aetherphone/Core/Shell/Home/HomeGridRenderer.cs
```diff
@@ -149,7 +149,7 @@ private void DrawTile(in HomeMetrics metrics, PhoneTheme theme, HomeTile tile, R
         {
             HomeTileView.DrawFolder(center, rect.Width, tile, theme,
                 interaction.TapScale(tile) * interaction.Magnify(center, metrics.CellWidth),
-                labelAlpha, showLabels, Loc.T(L.Home.NewFolder), metrics.CellWidth, zoom);
+                labelAlpha, showLabels, Loc.T(L.Home.NewFolder), metrics.CellWidth, zoom, shortcuts.Icon);
```
XeldarAlz: Blocking: `shortcuts.Icon` is an instance method group, so this allocates a fresh delegate every frame for every folder tile (measured 64 B per conversion; the compiler only caches static method groups). Draw code must not allocate. Pass the `ShortcutStore` into `DrawFolder`, or cache one `Func` field here. Same at line 312.

src/Aetherphone/Core/Home/HomeLayoutService.cs
```diff
@@ -1020,9 +1105,19 @@ private static HomeItem BuildStoredItem(HomeTile tile)
...
+                if (member.IsShortcut)
+                {
+                    item.Members.Add(new HomeItem
+                    {
+                        Kind = "shortcut", ShortcutId = member.Shortcut!.Id.ToString("N"),
+                    });
+                    continue;
+                }
+
+                item.Members.Add(new HomeItem { Kind = "app", AppId = member.App!.Id });
```
XeldarAlz: Non-blocking: folders stop serializing `AppIds` entirely, so a user who rolls back to an earlier plugin build loads every folder as empty and loses the grouping (the apps return loose via `LoadInstalled`). Cheap insurance: keep filling `item.AppIds` for app members for one release, since load already prefers `Members`.

src/Aetherphone/Core/Home/HomeLayoutService.cs
```diff
@@ -789,6 +819,52 @@ private void AppendSeedApps(List<HomeTile> page, string[] ids, HashSet<string> p
...
+                if (ResolveFolderMember(item.Members[index], placed) is { } member)
+                {
+                    contents.Add(member);
+                }
+            }
+
+            return contents;
+        }
+
+        var apps = ResolveApps(item.AppIds, placed);
```
XeldarAlz, line 838: Non-blocking: neither the new `Members` round trip nor this legacy `AppIds` fallback has a test, and the shortcut path is exactly what regresses silently. `HomeLayoutServiceInstallTests` already has the fakes to cover a save then load with a shortcut inside a folder.

XeldarAlz (changes requested):
Ran `dotnet build Aetherphone.sln -c Release` locally: FAILED with 2 errors in Aetherphone.Tests, same reason CI Build (Windows) is red, so no tests ran. Not verified in-game. The folder member model itself is the right shape; two blocking items and two follow-ups are inline.

XeldarAlz:
Pushed `ee7cb81b` to this branch so it stops sitting on a red build. Four things, all from the review above:

**Blocking, both fixed**

- `HomeLayoutServiceInstallTests.PageIndexOf` still read `tile.Apps`, which is what turned the solution build red (CS1061 x2). It now walks `tile.Members[memberIndex].App?.Id`.
- `HomeTileView.DrawFolder` took the icon lookup as `Func<...>? shortcutIcon = null`, and the two call sites passed the `shortcuts.Icon` method group, so every folder tile converted a fresh delegate every frame. The parameter is required now (moved ahead of `zoom`) and `HomeGridRenderer` caches one `shortcutIcon` field built in its constructor, so the conversion happens once at startup instead of per frame.

**Follow-ups, both taken**

- `BuildStoredItem` mirrors app members back into `item.AppIds` alongside `Members`. Load already prefers `Members`, so this changes nothing going forward, but a user who rolls back to an earlier build keeps their app groupings instead of loading every folder empty. Shortcut members cannot be expressed in the old format, so a rollback still drops those.
- New `HomeLayoutServiceFolderTests` covers the parts that regress silently: a shortcut dragged onto an app surviving a save and reload, the legacy `AppIds` fallback still grouping apps, a shortcut inside a folder still reading as pinned (and `AddShortcut` refusing to duplicate it), and deleting a shortcut that lives in a folder pulling it out cleanly.

`dotnet build Aetherphone.sln -c Release` is green and all 517 tests pass locally. The member model you wrote is unchanged, these are just the sharp edges around it. Still not verified in-game, so I will do that pass before merging.

src/Aetherphone/Core/Home/HomeTile.cs
XeldarAlz: Fixed in ee7cb81b: PageIndexOf walks tile.Members[memberIndex].App?.Id. The solution build is green and CI Build (Windows) passes.

src/Aetherphone/Core/Shell/Home/HomeGridRenderer.cs
XeldarAlz: Fixed in ee7cb81b: the icon lookup is a required parameter on DrawFolder now (moved ahead of zoom so it can be non-optional), and HomeGridRenderer caches one shortcutIcon field built in its constructor. Both call sites pass the field, so the method group conversion happens once at startup instead of per frame per folder.

src/Aetherphone/Core/Home/HomeLayoutService.cs
XeldarAlz: Done in ee7cb81b: BuildStoredItem mirrors app members into item.AppIds alongside Members, with a why comment so it does not read as dead duplication. Covered by HomeLayoutServiceFolderTests.ASavedFolder_StillWritesAppIdsForBuildsThatOnlyReadThem. Shortcut members still cannot be expressed in the old format, so a rollback drops those.

src/Aetherphone/Core/Home/HomeLayoutService.cs
XeldarAlz: Done in ee7cb81b: new HomeLayoutServiceFolderTests covers the shortcut save and load round trip, this legacy AppIds fallback, a shortcut inside a folder reading as pinned, and deleting a shortcut that lives in a folder. 517 tests pass.

XeldarAlz (approved):
All four review items are addressed in ee7cb81b, threads resolved. Build and Guards are green and the 517 unit tests pass. The Members model is a clean generalisation of the old app-only folder, and folder persistence is now covered by tests. Merging with the in-game pass on the folder drag, draw, and run paths still outstanding.

PR #99, opened 2026-08-07
(Chats) update thread logic to ensure scroll is based o nthe l…

Deldee:
…oaded data, instead of the previous conversaion/unloaded contentSize

## What
Updated ChatTranscript and ChatThreadView to ensure that ScrollY is set only once the data is laoded

## Why

Previously, opening longer chat caused the scroll to start roughly midway, based on the data loaded o the first frame, instead of the data once fully loaded

Closes #

## Checklist

- [ ] Open a chat in game, ensure the scrolling is always at the bottom, especially on longer chats

### DIFF
--- a/src/Aetherphone/Windows/Components/ChatThreadView.cs
+++ b/src/Aetherphone/Windows/Components/ChatThreadView.cs
@@ -245,6 +245,7 @@ internal abstract class ChatThreadView<TMessage, TThread> : IDisposable, IChatTr
             composer.CancelVoice();
             voicePlayer.Stop();
             OnThreadOpened(threadId);
+			transcript.RequestSnapToBottom()
         }
 
         if (pendingPrefill is { } prefill)
--- a/src/Aetherphone/Windows/Components/ChatTranscript.cs
+++ b/src/Aetherphone/Windows/Components/ChatTranscript.cs
@@ -278,7 +278,7 @@ internal sealed class ChatTranscript
                 return;
             }
 
-            SyncFollow(model.ThreadId);
+            SyncFollow(model.ThreadId, model.Loading);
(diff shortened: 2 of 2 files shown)


XeldarAlz:
Thanks for this, and for the write up. The diagnosis was right: the transcript was deciding its scroll position before the real content height was known, and a long chat opened partway up because of it.

The fix has landed on master as cf59fcf3 (merged in c9d852d5), so I am closing this one out.

I ended up taking a narrower version of your change. Two notes on why, in case they are useful:

The `RequestSnapToBottom()` on thread open is the part that does the work, and that is what shipped. It also turns out to be a restoration: e82fa279 originally added a snap on open, and it was dropped when the thread UI moved into the shared ChatThreadView in 04f1aad9. The failing case is reopening a thread the transcript has already tracked, where `followBottom` was being derived from a scroll offset inherited from whatever screen drew through the shared AppSurface child. A low inherited offset also tripped `MaybeLoadOlder`, which then pinned the scroll for up to 20 seconds while the older page loaded.

The `SyncFollow(threadId, loading)` gate cannot fire on the load path. `OpenThread` clears `messages` before setting `loadingThread = true`, and `StoreWork.Run` clears the flag in the same `finally` that follows the assignment, so while `Loading` is true the transcript has zero messages and takes the early return in `ChatTranscript.Draw` before `SyncFollow` is reached.

One thing worth flagging for next time: `transcript.RequestSnapToBottom()` was missing its semicolon and used tab indentation, so the branch did not build. CI does not run automatically on fork PRs here, so a local `dotnet build -c Release` before pushing is the fastest way to catch that.

Appreciated, this was a real bug and a good catch.

PR #100, opened 2026-08-07
fix(social): fixed stale comment and reaction counts particularly on rechirp wrappers

Farroness:
## What

Fixed the feed action header bar to use post.CommentCount; instead of an if statement to ensure that we're using the comment count coming from the server and not the local array. Also added logic to handle updating comments and reacts for posts within rechirp wrappers.

## Why

Discord Issue: https://discord.com/channels/1527213298087493732/1531772997084385381

## How to test

1. Open Chirper.
2. Review posts to ensure the comment badge and replies are showing the same values. Check rechirps and original posts to make sure comments and react counters are staying in sync.

## Checklist

### DIFF
--- a/src/Aetherphone/Apps/Chirper/ChirperApp.cs
+++ b/src/Aetherphone/Apps/Chirper/ChirperApp.cs
@@ -1080,7 +1080,7 @@ internal sealed partial class ChirperApp : IPhoneApp
             }
 
             var comments = store.DetailComments;
-            var commentTotal = store.HasMoreComments ? Math.Max(post.CommentCount, comments.Length) : comments.Length;
+            var commentTotal = post.CommentCount;
             ImGui.Dummy(new Vector2(0f, 2f * scale));
             ui.SectionHeading(commentTotal > 0
                 ? $"{Loc.T(L.Chirper.RepliesTitle)} · {commentTotal}"
--- a/src/Aetherphone/Core/Aethernet/CopyOnWrite.cs
+++ b/src/Aetherphone/Core/Aethernet/CopyOnWrite.cs
@@ -199,6 +199,34 @@ internal static class CopyOnWrite
         return source;
     }
 
+    // Replaces a post that appears directly in the array, AND updates the ReferencedPost
+    // inside any rechirp wrappers (posts whose RepostOfId matches the updated post's Id).
+    public static PostDto[] ReplaceOrUpdateReferenced(PostDto[] source, PostDto updated)
+    {
+        var changed = false;
+        var result = new PostDto[source.Length];
+        for (var index = 0; index < source.Length; index++)
+        {
+            var item = source[index];
+            if (item.Id == updated.Id)
+            {
+                result[index] = updated;
+                changed = true;
+            }
+            else if (item.RepostOfId == updated.Id && item.ReferencedPost is not null)
+            {
+                result[index] = item with { ReferencedPost = updated };
+                changed = true;
+            }
+            else
+            {
+                result[index] = item;
+            }
+        }
+
+        return changed ? result : source;
+    }
+
     public static T[] Map<T>(T[] source, Func<T, bool> match, Func<T, T> transform)
     {
         var changed = false;
--- a/src/Aetherphone/Core/Social/SocialFeedStore.cs
+++ b/src/Aetherphone/Core/Social/SocialFeedStore.cs
@@ -100,6 +100,10 @@ internal abstract class SocialFeedStore : IDisposable
         if (string.Equals(removal.Kind, ContentRemovalKinds.Comment, StringComparison.Ordinal))
         {
             detailComments = CopyOnWrite.RemoveById(detailComments, removal.ContentId);
+            if (removal.ParentId is { } postId)
+            {
+                BumpCommentCount(postId, -1);
+            }
         }
     }
 
@@ -1023,9 +1027,9 @@ internal abstract class SocialFeedStore : IDisposable
 
     protected void ReplacePost(PostDto updated)
     {
-        forYouLane.Items = CopyOnWrite.Replace(forYouLane.Items, updated);
-        followingLane.Items = CopyOnWrite.Replace(followingLane.Items, updated);
-        profileLane.Items = CopyOnWrite.Replace(profileLane.Items, updated);
+        forYouLane.Items = CopyOnWrite.ReplaceOrUpdateReferenced(forYouLane.Items, updated);
+        followingLane.Items = CopyOnWrite.ReplaceOrUpdateReferenced(followingLane.Items, updated);
+        profileLane.Items = CopyOnWrite.ReplaceOrUpdateReferenced(profileLane.Items, updated);
         if (detailPost is { } current && current.Id == updated.Id)
         {
             detailPost = updated;
@@ -1089,8 +1093,16 @@ internal abstract class SocialFeedStore : IDisposable
         wasFollowing == following ? 0 : following ? 1 : -1;
 
     private static PostDto[] MapCommentCount(PostDto[] source, string postId, int delta) =>
-        CopyOnWrite.MapById(source, postId,
-            post => post with { CommentCount = Math.Max(0, post.CommentCount + delta) });
+        CopyOnWrite.Map(source,
+            post => post.Id == postId || post.RepostOfId == postId,
+            post => post.Id == postId
+                ? post with { CommentCount = Math.Max(0, post.CommentCount + delta) }
+                : post with
+                {
+                    ReferencedPost = post.ReferencedPost is { } referenced
+                        ? referenced with { CommentCount = Math.Max(0, referenced.CommentCount + delta) }
+                        : null
+                });
 
     private static UserDto[] MapFollow(UserDto[] source, string userId, bool following, bool requested) =>
         CopyOnWrite.Map(source,


src/Aetherphone/Core/Social/SocialFeedStore.cs
```diff
@@ -100,6 +100,10 @@ private void OnContentRemoved(ContentRemovalSignal removal)
         if (string.Equals(removal.Kind, ContentRemovalKinds.Comment, StringComparison.Ordinal))
         {
             detailComments = CopyOnWrite.RemoveById(detailComments, removal.ContentId);
+            if (removal.ParentId is { } postId)
+            {
+                BumpCommentCount(postId, -1);
```
XeldarAlz, line 105: Blocking: this double-decrements on self-delete. `DeleteComment` already applies the optimistic -1 (:446, :466) and `content.removed` reaches the actor's own socket, so deleting your own comment drops the count by 2. The bump was omitted here deliberately. Needs an echo guard, e.g. a small ring of self-removed comment ids consumed here.

src/Aetherphone/Core/Social/SocialFeedStore.cs
```diff
@@ -100,6 +100,10 @@ private void OnContentRemoved(ContentRemovalSignal removal)
         if (string.Equals(removal.Kind, ContentRemovalKinds.Comment, StringComparison.Ordinal))
         {
             detailComments = CopyOnWrite.RemoveById(detailComments, removal.ContentId);
+            if (removal.ParentId is { } postId)
```
XeldarAlz, line 103: Non-blocking: `ModerationNoticeService.cs:207` publishes with `ParentId: null`, so the author-side fallback never reaches this branch. A removal seen over the WS fixes the count, the same removal seen through the moderation fallback does not.

src/Aetherphone/Core/Social/SocialFeedStore.cs
```diff
@@ -100,6 +100,10 @@ private void OnContentRemoved(ContentRemovalSignal removal)
         if (string.Equals(removal.Kind, ContentRemovalKinds.Comment, StringComparison.Ordinal))
```
XeldarAlz, line 100: Non-blocking: no `removal.App` guard, and this store is instantiated for both Chirper and Aethergram, so both handlers run for every removal. Harmless while post ids are globally unique, but the new bump turns any cross-app id collision into a silently wrong count.

src/Aetherphone/Apps/Chirper/ChirperApp.cs
```diff
@@ -1080,7 +1080,7 @@ private void DrawThread(Rect area, string postId)
             }
 
             var comments = store.DetailComments;
-            var commentTotal = store.HasMoreComments ? Math.Max(post.CommentCount, comments.Length) : comments.Length;
+            var commentTotal = post.CommentCount;
```
XeldarAlz, line 1083: Blocking: dropping the `Math.Max(post.CommentCount, comments.Length)` floor lets the header print fewer replies than are rendered right below it. With the double-decrement in `OnContentRemoved`, deleting a reply shows `Replies N-2` over N-1 rows. Keeping `Math.Max` but making it unconditional fixes the badge/header mismatch without the under-count hole.

XeldarAlz, line 1083: Blocking: `OpenThreadFromLink` -> `OpenDetailById` (:1366) fetches by id, so a link to a repost id puts the wrapper in `detailPost`. `DrawPost` unwraps to `ReferencedPost`, but this line reads the wrapper's own `CommentCount`, so the header shows 0 while the original's replies list below it.

XeldarAlz, line 1083: Blocking: `AethergramApp.Detail.cs:218` and `VelvetShell.Detail.cs:223` still use the old formula against the same shared store, so the reply header now behaves differently per app. Whichever formula wins should apply to all three.

src/Aetherphone/Core/Aethernet/CopyOnWrite.cs
```diff
@@ -199,6 +199,34 @@ public static T[] MapById<T>(T[] source, string id, Func<T, T> transform) where
         return source;
     }
 
+    // Replaces a post that appears directly in the array, AND updates the ReferencedPost
+    // inside any rechirp wrappers (posts whose RepostOfId matches the updated post's Id).
+    public static PostDto[] ReplaceOrUpdateReferenced(PostDto[] source, PostDto updated)
```
XeldarAlz, line 204: Blocking: `CopyOnWrite` is a domain-free generic helper, every other member is `T`. This pulls Chirper's rechirp concept into `Core/Aethernet` and re-implements `CopyOnWrite.Map`. It belongs next to `MapCommentCount`/`MapFollowByAuthor` in `SocialFeedStore`, where both new mappers collapse into one `MapPostAndWrappers(source, postId, transform)`.

src/Aetherphone/Core/Aethernet/CopyOnWrite.cs
```diff
@@ -199,6 +199,34 @@ public static T[] MapById<T>(T[] source, string id, Func<T, T> transform) where
         return source;
     }
 
+    // Replaces a post that appears directly in the array, AND updates the ReferencedPost
```
XeldarAlz, line 202: Nit: CLAUDE.md forbids comments unless they carry a why the code cannot express, and this one restates the what. Moving the method into `SocialFeedStore` as suggested removes the need for it.

src/Aetherphone/Core/Aethernet/CopyOnWrite.cs
```diff
@@ -199,6 +199,34 @@ public static T[] MapById<T>(T[] source, string id, Func<T, T> transform) where
...
+        var result = new PostDto[source.Length];
+        for (var index = 0; index < source.Length; index++)
+        {
+            var item = source[index];
+            if (item.Id == updated.Id)
+            {
+                result[index] = updated;
+                changed = true;
+            }
+            else if (item.RepostOfId == updated.Id && item.ReferencedPost is not null)
```
XeldarAlz, line 216: Nit: `src/Aetherphone.Tests` already covers this class of pure helper (`FeedLaneTests`, `IdentifiedMergeTests`). Two cases would pin the new behavior: the wrapper's `ReferencedPost` gets updated, and unchanged input comes back by reference.

src/Aetherphone/Core/Social/SocialFeedStore.cs
```diff
@@ -1023,9 +1027,9 @@ protected void AcceptCreatedPost(PostDto created)
 
     protected void ReplacePost(PostDto updated)
     {
-        forYouLane.Items = CopyOnWrite.Replace(forYouLane.Items, updated);
-        followingLane.Items = CopyOnWrite.Replace(followingLane.Items, updated);
-        profileLane.Items = CopyOnWrite.Replace(profileLane.Items, updated);
+        forYouLane.Items = CopyOnWrite.ReplaceOrUpdateReferenced(forYouLane.Items, updated);
```
XeldarAlz, line 1030: Non-blocking: `savedLane` and `taggedLane` are still untouched here and in `BumpCommentCount`, so Saved and Tagged keep stale reaction and comment counts. `detailPost` below is also still matched by `Id` only, so a wrapper in detail keeps a stale embedded original.

src/Aetherphone/Core/Social/SocialFeedStore.cs
```diff
@@ -1089,8 +1093,16 @@ private static int FollowerDelta(bool wasFollowing, bool following) =>
         wasFollowing == following ? 0 : following ? 1 : -1;
 
     private static PostDto[] MapCommentCount(PostDto[] source, string postId, int delta) =>
-        CopyOnWrite.MapById(source, postId,
-            post => post with { CommentCount = Math.Max(0, post.CommentCount + delta) });
+        CopyOnWrite.Map(source,
+            post => post.Id == postId || post.RepostOfId == postId,
```
XeldarAlz, line 1097: Non-blocking: the predicate matches `RepostOfId == postId` without checking `ReferencedPost`, so a wrapper whose original is unavailable (`ChirperApp.cs:385`) hits `post with { ReferencedPost = null }`: a fresh record plus `changed = true`, forcing a full array copy for a no-op. `ReplaceOrUpdateReferenced` does check.

src/Aetherphone/Core/Social/SocialFeedStore.cs
```diff
@@ -1089,8 +1093,16 @@ private static int FollowerDelta(bool wasFollowing, bool following) =>
...
+            post => post.Id == postId
+                ? post with { CommentCount = Math.Max(0, post.CommentCount + delta) }
+                : post with
+                {
+                    ReferencedPost = post.ReferencedPost is { } referenced
+                        ? referenced with { CommentCount = Math.Max(0, referenced.CommentCount + delta) }
+                        : null
```
XeldarAlz, line 1104: Non-blocking: `MapFollowByAuthor` (:1117) has the identical wrapper hole, follow state on a rechirped original never updates. One line once the shared `MapPostAndWrappers` helper exists.

XeldarAlz (changes requested):
Build passes on ccfd0e23 (Release, 0 errors). Not exercised in game.

The diagnosis is right: `DrawPost` renders `ReferencedPost`, so matching on `Id` alone could never reach what is actually on screen. Blockers are inline, chiefly a double-decrement on self-delete that the old handler deliberately avoided.

Title nit: repo style is imperative, e.g. `fix(social): keep rechirp wrappers in sync with comment and reaction counts`.

Farroness:
Alright this is involving a lot more wiring than I anticipated. I'll review your comments and keep researching. This bug might be a little over my head.

Farroness:
Alright going to close this as the issue wasn't as simple as I thought it was and I have other beginner friendly bugs/feats I can work on instead, going to leave this to one of the more experienced devs on the team.

PR #101, opened 2026-08-08
fix(crypto): derive recovery-code key with managed PBKDF2

BluntEXE:
## Summary

Recovery-code setup fails on some Wine builds with `Unknown error (0xc1000008)`, thrown from `Pbkdf2Implementation.FillKeyDerivation` inside `RecoveryKey.Wrap`. `.NET`'s `Rfc2898DeriveBytes.Pbkdf2` one-shot method calls into Windows CNG (`BCryptDeriveKeyPBKDF2`) on Windows-targeted builds. A tester isolated the failure by trying SHA-256 and SHA-1, one iteration and 600k: every combination threw the same error, ruling out a parameter problem and pointing at the CNG call itself failing under their Wine build.

`DeriveKey` now uses the managed `Rfc2898DeriveBytes` instance API (`new Rfc2898DeriveBytes(password, salt, iterations, HashAlgorithmName.SHA256)` + `GetBytes`) instead of the static one-shot method. This path skips the CNG call that crashed on Wine, and it zeroes its internal HMAC state on `Dispose`, closing a gap where secret material could outlive the surrounding `ZeroMemory`. The constructor is marked obsolete under `SYSLIB0060` in favor of the static method we're deliberately avoiding, so the warning is suppressed locally.

An earlier revision of this PR used BouncyCastle's `Pkcs5S2ParametersGenerator` for the same reason. Switched to the managed API after review flagged that BouncyCastle's `Init`/`GetKey` clone the password and key buffers internally without zeroing them.

Output is byte-identical to the original implementation (both are standard PBKDF2-HMAC-SHA256), so existing recovery codes and wrapped keys keep working with no migration step. `DeriveKeyMatchesRfc7914Pbkdf2HmacSha256Vector` pins the output against both PBKDF2-HMAC-SHA256 vectors from RFC 7914 section 11, so that claim is enforced in CI rather than argued from the spec alone.

## Test plan

- [x] All crypto tests pass, including the new pinned RFC 7914 vector test.
- [x] Verified live on the tester's Wine build with the BouncyCastle version: recovery code setup succeeded with no errors, warnings, or exceptions in the log.
- [x] Re-verified on the tester's build with the managed `Rfc2898DeriveBytes` version: recovery code setup still succeeds.


src/Aetherphone/Core/Crypto/RecoveryKey.cs
```diff
@@ -144,7 +147,10 @@ private static byte[] DeriveKey(string canonicalCode, byte[] salt, int iteration
         var password = Encoding.UTF8.GetBytes(canonicalCode);
         try
         {
-            return Rfc2898DeriveBytes.Pbkdf2(password, salt, iterations, HashAlgorithmName.SHA256, WrapKeyBytes);
+            var generator = new Pkcs5S2ParametersGenerator(new Sha256Digest());
+            generator.Init(password, salt, iterations);
+            var derived = (KeyParameter)generator.GenerateDerivedParameters("AES", WrapKeyBytes * 8);
+            return derived.GetKey();
```
XeldarAlz: Init clones the password and GetKey returns a copy of the KeyParameter's internal buffer, and neither internal array is ever zeroed, so secret material now outlives the ZeroMemory in the finally. The managed instance Rfc2898DeriveBytes also skips the CNG one-shot and zeroes its internals on Dispose (needs a SYSLIB0060 pragma), if you want to keep the hygiene intact.

src/Aetherphone/Core/Crypto/RecoveryKey.cs
```diff
@@ -144,7 +147,10 @@ private static byte[] DeriveKey(string canonicalCode, byte[] salt, int iteration
         var password = Encoding.UTF8.GetBytes(canonicalCode);
         try
         {
-            return Rfc2898DeriveBytes.Pbkdf2(password, salt, iterations, HashAlgorithmName.SHA256, WrapKeyBytes);
+            var generator = new Pkcs5S2ParametersGenerator(new Sha256Digest());
```
XeldarAlz: The existing tests roundtrip through the new code on both sides, so byte compatibility with escrows written by the old implementation rests on the spec argument alone. A pinned PBKDF2-HMAC-SHA256 vector test (RFC 7914 section 11 ships two) would lock it in CI.

XeldarAlz (commented):
Checked this against the current dev encryption state: DeriveKey there still routes through the CNG one-shot, and BouncyCastle already ships for the same Wine reason since 60edb655, so the fix stays valid as is. Two nits inline.

src/Aetherphone/Core/Crypto/RecoveryKey.cs
BluntEXE: Switched to the managed Rfc2898DeriveBytes instance API (SYSLIB0060 pragma'd) instead of BouncyCastle. It skips the CNG one-shot path that caused the original crash and zeroes its internal HMAC state on Dispose, so the derived-key hygiene is back to matching the finally block's ZeroMemory. Pushed in d52c3f0e.

src/Aetherphone/Core/Crypto/RecoveryKey.cs
BluntEXE: Added DeriveKeyMatchesRfc7914Pbkdf2HmacSha256Vector, pinned to both PBKDF2-HMAC-SHA256 vectors from RFC 7914 section 11 (truncated to our 32-byte key length, cross-checked against Python's hashlib before writing them in). Pushed in d52c3f0e.

BluntEXE:
Heads up: the failing `Build (Windows)` check here isn't from this PR. It's 19 pre-existing failures in the casino/bingo/wheel/scratch tests (`CasinoWheelWireContractTests`, `BingoRulesTests`, `CasinoBingoWireContractTests`, `WheelRulesTests`, `BarkeepRulesTests`, `ScratchRulesTests`), inherited from merging master in to resolve the conflict in `RecoveryKeyTests.cs`. Confirmed master's own CI is already red on the tip commit (0ac23c11, run 31436589250) from the casino work, before this branch touched it. All `RecoveryKeyTests`, old and new, pass in this run.

XeldarAlz (approved):
One small thing before merge: the PR body still describes the BouncyCastle approach, and the live Wine verification it cites predates the switch to the managed API. Worth refreshing the body, and ideally a quick re-run on the tester's box.

BluntEXE:
Re-verified with the tester on the managed `Rfc2898DeriveBytes` build (commit 311016ef). Recovery code setup still succeeds on their Wine build, no errors. Ready for another look.

PR #110, opened 2026-08-09
fix(games): fixed text wrapping issue in solitaire for cards that are on the left most slot.

Farroness:
## What

Replaced typography.drawcentered call with a imgui draw to correct text wrapping from breaking on certain cards in solitaire that show on the left move slot. 

## Why

Typography.DrawCentered AutoWrapWidth did not play nicely with how solitaire labels are setup. Fixing the issue within Typography.cs has a wide blast radius so the imgui solution is cleaner for this specific bug.

Closes # https://discord.com/channels/1527213298087493732/1532479753737474088

## How to test

1. Start solitaire.
2. Create a stack that includes "K" or "10" labels. 
3. Move stack to left most slot.
4. Observe whether the card label still wraps.

## Checklist

### DIFF
--- a/src/Aetherphone/Windows/Components/PlayingCards.cs
+++ b/src/Aetherphone/Windows/Components/PlayingCards.cs
@@ -86,7 +86,7 @@ internal static class PlayingCards
         var labelScale = MathF.Max(0.5f, MathF.Min(0.9f, rect.Width / (40f * scale)));
         var label = RankLabel(Rank(card));
         var corner = new Vector2(rect.Min.X + rect.Width * 0.18f, rect.Min.Y + rect.Height * 0.16f);
-        Typography.DrawCentered(drawList, corner, label, color, labelScale, FontWeight.Bold);
+        DrawRankLabel(drawList, corner, label, color, labelScale);
         DrawSuit(drawList, new Vector2(corner.X, corner.Y + rect.Height * 0.2f), rect.Width * 0.1f, suit, color);
         DrawSuit(drawList, new Vector2(rect.Center.X, rect.Center.Y + rect.Height * 0.08f), rect.Width * 0.24f, suit,
             color);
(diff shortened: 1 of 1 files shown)


XeldarAlz:
Did you verify in game? Boxes are unchecked

Farroness:
I verified through several solitaire games ingame. 

Fix:
image

image

Farroness:
Here's a screenshot of the bug on our current Aetherphone version just for comparison:

image

XeldarAlz (approved):
lgtm

src/Aetherphone/Apps/Games/Solitaire/SolitaireRenderer.cs
```diff
@@ -160,7 +161,17 @@ private void DrawCardFace(ImDrawListPtr drawList, in Rect rect, int card, float
...
+        using (Plugin.Fonts.Push(labelScale, FontWeight.Bold))
+            {
+                Plugin.Fonts.NoticeText(label);
+                var size = ImGui.CalcTextSize(label);
+                ImGui.SetCursorScreenPos(corner - size * 0.5f);
+                using (ImRaii.PushColor(ImGuiCol.Text, color))
+                {
+                    ImGui.TextUnformatted(label);
+                }
+            }
```
XeldarAlz: `SetCursorScreenPos` + `TextUnformatted` submits a layout item mid-draw (moves the cursor and grows the window content size every frame, including while a card is dragged) and ignores the `drawList` this method already receives. Use the draw-list path, same as `DrawCentered`'s no-wrap branch:

```suggestion
        using (Plugin.Fonts.Push(labelScale, FontWeight.Bold))
        {
            Plugin.Fonts.NoticeText(label);
            var size = ImGui.CalcTextSize(label);
            drawList.AddText(ImGui.GetFont(), ImGui.GetFontSize(), corner - size * 0.5f, ImGui.GetColorU32(color),
                label);
        }
```

src/Aetherphone/Apps/Games/Solitaire/SolitaireRenderer.cs
```diff
@@ -3,6 +3,7 @@
 using Aetherphone.Core.Theme;
 using Aetherphone.Windows.Components;
 using Dalamud.Bindings.ImGui;
+using Dalamud.Interface.Utility.Raii;
```
XeldarAlz: Goes unused once the label renders through `drawList.AddText`.

XeldarAlz (changes requested):
Diagnosis is right: both DrawCentered overloads auto-wrap, so bypassing Typography here is fine. One change needed before merge: render the label through the draw list instead of ImGui items, suggestion inline.

XeldarAlz:
Pushed the change onto this branch rather than waiting on a round trip.

Master moved the card art out of the solitaire renderer into the shared painter at `src/Aetherphone/Windows/Components/PlayingCards.cs`, so `SolitaireRenderer.DrawCardFace` (the method this branch edited) no longer exists and the branch was conflicting. I rebased onto master and applied the same fix in its new home, through the draw list as suggested, so the blackjack felt gets it too.

`DrawRankLabel` paints the rank with `drawList.AddText`, which is the branch `Typography.DrawCentered` already takes when the text fits: no auto wrap width, no layout item submitted mid draw.

Build and tests are green. Note the branch was force pushed, so run `git fetch && git reset --hard origin/fix-solitaire-width` before working on it again.

PR #111, opened 2026-08-09
feat(composer): support shift+enter newlines in chat and post composers

### LINKED ISSUE #108: [Feature] Ability to add new lines in textboxes
### What problem are you trying to solve?

Messages now cannot be formatted with extra, manual, new lines.

### What did you have in mind?

I would like to be able to use shift+enter to create a newline, this way i can structure my messages better.

### Alternatives you considered

_No response_

### Anything else?

_No response_

TomSchillemans:
close #108 

## Summary

Adds shift+Enter newline support to the chat composer and post composers, and fixes two related bugs.

## Changes

- **ChatComposer**: Reworked to use a multiline input with a newline char filter that allows `\n` only when shift is held; plain Enter submits.
- **Duplicate DM fix**: Added a `composerEpoch`-based widget ID and an Enter key-press guard to prevent duplicate DM sends caused by the stale `InputText` buffer re-syncing after a send.
- **Newline cutoff fix**: Fixed a trailing-newline height undercount in `DrawTextBubble` so messages ending in a newline are not cut off.
- **Aethergram / Chirper**: Enabled `allowNewlines` for post captions and compose bodies.

## Testing

- Shift+Enter inserts a newline in DMs and post captions; plain Enter sends.
- No duplicate DM messages on repeated sends.
- Messages ending in a newline render fully without being cut off.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ obj/
 bin/
 *.user
 .vscode/
+.kiro/
 
 # Node tooling (tools/icon-generator)
 node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ bin/
 *.user
 .vscode/
 .kiro/
+.abacusai/
+kilo.jsonc
 
 # Node tooling (tools/icon-generator)
 node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -4,9 +4,6 @@ obj/
 bin/
 *.user
 .vscode/
-.kiro/
-.abacusai/
-kilo.jsonc
 
 # Node tooling (tools/icon-generator)
 node_modules/
--- a/src/Aetherphone/Windows/Components/ChatComposer.cs
+++ b/src/Aetherphone/Windows/Components/ChatComposer.cs
@@ -30,6 +30,9 @@ internal sealed class ChatComposer : IDisposable
 {
     private const int TextKind = 0;
     private const float AccessoryBarHeight = 46f;
+    private const float ComposerBaseHeight = 56f;
+    private const float ComposerLineHeight = 20f;
+    private const int ComposerMaxLines = 4;
     private static readonly Vector4 White = new(1f, 1f, 1f, 1f);
     private static readonly Vector4 FieldFill = new(1f, 1f, 1f, 0.10f);
     private static readonly Vector4 BarFill = new(1f, 1f, 1f, 0.05f);
@@ -37,6 +40,7 @@ internal sealed class ChatComposer : IDisposable
     private readonly VoiceNoteRecorder recorder = new();
     private readonly EmojiPicker emojiPicker = new();
     private string draft = string.Empty;
+    private int composerEpoch;
     private bool focus;
     private bool emojiOpen;
     private string? replyTargetId;
@@ -60,6 +64,44 @@ internal sealed class ChatComposer : IDisposable
     public float AccessoryHeight => replyTargetId is not null || editTargetId is not null
         ? AccessoryBarHeight * UiScale.Current
         : 0f;
+
+    public float ComposerHeight
+    {
+        get
+        {
+            var scale = UiScale.Current;
+            var lines = Math.Clamp(CountLines(draft), 1, ComposerMaxLines);
+            return (ComposerBaseHeight + (lines - 1) * ComposerLineHeight) * scale;
+        }
+    }
+
+    private static int NewlineFilterCallback(ImGuiInputTextCallbackDataPtr data)
+    {
+        if (data.EventFlag == ImGuiInputTextFlags.CallbackCharFilter)
+        {
+            if (data.EventChar is '\n' or '\r')
+            {
+                var shift = ImGui.IsKeyDown(ImGuiKey.LeftShift) || ImGui.IsKeyDown(ImGuiKey.RightShift);
+                data.EventChar = shift ? '\n' : (char)0;
+            }
+        }
+
+        return 0;
+    }
+
+    internal static int CountLines(string text)
+    {
+        var lines = 1;
+        for (var index = 0; index < text.Length; index++)
+        {
+            if (text[index] == '\n')
+            {
+                lines++;
+            }
+        }
+
+        return lines;
+    }
 
     public void BeginReply(string messageId, string senderName, string preview)
     {
@@ -225,9 +267,9 @@ internal sealed class ChatComposer : IDisposable
         }
 
         var textLeft = emojiMax.X + 4f * scale;
-        ImGui.SetCursorScreenPos(new Vector2(textLeft,
-            (pillMin.Y + pillMax.Y) * 0.5f - ImGui.GetFrameHeight() * 0.5f));
-        ImGui.SetNextItemWidth(MathF.Max(1f, textRight - textLeft));
+        ImGui.SetCursorScreenPos(new Vector2(textLeft, pillMin.Y + 4f * scale));
+        var inputSize = new Vector2(MathF.Max(1f, textRight - textLeft),
+            MathF.Max(1f, pillMax.Y - pillMin.Y - 8f * scale));
         if (focus)
         {
             ImGui.SetKeyboardFocusHere();
@@ -235,15 +277,20 @@ internal sealed class ChatComposer : IDisposable
         }
 
         var submitted = false;
+        var shiftDown = ImGui.IsKeyDown(ImGuiKey.LeftShift) || ImGui.IsKeyDown(ImGuiKey.RightShift);
         Plugin.Fonts.NoticeText(draft);
         using (ImRaii.PushColor(ImGuiCol.FrameBg, new Vector4(0f, 0f, 0f, 0f)))
         using (ImRaii.PushColor(ImGuiCol.Text, theme.TextStrong))
         {
-            if (ImGui.InputTextWithHint("##chatComposerInput", Loc.T(L.Velvet.MessageHint), ref draft, model.MaxLength,
-                    ImGuiInputTextFlags.EnterReturnsTrue))
-            {
-                submitted = true;
-            }
+            ImGui.InputTextMultiline("##chatComposerInput" + composerEpoch, ref draft, model.MaxLength, inputSize,
+                ImGuiInputTextFlags.None, NewlineFilterCallback);
+        }
+
+        if (draft.Length == 0)
+        {
+            var hintPosition = new Vector2(textLeft + ImGui.GetStyle().FramePadding.X,
+                pillMin.Y + 4f * scale + ImGui.GetStyle().FramePadding.Y);
+            drawList.AddText(hintPosition, ImGui.GetColorU32(ui.MutedInk), Loc.T(L.Velvet.MessageHint));
         }
 
         var hasDraft = draft.Trim().Length > 0;
@@ -285,6 +332,11 @@ internal sealed class ChatComposer : IDisposable
             AppSkin.Icon(sendCenter, FontAwesomeIcon.PaperPlane.ToIconString(), White, 0.9f);
         }
 
+        if (ImGui.IsKeyPressed(ImGuiKey.Enter, false) && !shiftDown)
+        {
+            submitted = true;
+        }
+
         if (submitted && canSend)
         {
             if (editTargetId is { } editId)
@@ -299,6 +351,7 @@ internal sealed class ChatComposer : IDisposable
                 ClearReply();
             }
 
+            composerEpoch++;
             emojiOpen = false;
             focus = true;
         }
--- a/src/Aetherphone/Windows/Components/ChatComposer.cs
+++ b/src/Aetherphone/Windows/Components/ChatComposer.cs
@@ -41,6 +41,7 @@ internal sealed class ChatComposer : IDisposable
     private readonly EmojiPicker emojiPicker = new();
     private string draft = string.Empty;
     private int composerEpoch;
+    private string inputLabel = "##chatComposerInput0";
     private bool focus;
     private bool emojiOpen;
     private string? replyTargetId;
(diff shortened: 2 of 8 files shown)


Farroness:
Hi @TomSchillemans Just FYI, Xel the maintainer mentioned that they're pretty busy this week so it may be a little bit before he can review new PRs. He's also getting things setup for a new release this week I believe.

src/Aetherphone/Windows/Components/ChatComposer.cs
```diff
@@ -285,6 +332,11 @@ private void DrawInputComposer(Rect area, in ChatComposerModel model)
             AppSkin.Icon(sendCenter, FontAwesomeIcon.PaperPlane.ToIconString(), White, 0.9f);
         }
 
+        if (ImGui.IsKeyPressed(ImGuiKey.Enter, false) && !shiftDown)
```
XeldarAlz: This fires on any Enter press while the thread view is drawn, even when the input is not focused (thread search field, possibly game chat since Dalamud feeds keys into ImGui io). Capture `ImGui.IsItemActive()` right after the `InputTextMultiline` call and gate the submit on it; `EnterReturnsTrue` only fired while the input was active.

XeldarAlz: The old `EnterReturnsTrue` path also accepted `KeypadEnter`; checking only `ImGuiKey.Enter` drops numpad Enter submits.

src/Aetherphone/Windows/Components/ChatComposer.cs
```diff
@@ -225,25 +267,30 @@ private void DrawInputComposer(Rect area, in ChatComposerModel model)
...
         Plugin.Fonts.NoticeText(draft);
         using (ImRaii.PushColor(ImGuiCol.FrameBg, new Vector4(0f, 0f, 0f, 0f)))
         using (ImRaii.PushColor(ImGuiCol.Text, theme.TextStrong))
         {
-            if (ImGui.InputTextWithHint("##chatComposerInput", Loc.T(L.Velvet.MessageHint), ref draft, model.MaxLength,
-                    ImGuiInputTextFlags.EnterReturnsTrue))
-            {
-                submitted = true;
-            }
+            ImGui.InputTextMultiline("##chatComposerInput" + composerEpoch, ref draft, model.MaxLength, inputSize,
```
XeldarAlz: This concatenates a new ID string every frame in a draw path. Cache the label in a field and rebuild it only when `composerEpoch` increments.

src/Aetherphone/Windows/Components/SoftWrap.cs
```diff
@@ -102,18 +116,39 @@ public static string WrapText(string text, float wrapWidth)
         return builder.ToString();
     }
 
-    public static string StripNewlines(string text)
+    public static string StripSoftWraps(string text)
```
XeldarAlz, line 126: Nothing stops a pasted U+200B from entering the buffer (common in text copied from the web or Discord), so it gets treated as a soft-wrap marker here and can swallow a real newline that follows it. Dropping `EventChar == SoftWrapSentinel` in the char filter keeps the sentinel out-of-band.

src/Aetherphone/Windows/Components/ChatTranscript.cs
```diff
@@ -530,6 +530,10 @@ private void DrawTextBubble(TranscriptMessage message, int index, in ChatTranscr
         var wrap = available * 0.74f - paddingX * 2f;
         var linkLayout = deleted ? null : LinkText.LayoutFor(message.Body, wrap);
         var textSize = linkLayout is null ? ImGui.CalcTextSize(message.Body, false, wrap) : linkLayout.Size;
+        if (linkLayout is null && EndsWithNewline(message.Body))
```
XeldarAlz, line 533: This compensation skips the `linkLayout` branch, so a message containing a URL and ending in a newline can still clip. Worth checking whether `RichText.Build` has the same trailing-newline undercount.

src/Aetherphone/Windows/Components/ChatComposer.cs
```diff
@@ -61,6 +65,44 @@ public string Draft
...
+    }
+
+    private static int NewlineFilterCallback(ImGuiInputTextCallbackDataPtr data)
+    {
+        if (data.EventFlag == ImGuiInputTextFlags.CallbackCharFilter)
+        {
+            if (data.EventChar is '\n' or '\r')
+            {
+                var shift = ImGui.IsKeyDown(ImGuiKey.LeftShift) || ImGui.IsKeyDown(ImGuiKey.RightShift);
+                data.EventChar = shift ? '\n' : (char)0;
```
XeldarAlz: Pasted characters flow through this filter with shift up, so pasting multiline text flattens it to one line. Allowing `\n` when Enter is not currently pressed would distinguish paste from a typed Enter (same applies to `SoftWrap.ApplyEdit`).

.gitignore
```diff
@@ -5,6 +5,8 @@ bin/
 *.user
 .vscode/
 .kiro/
+.abacusai/
```
XeldarAlz: These look like local AI-tool artifacts, unrelated to the composer change. Better suited to a global gitignore or `.git/info/exclude`.

src/Aetherphone/Windows/Components/ChatComposer.cs
```diff
@@ -225,25 +269,32 @@ private void DrawInputComposer(Rect area, in ChatComposerModel model)
...
+        bool inputActive;
         using (ImRaii.PushColor(ImGuiCol.FrameBg, new Vector4(0f, 0f, 0f, 0f)))
         using (ImRaii.PushColor(ImGuiCol.Text, theme.TextStrong))
         {
-            if (ImGui.InputTextWithHint("##chatComposerInput", Loc.T(L.Velvet.MessageHint), ref draft, model.MaxLength,
-                    ImGuiInputTextFlags.EnterReturnsTrue))
-            {
-                submitted = true;
-            }
+            ImGui.InputTextMultiline(inputLabel, ref draft, model.MaxLength, inputSize,
```
XeldarAlz, line 288: Blocking: flags are `None` here, so `CallbackCharFilter` is never enabled and `NewlineFilterCallback` never runs; ImGui gates the callback on the flag. Enter therefore always inserts a newline, hidden in normal use only by the `Trim()` in `SendMessage`. Press Enter while `Sending` or on an empty draft: `canSend` is false, the newline stays, and repeats leave the draft as four blank lines with the hint gone and the box pinned at 4 lines. Pass `ImGuiInputTextFlags.CallbackCharFilter`.

src/Aetherphone/Windows/Components/SoftWrap.cs
```diff
@@ -162,12 +204,12 @@ private static float MeasureRange(StringBuilder builder, int start)
         return ImGui.CalcTextSize(builder.ToString(start, builder.Length - start)).X;
     }
 
-    private static int CountNewlines(string text, int limit)
+    private static int CountNonLogical(string text, int limit)
```
XeldarAlz, line 207: Blocking: this skips every newline, but `StripSoftWraps` only drops the one that follows a sentinel, so with `allowNewlines` the cursor runs one short per hard newline while the string keeps them. In Chirper, `Hello` then shift+Enter then `@xel` gives cursor 9 on a 10 char string: `MentionTokenScanner` queries `xe`, and committing writes `Hello`, a newline, then `@handle l`. `LogicalToWrappedIndex` (line 243) has the same rule. Skip the sentinel and only the newline right after it.

src/Aetherphone/Windows/Components/ChatComposer.cs
```diff
@@ -30,13 +30,18 @@ internal sealed class ChatComposer : IDisposable
 {
     private const int TextKind = 0;
     private const float AccessoryBarHeight = 46f;
+    private const float ComposerBaseHeight = 56f;
+    private const float ComposerLineHeight = 20f;
```
XeldarAlz, line 34: Non-blocking: 20f is not the widget's line height. `Fonts.Push(1f)` sizes off `UiBuilder.DefaultFontSizePx`, which follows the Dalamud font size setting, so above the default a line is taller than `20 * scale` and the 4 line box clips. Derive from `ImGui.GetTextLineHeight()` plus `FramePadding.Y * 2`.

src/Aetherphone/Windows/Components/ChatComposer.cs
```diff
@@ -61,6 +66,45 @@ public string Draft
         ? AccessoryBarHeight * UiScale.Current
         : 0f;
 
+    public float ComposerHeight
+    {
+        get
+        {
+            var scale = UiScale.Current;
+            var lines = Math.Clamp(CountLines(draft), 1, ComposerMaxLines);
```
XeldarAlz, line 74: Non-blocking: `CountLines` counts only newlines, so a long paragraph with none keeps the composer at 56px and the widget scrolls internally instead of growing. The scrollbar that then appears narrows the text column mid-typing, which is the StableContentWidth shake.

src/Aetherphone/Windows/Components/ChatComposer.cs
```diff
@@ -225,25 +269,32 @@ private void DrawInputComposer(Rect area, in ChatComposerModel model)
...
+            ImGui.InputTextMultiline(inputLabel, ref draft, model.MaxLength, inputSize,
+                ImGuiInputTextFlags.None, NewlineFilterCallback);
+            inputActive = ImGui.IsItemActive();
+        }
+
+        if (draft.Length == 0)
+        {
+            var hintPosition = new Vector2(textLeft + ImGui.GetStyle().FramePadding.X,
+                pillMin.Y + 4f * scale + ImGui.GetStyle().FramePadding.Y);
+            drawList.AddText(hintPosition, ImGui.GetColorU32(ui.MutedInk), Loc.T(L.Velvet.MessageHint));
```
XeldarAlz, line 297: Non-blocking: `AddText` takes no wrap width or clip rect, so unlike `InputTextWithHint` the hint is not bounded by the field. At the 240px phone width with the image and location icons both on, the text column is around 180 design px and the de, pt and ru `velvet.messageHint` are close to that, so it draws over the trailing icons. Pass `inputSize.X` as the wrap width.

src/Aetherphone/Windows/Components/ChatComposer.cs
```diff
@@ -299,6 +356,8 @@ private void DrawInputComposer(Rect area, in ChatComposerModel model)
                 ClearReply();
             }
 
+            composerEpoch++;
```
XeldarAlz, line 359: Non-blocking: the epoch resets only after a send, but `Draft`, `BeginEdit`, `ClearEdit`, `Clear` and the emoji append all write `draft` behind the same active widget. If the stale buffer is real, opening a thread with a saved draft or starting an edit while the field is focused hits it too. Fold this into one `ResetInput()` that those paths call.

src/Aetherphone/Windows/Components/SoftWrap.cs
```diff
@@ -102,18 +123,39 @@ public static string WrapText(string text, float wrapWidth)
         return builder.ToString();
     }
 
-    public static string StripNewlines(string text)
+    public static string StripSoftWraps(string text)
```
XeldarAlz, line 126: Non-blocking: `SoftWrap` is pure and `Aetherphone.Tests` exists, but nothing covers it. A `WrapText` to `StripSoftWraps` round trip, plus a cursor mapping case over text with hard newlines, is exactly what would have caught the off-by-one above.

src/Aetherphone/Windows/Components/SoftWrap.cs
```diff
@@ -75,14 +96,14 @@ public static string WrapText(string text, float wrapWidth)
             {
                 if (wordStart > lineStart)
                 {
-                    builder.Insert(wordStart, '\n');
-                    lineStart = wordStart + 1;
+                    builder.Insert(wordStart, SoftWrapSentinel + "\n");
+                    lineStart = wordStart + 2;
```
XeldarAlz, line 100: Nit: the `+ 2` hides the length of the sentinel plus newline pair in a literal. A named const for the pair keeps this in step if the marker ever changes.

src/Aetherphone/Windows/Components/ChatTranscript.cs
```diff
@@ -530,6 +530,12 @@ private void DrawTextBubble(TranscriptMessage message, int index, in ChatTranscr
...
+        {
+            // RichText.Build already counts a trailing newline as a blank line via its own
```
XeldarAlz, line 535: Nit: comments are out per the contributor rules. The reasoning does check out though: `RichText.Build` ends on `y + lineHeight` while `CalcTextSize` adds a final line only when the last one is non-empty, so the `+1` is right for any number of trailing newlines.

src/Aetherphone/Windows/Components/ChatComposer.cs
```diff
@@ -61,6 +66,45 @@ public string Draft
...
+                var shift = ImGui.IsKeyDown(ImGuiKey.LeftShift) || ImGui.IsKeyDown(ImGuiKey.RightShift);
+                var enterPressed = ImGui.IsKeyDown(ImGuiKey.Enter) || ImGui.IsKeyDown(ImGuiKey.KeypadEnter);
+                data.EventChar = shift || !enterPressed ? '\n' : (char)0;
+            }
+        }
+
+        return 0;
+    }
+
+    internal static int CountLines(string text)
```
XeldarAlz, line 94: Nit: `internal` with no caller outside this class. `private` unless a test needs it.

XeldarAlz (changes requested):
Built the branch in an isolated worktree: Release 0 errors, 1144/1144 tests green. Not run in-game, so the ImGui findings come from reading the widget flags rather than watching the composer. Two blocking, the rest are follow-ups. The branch also needs a rebase onto dev, where these five files moved under Windows/Components/Chat and Windows/Components/Primitives.

PR #113, opened 2026-08-10
feat(velvet)Persistent and manageable "Not Interested"

Deldee:
## What
Updated Velvet's "not interested" button to : 
- Be persistent using plugin config json file
- Be reversible, using a new section under the setting app
- Be reset in between account

## Why

Velvet's "not interested" only lived in memory, it went away as soon as the phone was closed and reloaded, 

## How to test
- Open the app in game
- Open Velvet
- Mark any profile as "Not interested"
- Search for said profile (it should not appear)
- Reload the plugin
- Search for the profile again (it should not appear)
- Navigate to Setting > "Not Interested"
- The list of Hidden users should appear
- Click "Remove"
- Search for the profile again (It should appear)

## Checklist


src/Aetherphone/Core/Localization/L.cs
```diff
@@ -3035,6 +3035,8 @@ internal static class Velvet
         public static readonly LocString WhoHelp =
             new("velvet.whoHelp", "Choose who can send you a one line intro. Friends means friends of friends.");
         public static readonly LocString BlockedNone = new("velvet.blockedNone", "No one blocked.");
+        public static readonly LocString NotInterestedNone = new("velvet.notInterestedNone", "No one marked as not interested.");
```
XeldarAlz, line 3038: These two new keys need matching entries in all nine locale JSONs under src/Aetherphone/Localization/ (de, en, es, fr, ja, pt, ru, tr, zh) in the same commit, per the localization lockstep rule.

src/Aetherphone/Apps/Velvet/VelvetStore.cs
```diff
@@ -234,7 +247,10 @@ protected override void OnAccountSwitched()
         accessBlocked = false;
         meGate.Reset();
         discoverResults = Array.Empty<VelvetProfileDto>();
+
         hiddenFromDiscover = Array.Empty<string>();
+        notInterestedLoaded = false;
```
XeldarAlz, line 253: Clear the cached profiles here as well (`notInterested = Array.Empty<VelvetProfileDto>();`). Until the refresh completes, the settings page still renders the list from the previous account.

src/Aetherphone/Apps/Velvet/VelvetStore.cs
```diff
@@ -1139,17 +1157,49 @@ public void Block(string userId, Action<bool> onComplete)
...
         if (Array.IndexOf(hidden, userId) < 0)
         {
             var grown = new string[hidden.Length + 1];
             Array.Copy(hidden, grown, hidden.Length);
             grown[hidden.Length] = userId;
             hiddenFromDiscover = grown;
+
+            var accountId = MyUserId;
+            work.Run("discover hidden save",
```
XeldarAlz: `StoreWork.Run` is an unordered `Task.Run`, so this save can execute before the initial disk load has merged. In that window the file gets overwritten with only the in-memory ids and previously persisted entries are lost. Serializing load-then-save closes the gap.

src/Aetherphone/Apps/Velvet/VelvetStore.cs
```diff
@@ -1197,6 +1247,56 @@ public void RefreshBlocked()
...
+        {
+            return;
+        }
+
+        var accountId = MyUserId;
+        loadingNotInterested = true;
+        var epoch = accountEpoch;
+        work.Run("notInterested", async token =>
+        {
+            if (!hiddenIdsLoaded)
```
XeldarAlz: This duplicates the disk load in `EnsureHiddenLoaded` but never checks or sets `loadingHiddenIds`, so both can run at once and race the read-modify-write on `hiddenFromDiscover`. Routing both paths through `EnsureHiddenLoaded` removes the duplication and the race.

src/Aetherphone/Apps/Velvet/VelvetStore.cs
```diff
@@ -1197,6 +1247,56 @@ public void RefreshBlocked()
...
+
+            if (epoch == accountEpoch)
+            {
+                notInterested = list.ToArray();
+            }
+        }, () =>
+        {
+            if (epoch == accountEpoch)
+            {
+                notInterestedLoaded = true;
```
XeldarAlz, line 1312: This cleanup runs in a finally block, so a failed `UserAsync` still marks the list as loaded and the page shows the empty state with no retry. Worth distinguishing a failed load from a genuinely empty list.

src/Aetherphone/Apps/Velvet/VelvetStore.cs
```diff
@@ -1197,6 +1247,56 @@ public void RefreshBlocked()
...
+                hiddenFromDiscover = MergeHidden(hiddenFromDiscover, stored);
+                discoverResults = WithoutHidden(discoverResults);
+                hiddenIdsLoaded = true;
+            }
+
+            var ids = hiddenFromDiscover;
+            var list = new List<VelvetProfileDto>(ids.Length);
+            for (var index = 0; index < ids.Length; index++)
+            {
+                var user = await client.UserAsync(ids[index], token).ConfigureAwait(false);
```
XeldarAlz, line 1294: One sequential `UserAsync` per hidden id, refetched on every invalidation. Fine for small lists; a parallel or batched fetch would be worth it if this can grow.

src/Aetherphone/Apps/Velvet/VelvetDiscovernotInterestedArchive.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Text;
+
+namespace Aetherphone.Apps.Velvet;
+
+internal sealed class StoredHiddenList
+{
+    [JsonProperty("ids")] public List<string> UserIds { get; set; } = new();
+}
+
+internal sealed class VelvetDiscovernotInterestedArchive
```
XeldarAlz: PascalCase break in the type name (`Discovernot`). `VelvetNotInterestedArchive` would match the standardized naming; the log messages below also still say `DiscoverHiddenArchive`.

src/Aetherphone/Apps/Velvet/VelvetStore.cs
```diff
@@ -92,16 +92,24 @@ internal sealed class VelvetStore : ChatThreadStoreBase<VelvetMessageDto, Velvet
     private volatile UserDto[] blocked = Array.Empty<UserDto>();
     private volatile bool loadingBlocked;
     private volatile bool blockedLoaded;
+    private volatile VelvetProfileDto[] notInterested = Array.Empty<VelvetProfileDto>();
+    private readonly VelvetDiscovernotInterestedArchive notInterestedArchive;
+    private volatile bool hiddenIdsLoaded;
+    private volatile bool loadingHiddenIds;
```
XeldarAlz: The hidden vocabulary survives here (`hiddenIdsLoaded`, `loadingHiddenIds`, the discover hidden work labels) next to the notInterested members. Worth finishing the rename in one direction.

src/Aetherphone/Apps/Velvet/VelvetShell.Me.cs
```diff
@@ -84,6 +84,20 @@ private void DrawSettings(Rect area)
 
             Gap(18f);
             VSectionHeader.Overline(Loc.T(L.Velvet.SafetyHeader));
+            var NotInterestedRow = new VRowModel
```
XeldarAlz: Locals are camelCase: `notInterestedRow`, matching `blockedRow` below.

src/Aetherphone/Apps/Velvet/VelvetShell.Me.cs
```diff
@@ -159,7 +174,65 @@ private void DrawBlocked(Rect area)
...
+                {
+                    Title = DisplayNameOf(user.DisplayName, user.Handle),
+                    Subtitle = SocialIdentity.ProfileMeta(user.Handle, RegionCodeOf(user)),
+                    Height = 60f,
+                    Leading = VRowLeading.Avatar,
+                    AvatarRadius = 20f,
+                    Name = DisplayNameOf(user.DisplayName, user.Handle),
+                    World = string.Empty,
+                    AvatarUrl = user.AvatarUrl,
+                    RoleBadges = user.Badges,
```
XeldarAlz, line 219: Add `RoleBadgeIds = user.BadgeIds` so dynamic badges render here like they do on the profile, discover, and blocked rows.

src/Aetherphone/Apps/Velvet/VelvetShell.Me.cs
```diff
@@ -98,7 +112,8 @@ private void DrawSettings(Rect area)
                 router.Push(VelvetView.Blocked);
             }
 
-            Gap(40f);
+
```
XeldarAlz: This replaced the trailing `Gap(40f)`, so the settings list loses its bottom padding. Restore it and drop the blank lines.

src/Aetherphone/Apps/Velvet/VelvetShell.cs
```diff
@@ -73,15 +73,19 @@ internal sealed partial class VelvetShell : IPhoneApp
     private float sinceHeartbeat = HeartbeatSeconds;
     private bool cachedLalafell;
 
+
```
XeldarAlz: Stray blank lines.

src/Aetherphone/Apps/Velvet/VelvetShell.Me.cs
```diff
@@ -159,7 +174,65 @@ private void DrawBlocked(Rect area)
                 }
             }
 
+            Gap(40f);
+        }
+    }
+    private void DrawNotInterested(Rect area)
```
XeldarAlz, line 180: Formatting parity with `DrawBlocked`: blank line before this method, after the header block, and after the empty-state block. The file also gains a trailing blank line at EOF.

src/Aetherphone/Core/Localization/L.cs
```diff
@@ -3619,7 +3621,7 @@ internal static class Time
 
     internal static class Plugin
     {
-        public static readonly LocString CommandHelp =new("plugin.commandHelp", "Toggle the Aetherphone. /phone run [shortcut] runs a shortcut, /phone market [item] opens the market board, /phone reset recenters the phone, /phone test sends a sample notification.");
+        public static readonly LocString CommandHelp = new("plugin.commandHelp", "Toggle the Aetherphone. /phone run [shortcut] runs a shortcut, /phone market [item] opens the market board, /phone reset recenters the phone, /phone test sends a sample notification.");
```
XeldarAlz, line 3624: Unrelated whitespace fix. Fine to keep, though it belongs in its own chore commit.

XeldarAlz:
Commit subjects should follow the conventional format used in this repo, for example `feat(velvet): persist the not interested list per account`. A squash merge with that subject also works.

src/Aetherphone/Apps/Velvet/VelvetStore.cs
Deldee: I might rework this part, I originally didnt want to save the profile data locally, because I thought it included a full image
Since the avatar is only the url, I might save those when adding to not interested, and simply use the file content to fill the list

We'd skip the async calls when loading the list, but risk having data be out of date (badges, avatars, names, handle would match what it was at the time of adding, instead of at the current time)

Thoughts?

XeldarAlz: Keep the ids only, do not snapshot the profile. A local copy of names, avatars and badges is data nothing purges when a user is removed or banned, and it would bring back the stale badge problem on the exact screen that renders badges. You still need the fetch anyway to know the account still exists.

The sequential loop is fine at this size, so leave it as is. The real fix is a server side list mirroring /blocks/ plus discover excluding hidden users, which I will add to the backend as a follow up. This PR does not need to wait for it.

XeldarAlz:
Pushed a locale fix to your branch. The nine JSONs had been reindented from two spaces to four, which rewrote every line and caused the conflicts, so I restored the formatting and kept only your two new keys. zh.json was also invalid JSON (an old unescaped quote pair that the reindent made worse), so Chinese was not loading at all. Pull before your next push.

Still open, all in VelvetStore:

- Make EnsureNotInterestedLoaded the only disk reader and have RefreshNotInterested await it, and have the saves await the same gate. That closes both the duplicate load and the save landing before the initial load truncates the file.
- Clear notInterested in OnAccountSwitched, otherwise the page shows the previous account until the refresh lands.
- Do not set notInterestedLoaded when the fetch failed, otherwise a failure renders as an empty list with no retry.

Answered your fetch question in the thread above: keep it as it is. After that I will squash with feat(velvet): persist the not interested list per account.

Deldee:
Tested in game, I think that should cover it all unless i missed something

PR #114, opened 2026-08-11
feat(Wallet)add notification toggle for capped currency

Deldee:
(Wallet)Add notification toggle for capped currency (default on) so people can disable the notification at will

## Checklist


src/Aetherphone/Apps/Wallet/WalletApp.cs
```diff
@@ -140,6 +152,29 @@ private void DrawHeader(Rect content, float scale)
             1.15f, FontWeight.SemiBold);
     }
 
+    private bool DrawNotificationToggle(Rect content, float scale)
```
XeldarAlz, line 155: This is a line-for-line copy of `DrawNotificationPauseButton` in `LinkpearlApp.Chats.cs`. Please extract a shared header bell component in `Windows/Components` and call it from both apps.

src/Aetherphone/Configuration.cs
```diff
@@ -52,6 +52,7 @@ internal sealed class Configuration : IPluginConfiguration, IHomeConfiguration,
     public bool NotifyWeeklyReset { get; set; }
     public bool NotifyGrandCompanyReset { get; set; }
     public bool NotifyRetainerVentures { get; set; }
+    public bool NotifyWalletCapped { get; set; } = true;
```
XeldarAlz: This flag only gates the home badge (nothing else consumes `CountCapped`), and the convention for that is `ShowDailiesBadge` right below. Rename to `ShowWalletBadge`.

src/Aetherphone/Apps/Wallet/WalletApp.cs
```diff
@@ -140,6 +152,29 @@ private void DrawHeader(Rect content, float scale)
...
+        var max = center + new Vector2(radius, radius);
+        var paused = !configuration.NotifyWalletCapped;
+        var hovered = UiInteract.Hover(min, max);
+        var color = paused ? AppPalettes.Wallet.Accent : hovered ? AppPalettes.Wallet.TitleInk : AppPalettes.Wallet.MutedInk;
+        ProgressRing.CenterIcon(ImGui.GetWindowDrawList(), center,
+            paused ? FontAwesomeIcon.BellSlash : FontAwesomeIcon.Bell, color, 15f * scale);
+        var toggleRect = new Rect(min, max);
+        UiAnchors.Report("wallet.notifications.toggle", toggleRect);
+        HoverTooltip.Show(toggleRect,
+            Loc.T(paused ? L.Messages.ResumeNotifications : L.Messages.PauseNotifications));
```
XeldarAlz: The tooltip says notifications, but only the badge is affected. Add `wallet.showBadge` / `wallet.hideBadge` LocStrings in L.cs plus all nine JSONs in the same commit; `dailies.showBadge` already carries the badge term for every language.

src/Aetherphone/Core/Apps/AppRegistry.cs
```diff
@@ -75,7 +75,7 @@ public static AppBundle BuildDefault(PhoneServices services)
         apps.Add(new NewsApp(services.News, services.Media, services.Http, services.GameData));
         apps.Add(new CollectionsApp(services.Collections, services.Lodestone, services.Media, services.Http, services.GameData));
         apps.Add(new MarketApp(services.Market, services.MarketIndex, services.MarketAlerts, services.MarketLauncher, services.GameData, services.Textures, services.Configuration));
-        apps.Add(new WalletApp(services.GameData, services.Textures));
+        apps.Add(new WalletApp(services.GameData, services.Textures,services.Configuration));
```
XeldarAlz: Missing space after the comma.

XeldarAlz (commented):
Builds clean in Release and the default-on flag keeps current behavior, nice. A few consistency points inline.

Deldee:
Tested in game with the new commit, all issues should be resolved

XeldarAlz:
Merged, thanks for this one. The shared `NotificationToggleButton` came out clean, and Linkpearl dropping 20 lines onto it is exactly what I was hoping for.

I pushed two small commits to your branch before merging:

- The nine locale JSONs came back reindented to four spaces, which turned two new keys into a diff of roughly 82,000 lines. They are back at two spaces now, so the PR reads as the 18 lines it actually adds. That one is on us, not you: `.editorconfig` lists `json` in two sections with conflicting `indent_size`, and the later four space rule wins, so your editor was doing what the repo told it. Fixing that on master is on my list.
- The de, es, fr, pt and zh strings used the profile badge word (Abzeichen, insignia, badge, distintivo, 徽章), which is the term Settings uses for the badges the team grants, name colors and effects included. This toggle controls the count on the Home icon, and `dailies.showBadge` already names that in every language, so I switched to those: Kennzeichen, globo, pastille, contador, 角标. en, ja, ru and tr already matched.

Neither is anything to worry about for next time, the first was our config and the second is the kind of thing you only catch with the glossary in front of you.

Also: drop your Discord username here and we will set you up with the developer role on the server.

Deldee:
Ah, I was wondering why those files were coming up so big, that explains it
username is "deldee." (Lana Kyraille) on discord, I can do an application if you'd prefer

PR #118, opened 2026-08-15
feat(music): add live DJ listings from Rolladeck

eggoless:
## What

Adds live Twitch DJ listings from the Rolladeck public API to the Music app's Live tab, and aligns the DJ and venue detail pages with the community radio station page style.

## Why

The Music app's Live tab previously only showed community radio stations. Rolladeck tracks FFXIV DJs who stream on Twitch and share their in-game venue location. Surfacing those listings alongside community radio gives users one place to find live music whether it is a streamed set or a local station.

## How to test

1. Open the phone with `/phone` and navigate to Music.
2. Tap the Live tab (broadcast tower icon). Confirm the tab header reads "Live", not "Community Radio".
3. If any Rolladeck DJs are currently streaming with a venue address, their cards appear below On Air stations. If none are live, no section heading appears for them.
4. Tap a live DJ card to open the detail page. Confirm the action row (teleport ghost button left, circular play button right) sits between the avatar block and the venue card. Confirm the venue card shows the address in small muted text above the venue name.
5. Open a community radio station detail page. Confirm the Watch button is now a circular play icon. Confirm Teleport, Visit, and Discord appear as ghost buttons in a single horizontal row.
6. In the Live tab, drag the genre chip rail left and right. Confirm the phone window does not scroll while dragging.
7. Let the live DJ count exceed 9 and confirm the badge dot on the Live tab shows "9+" centered inside the circle.

## Checklist


src/Aetherphone/Apps/Music/Rolladeck/RolladeckService.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        if (!fetching && (force || data == null || (DateTime.UtcNow - lastFetch).TotalSeconds >= LiveCacheSecs))
+        {
+            fetching = true;
+            _ = Task.Run(FetchLiveAsync);
+        }
+
+        if (!scheduleFetching && (force || scheduleData == null || (DateTime.UtcNow - scheduleLastFetch).TotalSeconds >= SchedCacheSecs))
+        {
+            scheduleFetching = true;
```
XeldarAlz: Blocking: Schedule, HasSchedule and ScheduleLoading are never read anywhere, so this hits apiV1Schedule every 5 minutes (plus a forced call on every app open) for nothing. Drop the schedule path and its models (ScheduleEntry, ScheduleResponse, ActiveEventEntry, EventLineupEntry) until a screen uses them.

src/Aetherphone/Core/Localization/L.cs
```diff
@@ -6010,4 +6013,50 @@ internal static class Failure
         public static readonly LocString FeedbackLength = new("failure.feedbackLength", "Keep your feedback to {0} characters.");
         public static readonly LocString FeedbackTooManyImages = new("failure.feedbackTooManyImages", "You can attach at most {0} images.");
     }
+
+    internal static class Rolladeck
```
XeldarAlz, line 6157: Blocking: only 17 of these 42 keys are referenced; the tab, schedule-filter and notify strings look left over from the larger design and add ~225 dead entries across the nine JSONs for translators to maintain. Trim to the keys the shipped screens use.

src/Aetherphone/Apps/Music/Rolladeck/RolladeckService.cs
```diff
@@ -0,0 +1,106 @@
...
+            if (data == null)
+            {
+                return 0;
+            }
+
+            var count = 0;
+            for (var index = 0; index < data.LiveDJs.Count; index++)
+            {
+                var dj = data.LiveDJs[index];
+                if (dj.VenueName != null || dj.FormattedAddress.Length > 0)
```
XeldarAlz: Blocking: BadgeCount runs this every frame and FormattedAddress interpolates a new string per DJ per call. Cache a HasLocation bool (or the formatted string) when the response lands so the badge path stays allocation-free.

src/Aetherphone/Apps/Music/Rolladeck/RolladeckModels.cs
```diff
@@ -0,0 +1,239 @@
...
+    [JsonPropertyName("youtube")]       public string? Youtube       { get; set; }
+    [JsonPropertyName("tiktok")]        public string? Tiktok        { get; set; }
+    [JsonPropertyName("website")]       public string? Website       { get; set; }
+    [JsonPropertyName("musicPlatform")] public string? MusicPlatform { get; set; }
+
+    public bool    CanTeleport   => !string.IsNullOrEmpty(LifestreamArg);
+    public string  ServerLabel   => Server ?? Datacenter ?? "";
+    public string? RolladeckUrl  => TwitchUsername != null ? $"https://xivrolladeck.com/{TwitchUsername}" : null;
+
+    public string  NormalizedName  => RolladeckText.Normalize(DjName);
```
XeldarAlz: Blocking: NormalizedName and NormalizedTitle re-run Normalize on every access, which allocates every frame in the card and marquee paths whenever the name uses styled Unicode (the common case for DJs). Normalize once when the response is stored.

src/Aetherphone/Apps/Music/Rolladeck/RolladeckText.cs
```diff
@@ -0,0 +1,81 @@
+namespace Aetherphone.Apps.Music.Rolladeck;
+
+// Maps Unicode mathematical alphanumeric symbols (U+1D400–U+1D7FF) back to plain ASCII
+// so ImGui can render DJ names and stream titles that use "bold" or "italic" Unicode fonts.
+// Supplementary-plane characters that don't map to ASCII (emoji etc.) are stripped.
+internal static class RolladeckText
```
XeldarAlz, line 6: Blocking: foreach, braceless single-line ifs and abbreviated names (sb, v, off, dc) here and in RolladeckModels fall outside the repo conventions (braces on every body, no abbreviations, for over foreach). The codepoint table comments are fine to keep; the flow-narration ones ("Fast path", the else note) should go.

src/Aetherphone/Apps/Music/MusicApp.LiveDjs.cs
```diff
@@ -0,0 +1,1223 @@
...
+                                       card.Min.Y + card.Height * 0.5f);
+
+        AvatarView.DrawRemote(drawList, avatarCenter, avatarRadius, theme,
+            dj.DjName, dj.Server ?? string.Empty, dj.AvatarUrl, images, lodestone, 0.85f, 44);
+
+        var textLeft = avatarCenter.X + avatarRadius + 8f * scale;
+        var topY = card.Min.Y + 9f * scale;
+        var maxTextWidth = card.Max.X - textLeft - 80f * scale;
+
+        Marquee.DrawLeft("music.djCard." + (dj.DjSlug ?? dj.DjName),
```
XeldarAlz: Non-blocking: the home-shelf rows precompute their marquee keys and viewer strings in EnsureDjList, but these cards concat the key and string.Format the viewers every frame per card. Extend the cached pattern here.

src/Aetherphone/Apps/Music/MusicApp.LiveDjs.cs
```diff
@@ -0,0 +1,1223 @@
...
+            {
+                DrawLiveDjsEmpty(scale);
+            }
+
+            return;
+        }
+
+        ImGui.Dummy(new Vector2(0f, 14f * scale));
+        var origin = ImGui.GetCursorScreenPos();
+        var width = ImGui.GetContentRegionAvail().X;
```
XeldarAlz: Non-blocking: sibling Music pages size with ScrollLayout.StableContentWidth(); GetContentRegionAvail here (and in the two detail pages) reintroduces the scrollbar-width feedback shake when content height straddles the viewport.

src/Aetherphone/Apps/Music/Rolladeck/RolladeckService.cs
```diff
@@ -0,0 +1,106 @@
...
+                bearer: null,
+                token:  default);
+
+            if (result != null)
+            {
+                data      = result;
+                lastFetch = DateTime.UtcNow;
+            }
+        }
+        catch { }
```
XeldarAlz: Non-blocking: the bare catch swallows every failure, so an API outage renders an empty shelf with no retry. Consider the CommunityShelfStatus pattern: track a failed flag and surface a retry row.

src/Aetherphone/Apps/Music/MusicApp.LiveDjs.cs
```diff
@@ -0,0 +1,1223 @@
...
+        if (!string.IsNullOrEmpty(datacenter) &&
+            string.Equals(DcToRegion(datacenter), playerRegion, StringComparison.OrdinalIgnoreCase))
+        {
+            return 2;
+        }
+
+        return 3;
+    }
+
+    private static string? DcToRegion(string? dc) => dc switch
```
XeldarAlz: Non-blocking: this duplicates the datacenter-to-region knowledge already in Core/Housing/HousingRegions.cs, and any DC added later silently sorts as the farthest tier. Prefer exposing region-for-datacenter from one place.

src/Aetherphone/Apps/Music/MusicApp.Community.cs
```diff
@@ -191,25 +191,28 @@ private void DrawLive(in PhoneContext context)
...
-            DrawLiveGroup(scale, LiveGroup.Resting, Loc.T(L.Music.AllStationsSection));
+            DrawLiveDjsSection(scale);
+            DrawShelfHeading(Loc.T(L.Music.CommunityRadio), scale);
+            if (stations.Length == 0)
+            {
+                DrawCommunityShelfStatus(scale);
+            }
+            else
+            {
+                DrawTagFilterRail(scale, stations);
```
XeldarAlz, line 211: Non-blocking: ApplyTagFilter still filters the On Air group, but the rail now renders below it, so tapping a genre changes content above the control the user is touching. Intended? If not, filter only the groups under this heading.

src/Aetherphone/Windows/Components/ChipRail.cs
```diff
@@ -99,7 +99,7 @@ private static Rect ReserveRow(float scale)
         var origin = ImGui.GetCursorScreenPos();
         var width = ImGui.GetContentRegionAvail().X;
         var height = RowHeight * scale;
-        ImGui.Dummy(new Vector2(width, height));
+        ImGui.InvisibleButton("##chipRailRow", new Vector2(width, height));
```
XeldarAlz, line 103: Nit: a fixed id means two rails drawn in the same window scope would collide. Derive it from the rail instance or wrap in PushID.

src/Aetherphone/Core/Localization/L.cs
```diff
@@ -2703,6 +2703,9 @@ internal static class Music
         public static readonly LocString NoPlaylistsYet = new("music.noPlaylistsYet", "No playlists yet");
         public static readonly LocString SongOne = new("music.songOne", "1 song");
         public static readonly LocString SongsMany = new("music.songsMany", "{0} songs");
+        public static readonly LocString LiveDjs          = new("music.live_djs",           "Live Twitch DJs");
```
XeldarAlz: Nit: snake_case key; the file is camelCase throughout (music.poweredByRolladeck two lines down).

XeldarAlz (changes requested):
Release build clean, 0 warnings; 29/29 RolladeckText tests pass. Tested in-game on my side. Findings inline: 5 blocking (dead fetch, dead locale keys, allocation paths, style pass), rest non-blocking.

src/Aetherphone/Apps/Music/Rolladeck/RolladeckService.cs
eggoless: All removed.

src/Aetherphone/Core/Localization/L.cs
eggoless: Removed 25 keys (the summary listed FilterAtVenue as dead, but it's actively used at MusicApp.LiveDjs.cs:399 in the chip rail filter labels).  Both snake_case keys are now "music.liveDjs" / "music.liveDjsEmpty" in L.cs and all 9 JSON files.

src/Aetherphone/Apps/Music/Rolladeck/RolladeckService.cs
eggoless: private int liveCountWithAddress field, populated in FetchLiveAsync after data arrives; public int LiveCountWithAddress => liveCountWithAddress; just reads it.

src/Aetherphone/Apps/Music/Rolladeck/RolladeckModels.cs
eggoless: private string? normalizedName/normalizedTitle backing fields; InitNormalized() sets them once; called on every entry after fetch. Properties now read the cached fields.

src/Aetherphone/Apps/Music/Rolladeck/RolladeckText.cs
eggoless: StyleBases and DigitBases loops now use for. foreach kept only on EnumerateRunes() (not indexable — correct). All abbreviations renamed. "Fast path" and inline // else: comments removed. All if/else bodies braced. Codepoint table comments (// Mathematical Bold, etc.) kept.

src/Aetherphone/Apps/Music/MusicApp.LiveDjs.cs
eggoless: Extended EnsureDjList to populate sortedDjCardKeys, sortedDjCardTitleKeys, sortedDjsAtVenueCardKeys, sortedDjsAtVenueCardTitleKeys, and sortedDjsAtVenueViewerCounts parallel to their respective lists. DrawLiveDjCard now receives cardKey, titleKey, and viewerCount as parameters.

src/Aetherphone/Apps/Music/MusicApp.LiveDjs.cs
eggoless: Fixed in DrawLiveDjsSection — using ScrollLayout.StableContentWidth() now.

src/Aetherphone/Apps/Music/Rolladeck/RolladeckService.cs
eggoless: Added private bool fetchFailed and public bool Failed => fetchFailed. The catch block sets fetchFailed = true; the success path sets it back to false.

src/Aetherphone/Apps/Music/MusicApp.LiveDjs.cs
eggoless: Removed DcToRegion and playerRegion entirely. ProximityTier now calls HousingRegions.For() on both sides (datacenter and playerDc), with a guard against HousingRegions.Unknown to avoid false positives when the DC isn't recognized.

src/Aetherphone/Apps/Music/MusicApp.Community.cs
eggoless: Not intentional. Added GroupOf(stations[index]) == LiveGroup.OnAir as a short-circuit in ApplyTagFilter so On Air stations always pass the filter regardless of the selected tag.

src/Aetherphone/Windows/Components/ChipRail.cs
eggoless: ReserveRow now takes the ChipRail instance as a parameter and wraps the InvisibleButton in PushID(RuntimeHelpers.GetHashCode(rail)) / PopID.

src/Aetherphone/Core/Localization/L.cs
eggoless: Fixed as part of the locale key pass -- both music.live_djs and music.live_djs_empty renamed to camelCase in L.cs and all nine JSON files.

eggoless:
Fixes put in and rebased localization after Rising Stones CN/Allagan case update. 
Requested requested re-review.

XeldarAlz:
Hey @eggoless, first off: this came out great. The proximity sort, the name normalization, and the venue pages are really nice work.

One thought on the play button, open to discussion: we already have AetherStream, so instead of taking users out of the game to a browser, what do you think about playing these streams right on the phone? mpv with yt-dlp as its link resolver already plays a raw `https://twitch.tv/<user>` URL as-is, live HLS included, and the player UI already handles live media (LIVE pill, seeking disabled when there is no duration). So it looks like most of the work is already done for us.

If you like the idea, the two play circles (DJ detail sheet and the venue "LIVE NOW" card) could hand the stream to AetherStream instead of `Util.OpenLink`:

```csharp
var entry = new VideoQueueEntry(dj.TwitchUrl, dj.NormalizedTitle ?? dj.NormalizedName,
    "Rolladeck", null, dj.AvatarUrl);
videoQueue.PlayNow(entry);
navigation.Open("aetherstream");
```

Some pointers in case they help:

- `AetherStreamQueue.PlayNow` lives in `src/Aetherphone/Core/Video/AetherStreamQueue.cs`. `AppRegistry.BuildDefault` already receives the queue, so it just needs passing into the `MusicApp` ctor.
- Cross-app navigation is `INavigator.Open(appId)`; apps cache `context.Navigation` during `Draw` (see `AppStoreApp` for the pattern).
- If the media engine is not installed yet, nothing special is needed: AetherStream's setup gate takes over when the app opens, and the queued stream starts once the download finishes.
- For the venue card, `venue.DjTwitch` + `venue.LogoUrl` map the same way.
- `L.Rolladeck.Watch` ("Watch") is already declared and currently unused, so the label is free. Any new strings need the usual nine-JSON lockstep.
- It would be nice to keep the Rolladeck page reachable as a secondary link (the socials grid and the footer already cover that).
- One heads-up for testing: Twitch splices ad breaks into anonymous HLS sessions, so an occasional "commercial break in progress" segment is expected, not a bug.

Would you like to take this on, or should I handle it? Both work: if you are up for it, go ahead, and if you would rather not, I can add it to your branch myself on top of your commits.

eggoless:
Hiya @XeldarAlz . Unless I am mistaken, the AetherStream/Mogcast would not count as a viewer for the twitch streamer. That is incredibly problematic. If that is accurate, I couldn't support that approach and I wouldn't further that discussion. One of the goals of Rolladeck is to bring visibility to the content creators and effectively scraping their content would go against that. Though would be happy to be proven otherwise if Aetherstream can be proven to be supporting them appropriately.

The second caveat of using it is that it is removng all engagement (chat, follow, subscriptions) that normally go with the viewing experience. Rolladeck includes that (from twitch) as well as the additional information about both the DJs and Venues they are out - something AetherStream strips away. 

I can review the pointers section once I return later this evening, but the first item I mentioned it a hard stop for me.

XeldarAlz:
Hey @eggoless, AetherStream viewers do count toward the streamer's Twitch viewer count. So I have to push back on the first point, because the premise isn't accurate.

Here's what actually happens under the hood. The player requests an anonymous playback access token from Twitch's own GQL API, the same endpoint the website uses, and exchanges it at usher.ttvnw.net for the channel's HLS master playlist. From that moment on it holds a live session against Twitch's video CDN, re-fetching the live media playlist every couple of seconds for the entire watch, exactly like any other player. Twitch's concurrent viewer count is derived from these active video delivery sessions, not from logged-in web players. Streamlink works through the identical mechanism and has been counted for years.

The reason we want playback on the phone in the first place is immersion. Sending someone out of the game to a browser or a second monitor breaks the experience the venue scene is built on. The whole point of Aetherphone is that everything happens in the game world: you're standing at the venue, the DJ is playing, and you watch the stream on the phone in your hand without ever leaving that moment.

You're right about the engagement gap, and the plan is to close it rather than accept it, with Twitch chat built directly into the viewing surface. Sign-in is a one-time OAuth device flow: the phone shows a short code, the viewer confirms it at twitch.tv/activate, and from then on the plugin connects to Twitch's chat servers with the viewer's own account. No token pasting and no relay through our servers; the connection is theirs, the messages are theirs.

On screen it works like the rest of the phone: in the theater view the stream gets a live chat pane next to the video (below it in portrait), rendered with the same chat components as the phone's messaging apps, with Twitch name colors, sub and mod badges, and a composer at the bottom to send. The channel name and avatar stay pinned above the chat with follow and subscribe deep links. And because the join is authenticated, the viewer shows up in your chatters list under their own name, exactly as if they had the site open. Open on Twitch remains one tap away for the full experience (minutes watched and Drops only accrue in the web player).

eggoless:
Thanks for the active discussion on this @XeldarAlz . I want to make sure we come to the right decision that benefits both Aetherphone users, but also the content creators.

To address the engagement gap first: I agree that it makes sense to have it on the phone/in-game if we're not sacrificing in other areas. If viewers aren't counted for the content creator (for example) or if the Aetherphone user cannot follow when a DJ raids into another one, that delivers a worse user experience and creator experience respectively. Those are both challenges that would need to be overcome before we offer that as a service.

Locking into the viewer count, I did more research on whether it counts and I don't believe the current implementation you propose counts as viewers for the Twitch streamers.

Twitch moved away from counting raw HLS CDN sessions back in 2022/2023 when they added two requirements that a yt-dlp/mpv stack cannot satisfy: JavaScript-based tracking heartbeats and client-integrity (CI) tokens. Both are browser-side signals. The community has confirmed this directly: the Streamlink-twitch-gui project has a thread where users verified that watching via mpv through Streamlink does not credit channel points, does not update watch time stats, and does not count raids (https://github.com/streamlink/streamlink-twitch-gui/discussions/996). The Streamlink plugin docs also document the CI token requirement and the lengths needed just to get stream access (https://streamlink.github.io/cli/plugins/twitch.html).

Looking at the AetherStream implementation itself, the chain is: the Twitch URL goes to mpv, mpv calls yt-dlp to resolve it to an HLS playlist (configured to Aetherstream's mpv options), and mpv fetches segments directly from the CDN. There is no Twitch GQL heartbeat, no tracking pixel, and no client-integrity token sent. The heartbeat in WatchAlongSession is Aetherphone's own watch-along sync to Aethernet, nothing to do with Twitch. So as implemented today, a viewer watching a DJ through AetherStream would not register on Twitch's side at all.

A similar FFXIV plugin, Aetherwave, attempted this approach and is instructive here. Their README claims Streamlink "uses your existing Twitch credentials automatically," but the Streamlink documentation directly contradicts this: credentials must be manually extracted from the browser and passed explicitly via a command-line flag. Aetherwave relies on a separate Twitch XIV OAuth token to maintain viewer status through chat interaction, which at best puts them in an unconfirmed "maybe" on viewer counting. AetherStream has neither piece: no credentials into yt-dlp, no OAuth, no chat connection. It is a fully anonymous HLS pull with nothing on Twitch's side to tie it to a real account, which moves it from "maybe" to a definitive no.

Given all of that, I'd propose we keep link-out as the default play action for this PR, which is what is currently implemented. Though this breaks the complete immersion of Aetherphone, that keeps the creator experience safe. As a follow-up, if and when the in-app chat with OAuth is built out and we can confirm a method which meets the requirements to count viewers, it becomes much easier to recommend and we could revisit.

PR #124, opened 2026-08-16
refactor(camera): changed logic for hiding ui on shoot

Radvo:
## What

Changed the logic for hiding UI on camera shoot, now with a toggle glyph inside the camera UI. Refactor also touches on camera UI glyphs because of that addition. 
Because of how Dalamud handles plugin interface on game hiding UI, we toggle hiding plugin interface on hiding game ui off and back on during a camera shoot to preserve the phone on user screen.

## Why

Discord bug report https://discord.com/channels/1527213298087493732/1535732036390617399

Previous logic used AtkUnitManager.ToggleUI() which did not hide dialog addons and forced the game to back out of menus one step, which I suspect is the reason user was in an unrecoverable game state after having backed out of a menu that (maybe) shouldn't allow it.

Closes #

## How to test

Open camera, attempt to take photo with and without new UI toggle on various UI elements.
Check dalamud settings -> look & feel -> "Hide Plugin UI when the game UI is toggled off" is unaffected by using the camera.
Ideally would try to reproduce user bug report with specific housing refund dialog popup.

(Only tested on windows)

## Checklist


src/Aetherphone/Configuration.cs
```diff
@@ -99,6 +99,7 @@ internal sealed class Configuration : IPluginConfiguration, IHomeConfiguration,
     public bool CameraLandscape { get; set; }
     public bool CameraGrid { get; set; }
     public bool CameraFlash { get; set; } = true;
+    public bool CameraHideUi { get; set; }
```
XeldarAlz: Blocking: this defaults to false, so the camera stops hiding the HUD, nameplates and chat that it has always stripped, and every existing user's next photo silently gains UI. Ship it as `= true` to preserve current behaviour; the chip then reads as "turn it off".

src/Aetherphone/Apps/Camera/CameraChrome.cs
```diff
@@ -19,64 +20,104 @@ internal static class CameraChrome
...
-        var rowCenterY = barMax.Y + Metrics.Space.Lg * 3f * scale;
-        if (FlashChip(new Vector2(screen.Min.X + 34f * scale, rowCenterY), flashEnabled, scale))
+        var rowCenterY = barMax.Y - Metrics.Space.Xs * scale - ChipHalfExtent * scale;
+        if (FlashChip(new Vector2(screen.Min.X + ChipEdgeInset * scale, rowCenterY), flashEnabled, scale))
         {
             return CameraBarAction.ToggleFlash;
         }
 
-        return RotateChip(new Vector2(screen.Max.X - 34f * scale, rowCenterY), landscapeEnabled, scale)
+        if (HideUiChip(new Vector2(screen.Center.X, rowCenterY), hideUiEnabled, scale))
```
XeldarAlz: Non-blocking: the returns here fire on the mouse-down frame, so clicking flash skips drawing both later chips (and clicking this one skips rotate) for a frame. Evaluate all three, hold the action in a local, and return it after the last draw. Same in `SideBar`.

src/Aetherphone/Apps/Camera/CameraChrome.cs
```diff
@@ -19,64 +20,104 @@ internal static class CameraChrome
...
-        var hovered = UiInteract.Hover(min, max);
-        var rounding = Metrics.Radius.Field * scale;
-        var fill = flashEnabled
+        UiAnchors.Report("camera.flash", ChipRect(center, scale));
+        var clicked = ChipSurface(center, flashEnabled, scale);
+        DrawBolt(ImGui.GetWindowDrawList(), center, ChipGlyphExtent * scale, ChipIconColor(flashEnabled));
+        return clicked;
+    }
+
+    private static bool HideUiChip(Vector2 center, bool hideUiEnabled, float scale)
```
XeldarAlz: Nit: `FlashChip` reports a `camera.flash` anchor that the onboarding tour points at (TourRegistry.cs:95). One `UiAnchors.Report("camera.hideUi", ChipRect(center, scale))` here lets a tour step land on the new control later.

src/Aetherphone/Apps/Camera/CameraChrome.cs
```diff
@@ -11,6 +11,7 @@ internal enum CameraBarAction
 {
     None,
     ToggleFlash,
+    ToggleHideUi,
```
XeldarAlz: Nit: this adds a user-facing toggle, a config key and a new service, so the subject fits `feat(camera):` better than `refactor`. announce-commits.yml posts the subject line to Discord verbatim, so the type is public.

XeldarAlz (changes requested):
Build `-c Release` is clean (0 warnings) and 1349/1349 tests pass. Not verified in-game on my side, so the Dalamud `DisableUserUiHide` interplay rests on your testing.

Swapping `ToggleUi` for `RaptureAtkModule.SetUiVisibility` is the right diagnosis and `GameUiVisibility` is a clean service. One blocking item on the default value, the rest is polish.

src/Aetherphone/Configuration.cs
XeldarAlz: Fair push, and you are right that a lit "hide" chip is backwards. I would flip the toggle instead of the default: make it "keep UI", not "hide UI".

- unlit by default, so no mystery enabled button
- unlit still means the UI is hidden, which is what the camera has always done, so nobody's existing shots silently gain a HUD
- lit = effect is active, same grammar as the flash and rotate chips, no per-button exception
- `DrawHudGlyph` already handles it: pass `keepUi` for both the fill and `uiPresent`, so lit shows the populated frame and unlit shows the empty one

Mouseover hint yes: `HoverTooltip.Show(ChipRect(center, scale), Loc.T(...), HoverLabelSide.Below)`. `UiInteract.Hover` is a pure predicate, so the extra call inside `Show` is safe next to the one in `ChipSurface`. Needs one LocString in `L.Camera` (L.cs:2920) plus the key in all nine JSONs.

I would skip the "UI" text though: it is user-visible so it needs a LocString, and translations of it will not fit a 32px chip.

Radvo:
changed logic to 'show' instead of 'hide' to preserve old behavior
added mouseover hint + localization (claude opus 5)
added tour step for new showUI chip + localization (claude opus 5)
fixed chips not drawing for 1 frame
changed commits to feat() instead of refactor()

src/Aetherphone/Configuration.cs
XeldarAlz: Resolved. CameraShowUi defaults to false, so unlit still strips the HUD and nobody's existing shots change. Lit now means the effect is active, same grammar as flash and rotate. Tooltip plus the nine JSONs are in.

src/Aetherphone/Apps/Camera/CameraChrome.cs
XeldarAlz: Resolved. TopBar and SideBar both draw all three chips now and return the action from a local after the last draw, so clicking one no longer skips a later chip for a frame.

src/Aetherphone/Apps/Camera/CameraChrome.cs
XeldarAlz: Resolved. ShowUiChip reports `camera.showUi`, and the camera tour gained a step on it with the tour version bumped 2 to 3 so existing users see it.

src/Aetherphone/Apps/Camera/CameraChrome.cs
XeldarAlz: Half done. Both commits are `feat(camera):` now, but the PR title still reads `refactor(camera):`. On a squash merge the title is what announce-commits.yml posts, so worth retitling before merge.

XeldarAlz (commented):
Build and Guards pass on 17723de0, and all four review threads are addressed. One thing left: TourRegistry.cs conflicts with dev, which since split it into partials. The camera tour lives in TourRegistry.Media.cs:23 now, so the new step needs to move there. Every other file automerges clean.

Radvo:
addressed merge conflict
updated failing unit test with new assert values

XeldarAlz (approved):
All feedback addressed: default preserves the old hidden-UI behavior, chips no longer skip a frame on click, tour anchor and step are in, commits are feat(camera). CI green. Thanks for the thorough rework.

PR #127, opened 2026-08-18
fix(mogcast):added more robust way to render the ingame screen when DLSS or FSR is active

Farroness:
## What

Added a more robust check and implementation of the in-game screen for mogcast that works better with DLSS/FSR upscaling. We use HLSL to do draw a temp texture at the depth buffer's size with depth testing active, then draw a full-screen triangle that copies the temp texture into the real back buffer at it's larger size using alpha blending.

## Why

Currently if your internal resolution and output resolution do not match the in-game screen will not appear.

Closes #

## How I tested it

1. Opened mogcast on phone and played a video.
2. Changed DLSS and FSR settings at random to ensure the screen continued to show.

## Screenshots

Reach out for screenshots.

## Checklist

Strings:

Quality:

### DIFF
--- a/src/Aetherphone/Core/Video/ScreenPainter.cs
+++ b/src/Aetherphone/Core/Video/ScreenPainter.cs
@@ -30,6 +30,16 @@ internal sealed unsafe class ScreenPainter : IDisposable
 	private readonly DepthStencilState _depthState;
 	private readonly Buffer _cbuf;
 
+	private readonly VertexShader _blitVs;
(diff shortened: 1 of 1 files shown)


Farroness:
Just a heads up, this fix seems to work based off my testing, but we likely need a few more people to check it to make sure it's working as expected. I'm not sure if the implementation is the best, I had to rely on a lot of help for the HLSL portion.

Farroness:
Closing as this doesn't work at all with windowed mode.

PR #129, opened 2026-08-20
fix(shell): close the control center when an app turns landscape

Farroness:
## What

Updated shelloverlay with a check for landscape mode and added a control center dismiss to the mogcast player.

## Why

This prevents users from getting stuck in the control center with no way to exit besides reloading the phone.

Closes #

## How I tested it

1. Started a video in mogcast.
2. Entered fullscreen mode.
3. Attempted to open the control center.

## Screenshots

Ping in discord if needed.

## Backend coordination

N/A

## Checklist

Strings:

Quality:


src/Aetherphone/Core/Shell/ShellOverlayCoordinator.cs
```diff
@@ -191,10 +191,15 @@ public void DrawOverlays(in ChassisGeometry chassis, PhoneTheme theme, float del
             controlCenter.Dismiss();
         }
 
+        if (AppLandscape.Held(navigation.Current?.Id ?? string.Empty) && controlCenter.IsActive)
```
XeldarAlz: Non-blocking: worth naming this in the PR body, it is the non-obvious half of the fix. `ControlCenter.cs:624` evaluates `drag.Released` outside the `if (gesturesEnabled)` block, so a top-band drag begun before the gate flipped still calls `Open()`. Without this dismiss you get a panel the gate then refuses to close by gesture.

src/Aetherphone/Core/Shell/ShellOverlayCoordinator.cs
```diff
@@ -191,10 +191,15 @@ public void DrawOverlays(in ChassisGeometry chassis, PhoneTheme theme, float del
...
+        {
+            controlCenter.Dismiss();
+        }
+
         HandleEscape();
         controlCenter.Draw(screen, theme, delta,
             !navigation.IsTransitioning && !director.CapturesPointer && !state.IslandCaptures &&
-            !banOverlay.IsActive && navigation.Current?.Id != "camera",
+            !banOverlay.IsActive && navigation.Current?.Id != "camera" && !AppLandscape.Held(navigation.Current?.Id ?? string.Empty),
```
XeldarAlz: Non-blocking: the same expression is written and evaluated twice in one draw method. Hoist it, and use the idiom already at `PhoneShell.cs:175`. The `?? string.Empty` is dead weight: `Held` short-circuits on `holder.Length > 0`, and this project has no `<Nullable>enable</Nullable>`. Builds clean.

```suggestion
        var landscapeHeld = navigation.Current is { } landscapeApp && AppLandscape.Held(landscapeApp.Id);
        if (landscapeHeld && controlCenter.IsActive)
        {
            controlCenter.Dismiss();
        }

        HandleEscape();
        controlCenter.Draw(screen, theme, delta,
            !navigation.IsTransitioning && !director.CapturesPointer && !state.IslandCaptures &&
            !banOverlay.IsActive && navigation.Current?.Id != "camera" && !landscapeHeld,
```

XeldarAlz: Non-blocking: docs/app-framework.md:338 still says the Control Center is "opened with a tap on the band at the top of the screen" with no landscape caveat, so the docs checkbox is not quite true yet. One clause there, or at docs/architecture.md:129 where AppLandscape is described, closes it.

XeldarAlz (changes requested):
Release build clean (0 warnings) and 1366/1366 tests pass on 1c5eef3. Not verified in-game on my side. The logic is right and covers both halves of the bug, findings are inline.

Blocking: the commit subject breaks docs/conventions.md:164 (lowercase fragment, no trailing period, user-visible outcome rather than the diff). Suggest `fix(shell): close the control center when an app turns landscape`.

src/Aetherphone/Core/Shell/ShellOverlayCoordinator.cs
Farroness: Fixed

Farroness: Fixed

PR #132, opened 2026-08-22
feat(housing): add China region housing data source

NiGuangOwO:
## Summary

The international housing API (housing-api.yozoracho.dev) does not cover China region worlds, so the Housing app failed with `404 unknown_world` for any CN world and kept retrying.

## Changes

- **CN data source**: route China worlds to the `house.ffxiv.cyou` `/sales` endpoint, which returns every available plot for a server in a single request (no per-ward polling).
- **CN world list**: ship the static CN roster (4 data centers, 28 worlds) so CN worlds appear in the world picker under a new "China" region; map each CN district (TerritoryType) to its API `area` id.
- **Payload mapping**: CN sales entries map onto the existing `HousingPlot` model (size, price, first/last seen). Lottery fields stay `Unknown` because the CN source does not expose purchase/lottery data.
- **Diagnostics**: provider name and API endpoint in the Housing settings diagnostics now reflect the CN source when a CN world is selected.
- Ward mapping: the CN /sales payload uses 0-based Slot values; Slot + 1 maps to the existing 1-based ward numbering (same convention as the international API's ward_number).

## Testing

Verified against world 1076 (Platinum Phantom, CN): `/sales` returns all open plots across the 5 districts in a single request; build passes with 0 warnings / 0 errors.


src/Aetherphone/Core/Housing/HousingRestProvider.cs
```diff
@@ -36,34 +36,33 @@ public HousingRestProvider(HttpService http, HousingProviderKind kind, string di
...
-                var dataCenter = entry.DataCenterName ?? string.Empty;
-                worlds.Add(new HousingWorld(entry.Id, entry.Name, entry.DataCenterId, dataCenter,
-                    HousingRegions.For(dataCenter)));
             }
-
-            return worlds;
         }
+
+        CnHousing.AppendWorlds(worlds);
+        return worlds.Count == 0 ? null : worlds;
```
XeldarAlz: Blocking: `AppendWorlds` runs even when the request failed, so this can no longer return null. `HousingService.LoadWorldsAsync:508` reads the 28 CN worlds as success, overwrites the primed list and writes it to `worlds.json` (30 day TTL). One transient failure leaves a Global player CN only for the session.

src/Aetherphone/Core/Housing/HousingRestProvider.cs
```diff
@@ -36,34 +36,33 @@ public HousingRestProvider(HttpService http, HousingProviderKind kind, string di
...
-
-                var dataCenter = entry.DataCenterName ?? string.Empty;
-                worlds.Add(new HousingWorld(entry.Id, entry.Name, entry.DataCenterId, dataCenter,
-                    HousingRegions.For(dataCenter)));
             }
-
-            return worlds;
         }
+
+        CnHousing.AppendWorlds(worlds);
```
XeldarAlz: Non-blocking: this is one of two places that build the world list. `HousingCache.ReadWorlds` (HousingCache.cs:22) builds the same list from `worlds.json` and does not append, so a warm cache start shows no CN worlds until the network round trip lands. Merging the roster once in `HousingService` fixes both.

src/Aetherphone/Core/Housing/HousingRestProvider.cs
```diff
@@ -232,3 +315,71 @@ internal static class HousingPlotOrder
         return ward != 0 ? ward : first.Key.Plot.CompareTo(second.Key.Plot);
     };
 }
+
+internal static class CnHousing
+{
+    private static readonly (string DataCenter, (uint Id, string Name)[] Worlds)[] DataCenters =
```
XeldarAlz: Blocking: the roster is ungated, so 28 CN worlds and 4 CJK data center names land in every Global player's picker for worlds they cannot visit. The CJK names also burn about 90 of the 2000 glyph `LearnedGlyphCap` (FontService.cs:57) and force an atlas rebuild on first open. Gate on `gameData.IsChineseGameClient()`, precedent at AppAvailability.cs:41.

src/Aetherphone/Core/Housing/HousingRestProvider.cs
```diff
@@ -232,3 +315,71 @@ internal static class HousingPlotOrder
...
+
+    public static void AppendWorlds(List<HousingWorld> worlds)
+    {
+        for (var dcIndex = 0; dcIndex < DataCenters.Length; dcIndex++)
+        {
+            var (dataCenter, dcWorlds) = DataCenters[dcIndex];
+            for (var index = 0; index < dcWorlds.Length; index++)
+            {
+                var (id, name) = dcWorlds[index];
+                worlds.Add(new HousingWorld(id, name, dcIndex, dataCenter, HousingRegions.For(dataCenter)));
```
XeldarAlz: Non-blocking: `dcIndex` invents data center ids 0..3 that collide with the real `DataCenter` RowIds the global API returns and with the 0 = unknown sentinel used elsewhere (MusterCard.cs:73). Deriving the roster from the `World` sheet where region is 5 (GameData.cs:213 already maps it) gives real ids and names, and will not drift when SE adds a world.

src/Aetherphone/Core/Housing/HousingRestProvider.cs
```diff
@@ -164,6 +219,34 @@ private static bool TryMapPlot(uint worldId, uint districtId, PaissaOpenPlot ent
         return true;
     }
 
+    private static bool TryMapCnPlot(uint worldId, uint districtId, CnSalesPlot entry, out HousingPlot plot)
+    {
+        plot = null!;
+        var ward = entry.Slot + 1;
+        var plotNumber = entry.Id;
```
XeldarAlz, line 244: Non-blocking: worth a line in the PR body saying the asymmetry is intentional. Verified live: `Slot` is 0..29 so the `+1` above is right, and `ID` is already 1..60 so it must not be offset. Next to `NormalizePlot` it reads like an off by one, and someone will eventually fix it.

src/Aetherphone/Core/Housing/HousingRestProvider.cs
```diff
@@ -232,3 +315,71 @@ internal static class HousingPlotOrder
...
+            (1177u, "海猫茶屋"), (1178u, "柔风海湾"), (1179u, "琥珀原"),
+        ]),
+        ("豆豆柴",
+        [
+            (1192u, "水晶塔"), (1183u, "银泪湖"), (1180u, "太阳海岸"), (1186u, "伊修加德"),
+            (1201u, "红茶川"),
+        ]),
+    ];
+
+    private static readonly HashSet<uint> Ids = BuildIds();
```
XeldarAlz: Nit: `FrozenSet<uint>` here, matching the `FrozenDictionary` in HousingRegions.cs, since `IsChinaWorld` is reached from draw code via `HousingService.ProviderName`. `BuildIds` also uses two foreach loops, and `Cn` is an abbreviation where the repo spells it out (`IsChineseGameClient`, `UnavailableInChina`).

src/Aetherphone/Core/Housing/PaissaModels.cs
```diff
@@ -97,13 +97,33 @@ internal sealed class HousingWorldCacheEntry
...
+{
+    [JsonPropertyName("Server")] public uint Server { get; set; }
+    [JsonPropertyName("Area")] public int Area { get; set; }
+    [JsonPropertyName("Slot")] public int Slot { get; set; }
+    [JsonPropertyName("ID")] public int Id { get; set; }
+    [JsonPropertyName("Price")] public long Price { get; set; }
+    [JsonPropertyName("Size")] public int Size { get; set; }
+    [JsonPropertyName("FirstSeen")] public long FirstSeen { get; set; }
+    [JsonPropertyName("LastSeen")] public long LastSeen { get; set; }
+    [JsonPropertyName("State")] public int State { get; set; }
```
XeldarAlz, line 110: Non-blocking: `/sales?server=1076` returns `State` in {0,1,2} with `EndTime` nonzero on exactly the State 1 and 2 rows, so the phase data is there. Leaving Phase Unknown sends every CN plot to `phaseOther` (HousingFilterState.cs:256) and leaves watchlist reminders inert (HousingReminderService.cs:42). Map State and EndTime, drop the rest.

src/Aetherphone/Core/Housing/PaissaModels.cs
```diff
@@ -97,13 +97,33 @@ internal sealed class HousingWorldCacheEntry
...
+    [JsonPropertyName("Participate")] public int Participate { get; set; }
+    [JsonPropertyName("Winner")] public int Winner { get; set; }
+    [JsonPropertyName("EndTime")] public long EndTime { get; set; }
+    [JsonPropertyName("UpdateTime")] public long UpdateTime { get; set; }
+    [JsonPropertyName("PurchaseType")] public int PurchaseType { get; set; }
+    [JsonPropertyName("RegionType")] public int RegionType { get; set; }
```
XeldarAlz, line 116: Non-blocking: `RegionType` comes back 0/1/2 split 148/200/2617 across 2965 rows, a distribution that looks like FC only, individual only and unrestricted rather than noise. Worth confirming with upstream before hardcoding `Eligibility` to Unknown.

src/Aetherphone/Core/Housing/HousingService.cs
```diff
@@ -90,9 +90,11 @@ public HousingService(HttpService http, Configuration configuration, GameData ga
 
     public int LastStatusCode => api.LastStatusCode;
 
-    public string ProviderName => api.DisplayName;
+    public string ProviderName =>
+        CnHousing.IsChinaWorld(WorldId) ? HousingEndpoints.CnDisplayName : api.DisplayName;
```
XeldarAlz: Non-blocking: `HousingRestProvider` already takes (kind, displayName, baseUrl), so it was built to be instantiated per source. Branching inside the provider and again here puts one decision in two places. A second instance picked per world leaves these properties untouched and gives the CN source its own `HousingProviderKind`.

src/Aetherphone/Core/Housing/HousingRegions.cs
```diff
@@ -20,11 +20,15 @@ internal static class HousingRegions
         ["Mana"] = "Japan",
         ["Meteor"] = "Japan",
         ["Materia"] = "Oceania",
+        ["陆行鸟"] = "China",
```
XeldarAlz: Non-blocking: this says "China" while `GameData.RegionNameFromId` (GameData.cs:213) says the Chinese name for the same region 5, and this table is already a hardcoded copy of sheet data that diverged once on "North America" versus "North-America". One vocabulary would be better.

src/Aetherphone/Core/Housing/HousingEndpoints.cs
```diff
@@ -7,10 +7,16 @@ internal static class HousingEndpoints
     public const string BaseUrl = "https://housing-api.yozoracho.dev/api/v1";
     public const string DisplayName = "Aetherphone housing service";
 
+    public const string CnBaseUrl = "https://house.ffxiv.cyou/api";
```
XeldarAlz: Nit: docs/architecture.md:273 still says housing data comes from PaissaDB. A new region in the world picker is user visible, so README and the in-app changelog want a line too.

XeldarAlz (changes requested):
Verified on 3343e63: `dotnet build Aetherphone.sln -c Release` gives 0 warnings / 0 errors. Not run in game. Wire mapping checked live against `/api/sales?server=1076` (2965 rows) and it is correct. Findings inline: 2 blocking, 7 non-blocking, 2 nits.

NiGuangOwO:
Implemented the requested review changes and pushed the updates.

XeldarAlz:
Re-checked at `e03c9ab9`. Build clean, 9 of 11 addressed.

Anchor verified independently: 156 rows with `EndTime > 0` across 14 CN servers, all on the `2022-08-08 15:00Z` / 9 day grid. It is correct. Add a line naming its source.

**Blocking**, `HousingService.cs:670`: `WorldsForClient` needs a non-empty global list to substitute for, so the CN roster only appears if the international API succeeds. Cold cache plus unreachable `housing-api.yozoracho.dev`, the case this PR exists for, gives an empty picker. Short circuit before `api.GetWorldsAsync`; the roster is local sheet data.

Need from you, since I have no CN client: a screenshot of the picker showing all 28 worlds, and where `world.RowId != 1200` comes from (`GameData.cs:224`). On `RegionType` I withdraw my earlier doubt, a 30 world global sample puts individual only at 46 percent.

**Non-blocking**: do not cache `IsChineseGameClient()` in a ctor that runs off the framework thread, it returns false there by design (`GameData.cs:239`) and every other caller resolves it at use. The `EndTime` branch in `InferChinaLotteryPhase` duplicates the anchor grid. Still open: the `Slot + 1` note in the PR body, README, `ChangelogData.cs`.

NiGuangOwO:
Hi, this is the result on the Chinese server client. world.RowId == 1200 (亚马乌罗提) is a server that exists in the Chinese server client data but is not officially open. It is located on 豆豆柴. You can browse all the servers on our Chinese official website: https://ff.web.sdo.com/web8/index.html#/servers

I have completed the changes as required. Please let me know if you still have any questions.

QQ20260823-014756
QQ20260823-014825
QQ20260823-014842

XeldarAlz (approved):
Verified at `ec25721`: `dotnet build Aetherphone.sln -c Release` gives 0 warnings / 0 errors, full suite 1421 passed. Everything from the earlier review is addressed.

Yours: the picker no longer depends on the international API succeeding, the anchor and the RowId 1200 exclusion carry their sources, and the `Slot + 1` asymmetry is written down.

Mine, pushed to your branch: the `AdvanceOnGrid` dedup had inverted the fallback path. A grid point is where results end and entry begins, so seeding it with Entry gave results for the first 4 days of the cycle and entry for the next 5, disagreeing with your own EndTime branch on the same server at the same instant. `/sales?server=1076` settles it: State 1 rows end 5 days past a grid point, State 2 rows end on one. That path covers 2955 of 2966 rows there, so it was the wrong phase on nearly every plot. Fixed, parameter renamed to `endingPhase` so the seed cannot be misread again, plus 15 tests pinning the cycle and both paths to the same schedule.

Thanks for the CN client screenshots and the server list, they are what made this checkable from outside the region. Merging.

PR #133, opened 2026-08-22
feat(Hunt) Added an hunt application to track hunt marks

Deldee:
## What
Added a new app, static data is stored in "Hunts/"

## Why
Feature request

## Checklist

Strings:

Quality:


src/Aetherphone/Core/Hunts/HuntsClient.cs
```diff
@@ -0,0 +1,136 @@
...
+
+    public async Task<HuntsLoginResponse?> LoginAsync(string username, string password, string? sessionId,
+        CancellationToken cancelToken)
+    {
+        var bearer = await tokens.GetTokenAsync(cancelToken).ConfigureAwait(false);
+        var request = new HuntsLoginRequest
+        {
+            Username = username, Password = password, RememberMe = true, SessionId = sessionId,
+        };
+        const string url = "https://faloop.app/api/auth/user/login";
```
XeldarAlz, line 37: Blocking: this posts the user's Faloop password to a private API, and the PR also mirrors 27MB of faloop.app images plus their translated tips for redistribution. This needs explicit sign off from the Faloop maintainers before merge, covering the API, the credential prompt and the assets.

src/Aetherphone/Aetherphone.csproj
```diff
@@ -17,6 +17,15 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Content Include="Hunts\*.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="Hunts\Maps\*.jpg">
```
XeldarAlz, line 23: Blocking: 27MB of assets ship to every user, against 1.5MB bundled today. The maps are 1300 to 1819px square but drawn at MapMaxSize 260 (HuntsApp.Detail.cs:20), so roughly 11MB VRAM each and aliased, since Dalamud gives us no mipmaps. Downscale to about 768 square.

src/Aetherphone/Core/Hunts/HuntsLoginFlow.cs
```diff
@@ -0,0 +1,76 @@
...
+
+    public HuntsLoginFlow(HuntsClient client, HuntsAuthTokenStore tokens)
+    {
+        this.client = client;
+        this.tokens = tokens;
+    }
+
+    public bool Busy => busy;
+
+    public bool ConsumeSucceeded()
```
XeldarAlz, line 21: Blocking: ConsumeSucceeded is never called anywhere, so nothing reacts to a successful login. EnsureActive only runs in OnOpened and self gates on loaded, so after signing in the user lands back on the Failed/Try again state. Wire this to Refresh plus EnsureHistoryLoaded and pop to the list.

src/Aetherphone/Core/Hunts/HuntsService.cs
```diff
@@ -0,0 +1,1086 @@
...
+        }
+
+        var windowRanks = result?.Data?.Session?.Access?.MobStatus?.Windows;
+        if (windowRanks is not null)
+        {
+            allowedRanks = new HashSet<string>(windowRanks.Keys, StringComparer.OrdinalIgnoreCase);
+        }
+    }
+
+    private void OnRealtimeConnectedChanged(bool value)
```
XeldarAlz, line 502: Blocking: a reconnect only flips a flag, and there is no periodic poll anywhere in Core/Hunts. After a socket drop every spawn during the gap is lost for good while the header still shows a healthy connection. Re-pull the data center on the false to true transition.

src/Aetherphone/Core/Hunts/HuntsService.cs
```diff
@@ -0,0 +1,1086 @@
...
+            activeSpawnKeys.Clear();
+        }
+
+        for (var index = 0; index < cleared.Length; index++)
+        {
+            notifications.RemoveGroup(NotificationGroupKeyFor(cleared[index]));
+        }
+    }
+
+    private void SetRealtimeLocation(HuntsSocketMobIdentity identity, int zonePoiId)
```
XeldarAlz, line 760: Blocking: this and SetRealtimePhase/Zone/Reporters do read copy write on a volatile field with no lock, while windows and history are guarded. Both the socket receive loop and the StoreWork thread write here (SeedActiveSpawnsFromPoll), so a refresh overlapping a live report drops a pin.

src/Aetherphone/Core/Hunts/HuntJsonCatalogLoader.cs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public void EnsureLoaded()
+    {
+        if (loaded)
+        {
+            return;
+        }
+
+        loaded = true;
```
XeldarAlz: Blocking: loaded is set before the file is read, so a second thread returns the still empty dictionary. Both the draw thread and the work thread hit this (SynthesizeMissingWindows uses mobCatalog.ById). ReadAllText of 1.2MB also runs on the render thread. Lock it and preload off thread.

src/Aetherphone/Apps/Hunts/HuntsApp.List.cs
```diff
@@ -0,0 +1,390 @@
...
+            if (searchText.Length > 0
+                && !ResolveMobLabel(mob, window.MobId).Contains(searchText, StringComparison.OrdinalIgnoreCase))
+            {
+                continue;
+            }
+
+            filteredWindows.Add(window);
+        }
+
+        filteredWindows.Sort(compareByPercentageDescending);
```
XeldarAlz, line 128: Blocking: this sorts the whole set every frame and the comparator takes activeSpawnGate twice per comparison (line 99). With A and S ranks across 8 worlds that is 1000 to 2500 windows, so tens of thousands of lock acquisitions per frame. Filter and sort on data change, then cache.

src/Aetherphone/Core/Maps/TravelPlanner.cs
```diff
@@ -97,10 +98,95 @@ public static TravelDestination Resolve(uint territoryId, uint worldId, uint cur
             return new TravelDestination(TravelKind.AlreadyThere, string.Empty, string.Empty, 0, 0, 0);
         }
 
+        if (targetCoordinate is { } coordinate && ResolveNearestAetheryte(territoryId, coordinate) is { } nearest)
```
XeldarAlz, line 108: Blocking: this changes destination resolution for every SharedLocation caller (Maps, Muster, venue and location shares, Yellow Pages), not just hunts. ResolveNearestAetheryte has no attunement check either, so it can pick an aetheryte TeleportToAetheryte will refuse. Scope it to the hunts caller.

src/Aetherphone/Core/Hunts/HuntDataCenterWorlds.cs
```diff
@@ -0,0 +1,81 @@
+using Lumina.Excel.Sheets;
+
+namespace Aetherphone.Core.Hunts;
+
+internal static class HuntDataCenterWorlds
+{
+    public static readonly IReadOnlyDictionary<string, string[]> ByDataCenter =
+        new Dictionary<string, string[]>(StringComparer.OrdinalIgnoreCase)
+        {
+            ["Aether"] = new[]
```
XeldarAlz: Non-blocking: no CN data centers here, so on a CN client ResolveDataCenter yields a name Faloop does not know and the app is a permanent failure screen. Add hunts to UnavailableInChina in Core/Aethernet/AppAvailability.cs:13. The hardcoded world table will also rot on every new world.

src/Aetherphone/Core/Hunts/HuntsRealtimeClient.cs
```diff
@@ -0,0 +1,346 @@
...
+    {
+        var attempt = 0;
+        while (!token.IsCancellationRequested)
+        {
+            var sessionId = tokens.SessionId;
+            if (string.IsNullOrEmpty(sessionId))
+            {
+                try
+                {
+                    await Task.Delay(NoSessionRetryDelay, token).ConfigureAwait(false);
```
XeldarAlz, line 92: Non-blocking: Start is called unconditionally from the HuntsService constructor, so a user who never opens Hunts gets this loop waking every 2 seconds for the whole session. Start the client on login instead, or signal it when a session id appears.

src/Aetherphone/Core/Hunts/HuntsAuthStoreToken.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        var request = new HuntsRefreshRequest { SessionId = currentSessionId };
+        var response = await http
+            .PostJsonAsync(RefreshUrl, request, HuntsJsonContext.Default.HuntsRefreshRequest,
+                HuntsJsonContext.Default.HuntsRefreshResponse, bearer: null, cancelToken,
+                onFailure: failure => AepLog.Warning($"Hunts token refresh failed: {failure.Describe()}"))
+            .ConfigureAwait(false);
+
+        var newToken = response?.Data?.Token;
```
XeldarAlz, line 171: Non-blocking: a failed refresh returns null but leaves authenticated true, so a revoked Faloop session keeps the header icon green and both tabs enabled while every call 401s. Clear authenticated here when the refresh yields no token.

src/Aetherphone/Apps/Hunts/HuntsApp.List.cs
```diff
@@ -0,0 +1,390 @@
...
+                ImGui.GetColorU32(Palette.WithAlpha(ui.TitleInk, 0.05f)));
+            ImGui.SetMouseCursor(ImGuiMouseCursor.Hand);
+        }
+
+        var def = mobCatalog.Find(window.MobId);
+        var mobLabel = ResolveMobLabel(def, window.MobId);
+
+        var textLeft = row.Min.X;
+        var nameY = row.Center.Y - 16f * scale;
+        DrawMarkNameLine("hunts.window." + window.MobId + window.WorldId + window.ZoneInstance, def, mobLabel,
```
XeldarAlz: Non-blocking: this concat allocates per row per frame, as do Prettify (line 367), the phase label (line 357), the status suffix (line 324) and BuildSpawnInfoText (Detail.cs:1006, built even while the section is collapsed). Memoize the world slug names and hoist the row ids.

src/Aetherphone/Hunts/HuntLore/en.json
```diff
@@ -0,0 +1,218 @@
+{
```
XeldarAlz: Non-blocking: nothing reads this folder and the csproj only copies the top level Hunts json glob, so all ten files here are dead weight. HuntLore/HuntMobDescriptions.json is also byte identical to the root copy. Worth deleting before a translator edits the wrong file.

src/Aetherphone/Core/Game/CharacterWatch.cs
```diff
@@ -25,7 +25,6 @@ public void Start()
         }
 
         started = true;
-        OnTick(framework);
         framework.Update += OnTick;
```
XeldarAlz, line 29: Non-blocking: dropping the immediate OnTick call here delays the first Changed event by a frame for every consumer, including ContentId keyed session restore. If this was to avoid reading LocalPlayer too early, say so in the commit and fix it at the subscription site instead.

src/Aetherphone/Core/Localization/L.cs
```diff
@@ -5135,6 +5343,11 @@ internal static class Onboarding
         public static readonly LocString AetherStreamPartyBody = new("onboarding.aetherStreamPartyBody", "Start a party and players nearby can ask to join. You decide who comes in, and everyone stays on the same second of the same video.");
         public static readonly LocString AetherStreamSettingsTitle = new("onboarding.aetherStreamSettingsTitle", "Tune it to your machine");
         public static readonly LocString AetherStreamSettingsBody = new("onboarding.aetherStreamSettingsBody", "Maximum quality, hardware decoding, whether others can find your stream, and the components the player needs, all behind this cog.");
+        public static readonly LocString HuntsBody = new("onboarding.huntsBody", "A Faloop account is required for live spawn data. Browsing marks and mob info works without one.");
```
XeldarAlz: Nit: without a token Refresh fails and the list renders the Failed state, so browsing marks does not actually work without an account today. Either make guest browsing real or reword this.

docs/game-integration.md
```diff
@@ -204,6 +204,8 @@ Shortcuts can also be imported from other players as a text code (`ShortcutCode`
...
 
+**Zone instance sync**: `LifestreamBridge` also subscribes to `Lifestream.GetCurrentInstance`, `Lifestream.CanChangeInstance`, and `Lifestream.ChangeInstance`. The hunts app's mark detail page (`HuntsApp.Detail.cs`) arms a pending watcher (`ArmPendingInstanceSync`, mirroring `ArmPendingWorldHop`) whenever a navigate action targets a mark with a known `ZoneInstance`, regardless of whether that navigation needs a world hop, an aetheryte teleport, or the player is already in the target territory. Each `Framework.Update` tick, once the player's territory and world match the target, it compares `GetCurrentInstance()` against the mark's tracked instance and calls `ChangeInstance` when they differ and `CanChangeInstance()` allows it, retrying until it succeeds or the same timeout used for pending flags and world hops elapses.
```
XeldarAlz, line 210: Nit: a whole new app landed with no README entry and no ChangelogData.cs bullet. CLAUDE.md asks for both when user visible behavior changes.

XeldarAlz (changes requested):
Release build clean, 0 warnings. dotnet test: 1452 passed, 0 failed. Not verified in game. Localization lockstep holds and the nine locales are genuinely translated. Findings are inline: 8 blocking, 6 non-blocking, 2 nits.

XeldarAlz:
Please squash before this merges. The branch carries 161 commits, but 18 of them are the Velvet not interested work it was branched from rather than master. Master already has that as squashed commit e0447193, so those will never drop out on their own.

Rebase and squash the hunts work down to 1 to 10 commits, one per coherent step. The diff itself is clean, no Velvet files are touched, so this is history hygiene only.

Also worth confirming the base: this targets master, but dev looks like the integration branch.

src/Aetherphone/Core/Localization/L.cs
Deldee: This is expected
Faloop treats a "Refresh" without a sessionId as an unauthenticated guest account and stil returns a sessionId and valid JWT token, the only difference is that the websocket will not send those sessions live spawn data (this is how the website itself deals with guest as well)
Example of pulling mark data when logged out : 
image

src/Aetherphone/Core/Hunts/HuntsClient.cs
Deldee: At faloop's request : 
- Reworded part of the app to make it clear we are unaffiliated with faloop
- Redid the tips in all languages to not use them as is
- Removed the images from the mob pages

XeldarAlz (approved):
Re-verified at 6ad87ed: all eight blocking findings addressed (bundled assets down to 4.8MB with 768px maps and no mob portraits, login success wired to refresh and history, reconnect re-pulls the data center, realtime state behind a lock, catalog loader locked and preloaded off thread, list filter and sort cached on change, TravelPlanner change scoped to the hunts caller). Non-blocking items and nits addressed too, CI green. Merging.

Deldee:
Proof of conversation with faloop dev : 
image
image

PR #134, opened 2026-08-22
fix(windows): Add safety gate for URLs in chirper and aethergram

Farroness:
## What

Add URL safety gate for Aethergram and Chirper.

## Why

URLs are currently highlighted but do not do anything on click.

Closes #

## How I tested it

1. Posted Link on Chirper/Aethergram.
2. Clicked Link.
3. Checked that the safety pop up appears.
4. Tested no/yes functionality.

## Screenshots

See discord.

## Backend coordination

None needed.

## Checklist

Strings:

Quality:


src/Aetherphone/Windows/UrlActions.cs
```diff
@@ -28,6 +30,24 @@ public static void OpenInBrowser(string url, Action<Exception>? onError = null)
         }
     }
 
+    public static void AskThenOpen(ConfirmService confirm, string url)
```
XeldarAlz, line 33: Blocking: LinkText.Draw (Components/LinkText.cs:21) still calls OpenInBrowser directly, so links in ChocoChat and Velvet DMs, Linkpearl, Venues and Yellow Pages stay one click away. A DM from a stranger is exactly what this warning describes. Put the gate in LinkText rather than two app level copies; PhoneServices.Confirm is already a single instance.

src/Aetherphone/Windows/UrlActions.cs
```diff
@@ -28,6 +30,24 @@ public static void OpenInBrowser(string url, Action<Exception>? onError = null)
...
+    {
+        confirm.Ask(new ConfirmRequest
+        {
+            Title = Loc.T(L.Common.OpenLinkTitle),
+            Message = string.Empty,
+            Sections =
+            [
+                ConfirmSection.Paragraph(Loc.T(L.Common.OpenLinkWarning)),
+                ConfirmSection.Chip(Loc.T(L.Common.OpenLinkDestination), url),
```
XeldarAlz, line 42: Non-blocking: the chip prints the whole URL and ConfirmDialog.cs:80 never clamps cardHeight, so a long URL (a legal single token) grows the card until the buttons sit off screen. Truncate the path and keep the host visible, which is the part a user actually judges.

src/Aetherphone/Apps/Chirper/ChirperApp.cs
```diff
@@ -1649,6 +1652,11 @@ private void DrawRichBody(ImDrawListPtr drawList, RichTextLayout layout, Vector2
         {
             OpenHashtag(layout.Tags[hit.TargetIndex]);
         }
+        
+        if (hit.Kind == RichTextRunKind.Link && hit.Clicked)
+        {
+            UrlActions.AskThenOpen(confirmService, layout.Urls[hit.TargetIndex]);
```
XeldarAlz, line 1658: Non-blocking: VelvetShell.cs:530 handles Mention only, so Velvet links still render underlined with the open in browser tooltip and then do nothing on click. Same bug this PR fixes here, one more block covers it.

src/Aetherphone/Apps/Chirper/ChirperApp.cs
```diff
@@ -55,6 +56,7 @@ internal sealed partial class ChirperApp : IPhoneApp
     private readonly RemoteImageCache images;
     private readonly SocialNotificationService social;
     private readonly ConductGateService conduct;
+    private readonly ConfirmService confirmService;
```
XeldarAlz, line 59: Nit: Aethergram and the rest of the repo call this field confirm and assign it with this.confirm = confirm. Rename for consistency, and group the assignment with the other fields instead of dropping it between drawView and back.

src/Aetherphone/Apps/Chirper/ChirperApp.cs
```diff
@@ -1649,6 +1652,11 @@ private void DrawRichBody(ImDrawListPtr drawList, RichTextLayout layout, Vector2
         {
             OpenHashtag(layout.Tags[hit.TargetIndex]);
         }
+
```
XeldarAlz, line 1655: Nit: trailing whitespace on this blank line.

XeldarAlz (changes requested):
Build clean, 1366 tests pass. I pushed one commit fixing the localization findings: the dialog now uses common.openLinkTitle, common.openLinkDestination and common.openLinkConfirm with the existing common.cancel, and the French warning switches to vouvoiement. Remaining findings inline, one of them blocking.

PR #139, opened 2026-08-23
fix(settings): sort the notifications app list by locale name

Deldee:
Sort NotificationsPage's app rows by their localized display name using the current culture's comparer, instead of the fixed declaration order in NotificationChannels.All. The sort is cached per language so it only recomputes when the locale actually changes.

the was done to make the notification list cleaner to use
## Checklist

Quality:

### DIFF
--- a/src/Aetherphone/Apps/Settings/Pages/NotificationsPage.cs
+++ b/src/Aetherphone/Apps/Settings/Pages/NotificationsPage.cs
@@ -6,6 +6,7 @@ using Aetherphone.Core.Notifications;
 using Aetherphone.Windows.Components;
 using Dalamud.Bindings.ImGui;
 using Dalamud.Interface;
+using System.Globalization;
 
 namespace Aetherphone.Apps.Settings.Pages;
 
@@ -19,6 +20,9 @@ internal sealed class NotificationsPage : ISettingsPage
     private readonly ISettingsNavigator navigator;
     private readonly AppNotificationPage appPage;
     private readonly AppInstaller installer;
+    private (NotificationChannel Channel, string Name)[] sortedChannels =
+        Array.Empty<(NotificationChannel Channel, string Name)>();
+    private LanguageInfo? sortedLanguage;
 
     public NotificationsPage(Configuration configuration, ISettingsNavigator navigator, AppNotificationPage appPage,
         AppInstaller installer)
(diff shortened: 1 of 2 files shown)


src/Aetherphone/Apps/Settings/Pages/NotificationsPage.cs
```diff
@@ -57,18 +60,19 @@ public void Draw(in PhoneContext context, Rect body)
...
-                var channel = channels[index];
+                var entry = sortedChannels[index];
+                var channel = entry.Channel;
                 if (!installer.IsInstalled(channel.AppId))
                 {
                     continue;
                 }
 
-                if (SettingsRow.AppLink(apps.NextRow(), channel.AppId, channel.Accent, Loc.T(channel.Name),
+                if (SettingsRow.AppLink(apps.NextRow(), channel.AppId, channel.Accent, entry.Name,
```
XeldarAlz: Non-blocking: the sort exposes a pre-existing mislabel. The `character` channel uses L.Apps.Character ("Character") but that app is ActivityApp, shown as L.Character.Activity ("Activity") everywhere else, so it files under C instead of A. Pointing NotificationChannels.cs:27 at L.Character.Activity fixes it; fine to ride along here or go separately.

src/Aetherphone/Apps/Settings/Pages/NotificationsPage.cs
```diff
@@ -80,6 +84,25 @@ public void Draw(in PhoneContext context, Rect body)
...
+        }
+
+        var channels = NotificationChannels.All;
+        var sorted = new (NotificationChannel Channel, string Name)[channels.Count];
+        for (var index = 0; index < channels.Count; index++)
+        {
+            sorted[index] = (channels[index], Loc.T(channels[index].Name));
+        }
+
+        Array.Sort(sorted, static (left, right) => Loc.Culture.CompareInfo.Compare(left.Name, right.Name));
```
XeldarAlz: Nit: this compare is case-sensitive while the App Store sort ignores case, and Array.Sort is unstable with no secondary key. Suggest `CompareOptions.IgnoreCase` and, when it returns 0, `string.CompareOrdinal(left.Channel.AppId, right.Channel.AppId)` so the order is total and matches the store.

src/Aetherphone/Apps/Settings/Pages/NotificationsPage.cs
```diff
@@ -19,6 +19,9 @@ internal sealed class NotificationsPage : ISettingsPage
     private readonly ISettingsNavigator navigator;
     private readonly AppNotificationPage appPage;
     private readonly AppInstaller installer;
+    private (NotificationChannel Channel, string Name)[] sortedChannels =
```
XeldarAlz: Nit: the repo idiom for per-language caches is a `readonly` buffer refilled in place (GameChannels.Names, WheelRingArt.SpotLabels). A `readonly NotificationChannel[]` sorted in place with Loc.T inside the comparator keeps the field readonly, drops the tuple type, and lets the draw loop keep Loc.T(channel.Name) as before.

XeldarAlz (commented):
Verified: CI green on f5c0c3db (Release build + tests, Guards), clean merge against dev, no per-frame allocation, and the Loc.Current cache key matches the idiom already used in GameChannels and WheelRingArt. No blocking findings, three small inline notes. Two asks: was this checked in-game on /phone with a language switch, and the squash title reads better as feat(settings) since it adds behavior rather than fixing a bug.

PR #140, opened 2026-08-23
feat(velvet) add persistent, independent filters for discovery/feed

Deldee:
## What

Changed velvet filtering : 
- Instead of a single mute list, have two independent filters for discovery and feed
- Both filter are saved locally, so they persist between sessions
- Backward compatibility : if the old MuteList exist, instanciate the new filters with it, ignore this config from every boot onward in favor of the other filters
- Stores in a similar pattern used for not interested, so that configuration dont share across accounts

## Why
Discovery and feed show different enough that data that sharig the filter and mutelist doesnt make much sense, and filters reeset every session

Closes #

## How I tested it

 - Log in to velvet, check that the previous muted items were on
 - Edit the discovery filter
 - Ensure that filter applies
 - Swap to the feed tab, notice the original mutelist is present, but not the new filter added to discovery
 - set different filters for the feed
 - Reset the plugin
 - Both filter should still be uniquely present

## Checklist

Strings:

Quality:


src/Aetherphone/Configuration.cs
```diff
@@ -221,7 +221,8 @@ internal sealed class Configuration : IPluginConfiguration, IHomeConfiguration,
 
     public bool IsVelvetOnboarded() => VelvetOnboarded && VelvetOnboardedVersion >= VelvetOnboardVersion;
     public bool VelvetBlurByDefault { get; set; } = true;
-    public VelvetMutePreferences VelvetMutes { get; set; } = new();
+    public VelvetFilterPreferences VelvetMutes { get; set; } = new();
```
XeldarAlz, line 273: Blocking: renaming VelvetMutePreferences orphans the `$type` marker this property was persisted with. Existing configs hold the old type name here, so GetPluginConfig throws on the binder and falls back to `new Configuration()`: total config loss on upgrade. Needs a ConfigMigrations.TypeRenames pair plus a test.

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
```diff
@@ -10,28 +10,93 @@ namespace Aetherphone.Apps.Velvet;
...
+        ExcludeFor(surface).SaveInto(ExcludePreferencesFor(surface));
+        SaveFilters();
+    }
+
+    private void SaveFilters() => filterArchive.Save(filtersAccountId, storedFilters);
+
+    // Filters live per account, same as the not interested list, so switching accounts on the
+    // same install never mixes one person's filters into another's. Runs at construction and
+    // again every time the signed-in account actually changes.
+    private void LoadFiltersForAccount(string accountId)
```
XeldarAlz: Blocking: this restores the selections but never re-applies them. CurrentUser is null at construction, so OnOpened's ApplyFeedFilters runs empty (SetFeedFilter early-returns) and DrawDiscover fetches unfiltered. When /me lands and this runs, nothing refetches: the chips show the saved filters while the results ignore them until the app is reopened.

src/Aetherphone/Apps/Velvet/VelvetShell.cs
```diff
@@ -120,7 +124,23 @@ public VelvetShell(AethernetSession session, AethernetApi net, LodestoneService
...
+        if (initialAccountId is not null)
+        {
+            LoadFiltersForAccount(initialAccountId);
+        }
+    }
+
+    private void OnFilterAccountChanged()
+    {
+        var accountId = session.CurrentUser?.Id;
+        if (accountId is null || string.Equals(accountId, filtersAccountId, StringComparison.Ordinal))
```
XeldarAlz: Blocking: SwitchTo nulls CurrentUser before raising Changed, so this early-returns and filtersAccountId keeps pointing at the previous account while the new token is already installed. The chips shown, and every chip toggle saved, belong to the old account, then get overwritten once /me resolves for the new one.

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
```diff
@@ -10,28 +10,93 @@ namespace Aetherphone.Apps.Velvet;
 
 internal sealed partial class VelvetShell
 {
-    private readonly VelvetFilterSelection mutes = new();
     private readonly VelvetFilterSelection discoverInclude = new();
+    private readonly VelvetFilterSelection discoverExclude = new();
     private readonly VelvetFilterSelection feedInclude = new();
+    private readonly VelvetFilterSelection feedExclude = new();
+    private readonly StoredVelvetFilters storedFilters = new();
+    private string filtersAccountId = string.Empty;
```
XeldarAlz: Blocking: this stays empty until /me resolves, and VelvetFilterArchive.Save early-returns on an empty account id. IsSignedIn is token-based, so the Filters screen is fully usable meanwhile: every chip tap is silently dropped, then visibly reset when the load finally lands.

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
```diff
@@ -10,28 +10,93 @@ namespace Aetherphone.Apps.Velvet;
...
+        SaveFilters();
+    }
+
+    private void SaveExclude(VelvetPage surface)
+    {
+        ExcludeFor(surface).SaveInto(ExcludePreferencesFor(surface));
+        SaveFilters();
+    }
+
+    private void SaveFilters() => filterArchive.Save(filtersAccountId, storedFilters);
```
XeldarAlz: Blocking: serialize plus File.WriteAllText plus File.Move on the draw thread, and Load does ReadAllText from the session.Changed handler. A chip cycling include to exclude writes the whole file twice in one frame (206/211, same at 134/135). VelvetStore.cs:1225 wraps this same archive in Task.Run for exactly this reason.

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
```diff
@@ -10,28 +10,93 @@ namespace Aetherphone.Apps.Velvet;
 
 internal sealed partial class VelvetShell
 {
-    private readonly VelvetFilterSelection mutes = new();
     private readonly VelvetFilterSelection discoverInclude = new();
+    private readonly VelvetFilterSelection discoverExclude = new();
```
XeldarAlz: Blocking: this makes exclusions per-surface, so a hard limit excluded on Discover no longer hides Feed posts. The deleted velvet.filterMuteHint promised matches were hidden "everywhere in Velvet". Includes per-surface is right, but on an 18+ app exclusions should stay global, or the split needs a Discover-to-Feed migration.

src/Aetherphone/Apps/Velvet/VelvetFilterArchive.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        try
+        {
+            var path = PathFor(accountId);
+            if (!File.Exists(path))
+            {
+                return new StoredVelvetFilters();
+            }
+
+            var stored = JsonConvert.DeserializeObject<StoredVelvetFilters>(File.ReadAllText(path));
```
XeldarAlz, line 54: Non-blocking: only the root is guarded. A file with `"DiscoverInclude": null` deserializes to a non-null root, and LoadFrom(null) NREs at stored.Intent inside the session.Changed handler, outside this try/catch. VelvetNotInterestedArchive guards its payload (`stored?.UserIds is { Count: > 0 }`).

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
```diff
@@ -10,28 +10,93 @@ namespace Aetherphone.Apps.Velvet;
...
+    // signed in; the legacy list is never read again afterward.
+    private void MigrateLegacyMutes(string accountId)
     {
-        mutes.SaveInto(configuration.VelvetMutes);
+        if (accountId.Length == 0 || configuration.VelvetMutesMigrated)
+        {
+            return;
+        }
+
+        configuration.VelvetMutesMigrated = true;
```
XeldarAlz: Non-blocking: the one-shot flag is committed before the archive write is known to have succeeded. If SaveFilters at :91 fails, Save only logs a warning and the legacy mute list is gone with no retry. Set the flag after a confirmed write. SaveInto at 89/90 also overwrites rather than merging into what was loaded.

src/Aetherphone/Apps/Velvet/VelvetShell.Discover.cs
```diff
@@ -118,6 +118,7 @@ private void DrawDiscover(Rect area)
                             ConfirmButtonTone.Primary, "velvet.discover.clearFilters"))
                     {
                         discoverInclude.Clear();
+                        SaveInclude(VelvetPage.Discover);
```
XeldarAlz: Non-blocking: :110 gates this button on discoverInclude.Any only, so a list emptied purely by excludes offers no way out, and clearing here leaves discoverExclude applied right after telling the user the filters were cleared. Now that excludes are per-surface state, gate on both and clear both.

src/Aetherphone/Apps/Velvet/VelvetFilterArchive.cs
```diff
@@ -0,0 +1,91 @@
...
+                File.Move(temp, path, true);
+            }
+        }
+        catch (Exception exception)
+        {
+            AepLog.Warning(exception, $"VelvetFilterArchive write failed for {accountId}");
+        }
+    }
+
+    private string PathFor(string accountId)
```
XeldarAlz: Non-blocking: PathFor is byte-identical to VelvetNotInterestedArchive.PathFor, and this SHA-256-account-id-to-filename pattern also lives in ChatArchive, MessageArchive and DiskCache. Fifth copy, and it already diverged by dropping the payload null guard. Worth one shared helper.

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
```diff
@@ -62,28 +120,29 @@ private void DrawFilters(Rect area)
...
         }
 
-        if ((include.Any || mutes.Any) && ui.HeaderAction(area, Loc.T(L.Velvet.FilterClearAll), true))
+        if ((include.Any || exclude.Any) && ui.HeaderAction(area, Loc.T(L.Velvet.FilterClearAll), true))
         {
             include.Clear();
-            mutes.Clear();
-            ApplyMutesEverywhere();
+            exclude.Clear();
+            SaveInclude(surface);
```
XeldarAlz: Nit: SaveInclude then SaveExclude serializes and writes the whole file twice, and the first write lands with the pre-clear excludes still in it. Same at 206/211 when a chip cycles. One SyncSurface(surface) that copies both selections across then calls SaveFilters once halves the I/O.

src/Aetherphone/Configuration.cs
```diff
@@ -221,7 +221,8 @@ internal sealed class Configuration : IPluginConfiguration, IHomeConfiguration,
...
+    public bool VelvetMutesMigrated { get; set; }
```
XeldarAlz: Nit: once this flips, VelvetMutes is never read again but keeps serializing into every config save. Resetting it to a fresh instance after a successful migration stops the config carrying a dead block forever, and avoids a future removal needing another $type migration.

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
```diff
@@ -10,28 +10,93 @@ namespace Aetherphone.Apps.Velvet;
...
+
+    private void SaveExclude(VelvetPage surface)
+    {
+        ExcludeFor(surface).SaveInto(ExcludePreferencesFor(surface));
+        SaveFilters();
+    }
+
+    private void SaveFilters() => filterArchive.Save(filtersAccountId, storedFilters);
+
+    // Filters live per account, same as the not interested list, so switching accounts on the
```
XeldarAlz: Nit: these narrate control flow. docs/conventions.md:27 allows a comment only for a why the code cannot express, or the source of a magic constant, never what the code does. Same for 69-72. The method names and the guard clauses already carry both.

XeldarAlz (changes requested):
Release build clean (0 warnings, 0 errors) and 1412/1412 tests pass on 8a0f865c. Not tested in-game.

Six blocking findings inline. The first one, the VelvetMutePreferences rename with no ConfigMigrations entry, wipes every existing user's config on upgrade.

Title nit: `feat(velvet) add...` is missing the conventional-commit colon.

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
Deldee: @XeldarAlz 
I think that it makes more sense to filter it per surface, since both filter apply to different entity

The discovery filter applies to people, and contains field such as interests, sexual orientation, , relationship status, pace, etc
On the other hand, the feed applies to content, not to people, for which the aforementionned field dont really matter, sharing exclusion removes more content from the feed than would probably be expected (for example, if I excluded "non irl" on the discovery, everyone non-irl would be remove from the feed, despite it not having any logical relations)

As such, i think it'd be better to keep both filter per surface, since they apply to different context and viewing scenarios

While there are some scenarios where both would be the same (gender for example would be a big one), i think it'd be better to ultimately let the user decided who they want to look for and what they want to see idividually

Note that the current migration would keep the current filters they have in place, so their filters wouldnt suddenly be removed for people that already put limits on both

Note that this commit also removed the key saying that filter are platform wide, for this exact reason

If we want to keep filter globals, then we may want to offer the ability to swap between two profiles, to accomodate both reality, though this is more work than just considering them both different

Deldee: Alternatively, we could apply the shared exclude only to the relevant subtypes : 
Gender
Kinks
Limit

src/Aetherphone/Apps/Velvet/VelvetShell.cs
```diff
@@ -223,6 +241,13 @@ public void Draw(in PhoneContext context)
             return;
         }
 
+        if (session.CurrentUser is null)
```
XeldarAlz, line 272: Blocking: this spinners the whole app, and Velvet never triggers the fetch that clears it. CurrentUser is only filled by AccountClient.EnsureCurrentUser, which fires from CharacterSessionManager on a content id change and never retries, so one failed /me locks Velvet for the session. Either call EnsureCurrentUser here, or gate only the Filters entry instead of the app.

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
```diff
@@ -10,28 +10,114 @@ namespace Aetherphone.Apps.Velvet;
...
+    private async Task LoadFiltersForAccountAsync(string accountId)
+    {
+        filtersAccountId = accountId;
+        var loaded = await Task.Run(() => filterArchive.Load(accountId)).ConfigureAwait(false);
+        if (!string.Equals(filtersAccountId, accountId, StringComparison.Ordinal))
+        {
+            return;
+        }
+
+        var migrating = accountId.Length > 0 && !configuration.VelvetMutesMigrated;
```
XeldarAlz: Blocking: VelvetMutesMigrated is one global bool and :96 wipes VelvetMutes, but the legacy mutes were global config shared by every account. The first account to sign in absorbs them, and the second silently starts with no exclusions. Key the one shot per account id, and keep the legacy block until each has taken it.

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
```diff
@@ -10,28 +10,114 @@ namespace Aetherphone.Apps.Velvet;
...
+            if (migrating)
+            {
+                merged = MergeLegacyMutes();
+            }
+
+            LoadInclude(VelvetPage.Discover);
+            LoadExclude(VelvetPage.Discover);
+            LoadInclude(VelvetPage.Feed);
+            LoadExclude(VelvetPage.Feed);
+            ApplyDiscoverFilters();
```
XeldarAlz: Blocking: RefreshDiscover only checks IsSignedIn, so every signed in user now hits /velvet/discover at boot and on each account switch, including people who never accepted the 18+ gate or opened the app. DrawDiscover already self heals at VelvetShell.Discover.cs:56. Gate this pair on GateAccepted and IsVelvetOnboarded.

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
```diff
@@ -10,28 +10,114 @@ namespace Aetherphone.Apps.Velvet;
...
+    private void SaveInclude(VelvetPage surface) => IncludeFor(surface).SaveInto(IncludePreferencesFor(surface));
+
+    private void SaveExclude(VelvetPage surface) => ExcludeFor(surface).SaveInto(ExcludePreferencesFor(surface));
+
+    private void SaveFilters() => _ = SaveFiltersAsync();
+
+    private Task<bool> SaveFiltersAsync()
+    {
+        var accountId = filtersAccountId;
+        return Task.Run(() => filterArchive.Save(accountId, storedFilters));
```
XeldarAlz: Non-blocking: this captures the live storedFilters, not a copy. If an account switch lands between the queue and the serialize, :71-74 swaps in account B's data and this task writes it into account A's file. Snapshot into a fresh StoredVelvetFilters inside SyncSurface and pass that.

src/Aetherphone/Apps/Velvet/VelvetShell.Filters.cs
```diff
@@ -10,28 +10,114 @@ namespace Aetherphone.Apps.Velvet;
 
 internal sealed partial class VelvetShell
 {
-    private readonly VelvetFilterSelection mutes = new();
     private readonly VelvetFilterSelection discoverInclude = new();
+    private readonly VelvetFilterSelection discoverExclude = new();
```
XeldarAlz: Answering my earlier block: your reasoning holds for intent, relationship, pace and region, those are discovery vocabulary and mean nothing about a post. Gender, kinks and limits read identically on both surfaces and are the safety shaped ones, so keep those three global as you offered and leave the rest per surface. No need for filter profiles.

src/Aetherphone/Apps/Velvet/VelvetShell.Discover.cs
```diff
@@ -338,9 +337,10 @@ private void AddActiveFilterChip(in VChipModel model, int kind, int flag, string
         activeFilterExcluded.Add(excluded);
     }
 
-    private bool RemoveActiveFilter(VelvetFilterSelection include, int kind, int flag, string token, bool excluded)
+    private bool RemoveActiveFilter(VelvetFilterSelection include, VelvetFilterSelection exclude, int kind, int flag,
```
XeldarAlz, line 351: Nit: the bool return is no longer read at the only call site (:282), which now calls SyncSurface and ApplyFilters unconditionally. Make it void.

src/Aetherphone/Apps/Velvet/VelvetFilterSelection.cs
```diff
@@ -35,31 +35,50 @@ public void Clear()
...
 
+    public void MergeInto(VelvetFilterPreferences stored)
+    {
+        stored.Intent |= Intent;
+        stored.Gender |= Gender;
+        stored.Sexuality |= Sexuality;
+        stored.Relationship |= Relationship;
+        if (stored.Region.Length == 0)
+        {
+            stored.Region = Region;
```
XeldarAlz: Nit: unreachable. The only caller loads from configuration.VelvetMutes, and the legacy prefs never had a Region property, so this always copies an empty string. Drop the branch.

src/Aetherphone.Tests/ConfigMigrationsTests.cs
```diff
@@ -16,6 +16,16 @@ public void RewritesLegacyStarredMessageTypeName()
...
     }
 
+    [Fact]
+    public void RewritesLegacyVelvetMutePreferencesTypeName()
+    {
+        const string json =
+            """{"VelvetMutes":{"$type":"Aetherphone.Core.Social.VelvetMutePreferences, Aetherphone","Intent":1}}""";
+        var rewritten = ConfigMigrations.RewriteTypeNames(json);
+        Assert.Contains("Aetherphone.Core.Social.VelvetFilterPreferences, Aetherphone", rewritten);
+        Assert.DoesNotContain("VelvetMutePreferences", rewritten);
```
XeldarAlz, line 27: Non-blocking: the type rename got a test but the two pieces most likely to break silently did not. A VelvetFilterArchive save and load round trip, and a MergeLegacyMutes case asserting the legacy list reaches both excludes, would both be cheap here.

XeldarAlz (changes requested):
Release build clean (0 warnings), 1413/1413 tests pass, guards green. Not run in game.

Twelve of the thirteen earlier findings are fixed, one commit each: nice work. Three new blocking issues came in with the fixes, plus my answer on the exclusion split. The branch also conflicts with dev now (the VelvetShell constructor gained ChatHistory, Translation and EncryptionHelp), so it needs a rebase.

XeldarAlz:
Probably wait this more longer because of the current ongoing velvet update

XeldarAlz:
Might have to do again with the new system

PR #141, opened 2026-08-23
fix(ui): stop InputShield from hijacking an in-flight drag scroll

Deldee:
## What
Pushing a new router view shields the outgoing view for the transition by moving io.MousePos off-screen. DragScrollHost read that sentinel position straight into an already-pressed KineticScroller, which read as a huge swipe toward the sentinel and clamped the scroll to one end.

Tapping a button that both lives inside a scrollable AppSurface and pushes a new view in the same frame (Velvet's Discover/Feed filter button, for one) latches a press right before the shield engages, so the very next transitioning frame snaps the scroll to the bottom.

Cancel any in-flight gesture for a region while the shield is active instead of feeding it the off-screen pointer position.

## Why

Closes #

## How I tested it
 - Open Velvet's feed
 - Scroll down to load entries
 - Scroll up and click the filter button
 - Return to the Feed page
 - Confirm that thescroll does not snap to the bottom

## Checklist

Strings:

Quality:

### DIFF
--- a/src/Aetherphone/Windows/Components/DragScrollHost.cs
+++ b/src/Aetherphone/Windows/Components/DragScrollHost.cs
@@ -109,6 +109,16 @@ internal static class DragScrollHost
             return new Surface(region, 0f, false);
         }
 
+        if (InputShield.Active)
+        {
+            if (region.Pressed)
+            {
+                region.CancelGesture();
+            }
+
+            return new Surface(region, scroller.PullDistance, false);
+        }
+
         var io = ImGui.GetIO();
         var deltaSeconds = io.DeltaTime;
         var pointerY = io.MousePos.Y;


src/Aetherphone/Windows/Components/DragScrollHost.cs
```diff
@@ -109,6 +109,16 @@ public static Surface Begin(uint key)
             return new Surface(region, 0f, false);
         }
 
+        if (InputShield.Active)
+        {
+            if (region.Pressed)
```
XeldarAlz, line 114: Blocking: gating on `region.Pressed` leaves a fling or a released spring-back frozen, because this return also skips `scroller.Tick` and `SetScrollY`. Flick a feed, then park the cursor on the Dynamic Island (its hover feeds `ShieldBase`, PhoneShell.cs:314): momentum stops dead until you move off. Use `region.Pressed || scroller.IsControlling`, like the `!Enabled` branch above.

src/Aetherphone/Windows/Components/DragScrollHost.cs
```diff
@@ -109,6 +109,16 @@ public static Surface Begin(uint key)
...
+            {
+                region.CancelGesture();
+            }
+
+            return new Surface(region, scroller.PullDistance, false);
```
XeldarAlz, line 119: Blocking: `PullDistance` is reported without the `ImGui.Dummy(0, PullDistance)` spacer the normal path emits at the end of `Begin`, so a shield engaging during spring-back collapses the reserved gap while `Surface.Pull` keeps drawing the indicator, and the content jumps back when it lifts. Return `0f` here, as `!Enabled` does.

src/Aetherphone/Windows/Components/DragScrollHost.cs
```diff
@@ -109,6 +109,16 @@ public static Surface Begin(uint key)
             return new Surface(region, 0f, false);
         }
 
+        if (InputShield.Active)
```
XeldarAlz, line 112: Non-blocking: `InputShield.Active` is broader than the transition in the title. `ShieldBase` also folds in hover over the Dynamic Island, the notification banner and the shortcut pill, so a drag whose pointer crosses one now dies mid-swipe and cannot re-latch (re-latching needs `IsMouseClicked`). Still better than today's snap to bottom, but those captures are worth backing off while `DragScrollHost.AnyDragging`.

XeldarAlz (changes requested):
Release build clean, 1412/1412 tests pass. Not run in-game. Right root cause and right place for the guard; two blocking notes inline, both about this early return skipping the rest of the frame's scroller work.

XeldarAlz:
Merged, and the two blocking notes are fixed on `dev` in 0afcf330: the shield branch now cancels on `scroller.IsControlling` too and returns a zero pull, so a fling or spring-back no longer stalls and the missing `Dummy` spacer cannot desync from `Surface.Pull`.

Followed up on the non-blocking note in 33c8eacc: `islandCaptures` now yields to `DragScrollHost.AnyDragging`, so a drag whose pointer crosses the dynamic island keeps scrolling instead of dying mid-swipe. Thanks for tracking the off-screen `MousePos` down, that was the hard part.

PR #142, opened 2026-08-23
feat(Hunts) read hunt map from client, remove dead code in static files

Deldee:
## What

Removed the hunt maps to read from the game client
Cleaned up the static file to remove anything related to the old file, as well as unused section

## Why
Files are big, they're already on the client

Closes #

## How I tested it

- Open the app
- Load a mark
- Ensure landmines are accurate
- Load another mark, ensure landmine are accurate
- Teleport to an aetheryte from the map, to be certain

## Checklist

Strings:

Quality:


src/Aetherphone/Apps/Hunts/HuntsApp.Detail.cs
```diff
@@ -503,12 +504,17 @@ private bool DrawDetailZoneMap(float scale, int? confirmedPoiId, HuntsView view)
...
     }
 
-    private string ResolveZoneLabel(HuntZoneDefinition zone) =>
-        zone.Name.GetValueOrDefault(configuration.Language) ?? zone.Name.GetValueOrDefault("en") ??
-        Prettify(zone.Id);
+    private string ResolveZoneLabel(HuntZoneDefinition zone, uint territoryId) =>
+        ResolveLiveZoneName(territoryId) is { Length: > 0 } name ? name : Prettify(zone.Id);
+
+    private static string? ResolveLiveZoneName(uint territoryId) =>
+        territoryId != 0 && Plugin.DataManager.GetExcelSheet<TerritoryType>(HuntUiLanguage.GameClientLanguage())
```
XeldarAlz: Blocking: ExtractText() allocates a new string every frame the zone map is on screen, and Draw runs every frame. The old zone.Name lookup allocated nothing. Cache the resolved name per territory and UI locale, the way worldLabelCache does in HuntsApp.List.cs:238.

src/Aetherphone/Core/Hunts/HuntZoneMapTextures.cs
```diff
@@ -1,10 +1,107 @@
...
 
 namespace Aetherphone.Core.Hunts;
 
-internal static class HuntZoneMapTextures
+internal sealed class HuntZoneMapTextures
 {
-    private static readonly FileTextureCache Cache = new(Path.Combine("Hunts", "Maps"), ".jpg");
+    private readonly IDataManager data;
+    private readonly ITextureProvider textures;
+    private readonly Dictionary<uint, string?> texturePathByTerritory = new();
```
XeldarAlz, line 12: Blocking: this branch is 207 commits behind dev, which moved Core/FileTextureCache.cs to Core/Media/FileTextureCache.cs. The merge resurrects that file with no callers left. Rebase and delete Core/Media/FileTextureCache.cs, Release builds clean without it.

src/Aetherphone/Core/Hunts/HuntZoneCatalogue.cs
```diff
@@ -76,7 +77,7 @@ private void OnLoaded(Dictionary<string, HuntZoneDefinition> parsed)
     private static Dictionary<string, uint> BuildTerritoryIdLookup()
     {
         var lookup = new Dictionary<string, uint>(StringComparer.OrdinalIgnoreCase);
-        foreach (var territory in Plugin.DataManager.GetExcelSheet<TerritoryType>())
+        foreach (var territory in Plugin.DataManager.GetExcelSheet<TerritoryType>(ClientLanguage.English))
```
XeldarAlz: Non-blocking: GetExcelSheet<T>(language) throws UnsupportedLanguageException when the sheet supports neither the requested language nor None. This runs lazily from the draw path with no try/catch and the result is cached once. Worth guarding for the CN client.

src/Aetherphone/Core/Maps/MapPixelMath.cs
```diff
@@ -2,11 +2,16 @@ namespace Aetherphone.Core.Maps;
...
     {
         var scale = 41f / (Math.Max(sizeFactor, 1) / 100f);
-        var x = MathF.Round(10f * (rawX / 2048f * scale + 1f)) / 10f;
-        var y = MathF.Round(10f * (rawY / 2048f * scale + 1f)) / 10f;
+        var x = MathF.Round(10f * (rawX / FullCanvasSize * scale + 1f)) / 10f;
+        var y = MathF.Round(10f * (rawY / FullCanvasSize * scale + 1f)) / 10f;
         return (x, y);
     }
+
+    public static (float X, float Y) NormalizeToFullCanvas(float rawX, float rawY) =>
```
XeldarAlz, line 15: Non-blocking: pixelSize and offset were also the display crop. On the full texture the zone fills as little as 47% of the 260px panel (outer_la_noscea, north_shroud 52%, median 89%). Dots stay accurate, they just cluster. Fitting the initial view to the POI bounding box would fix it, but PhotoZoomView exposes only Reset().

src/Aetherphone/Core/Hunts/HuntZoneMapTextures.cs
```diff
@@ -1,10 +1,107 @@
...
+            if (FileExists(candidates[index]))
+            {
+                return candidates[index];
+            }
+        }
+
+        return null;
+    }
+
+    private static string[] TextureCandidates(string mapId)
```
XeldarAlz: Non-blocking: TextureCandidates, FileExists and ResolveTexturePath are copied from HousingGameMaps.cs:309-352, minus the _s.tex candidate. Worth extracting into Core/Maps, and confirming the dropped candidate was deliberate.

src/Aetherphone/Core/Hunts/HuntUiLanguage.cs
```diff
@@ -11,4 +12,12 @@ internal static class HuntUiLanguage
         "ja" => "ja",
         _ => "en",
     };
+
+    public static ClientLanguage GameClientLanguage() => Loc.Current.Code switch
```
XeldarAlz: Non-blocking: this sits next to HuntClientLanguage.Key(), which returns the actual client language, while this one returns the sheet language for the UI language. SheetLanguage() reads truer.

src/Aetherphone/Core/Hunts/HuntZoneModels.cs
```diff
@@ -25,25 +19,10 @@ internal sealed class HuntZoneDefinition
...
 internal sealed class HuntZoneMap
 {
-    [JsonPropertyName("pixelSize")]
-    public int PixelSize { get; set; }
-
-    [JsonPropertyName("offset")]
-    public HuntZoneMapOffset Offset { get; set; } = new();
-
     [JsonPropertyName("sizeFactor")]
     public int SizeFactor { get; set; } = 100;
```
XeldarAlz, line 23: Nit: 41 zones now carry an empty "map": {} in HuntPOI.json. HuntZoneDefinition.Map already defaults and SizeFactor already defaults to 100, so the key can go from those zones entirely.

src/Aetherphone/Apps/Hunts/HuntsApp.Detail.cs
```diff
@@ -503,12 +504,17 @@ private bool DrawDetailZoneMap(float scale, int? confirmedPoiId, HuntsView view)
         return true;
     }
 
-    private string ResolveZoneLabel(HuntZoneDefinition zone) =>
-        zone.Name.GetValueOrDefault(configuration.Language) ?? zone.Name.GetValueOrDefault("en") ??
-        Prettify(zone.Id);
+    private string ResolveZoneLabel(HuntZoneDefinition zone, uint territoryId) =>
```
XeldarAlz: Nit: zone is only read here for zone.Id. Passing the zone id would say what the method actually needs.

XeldarAlz (changes requested):
Merged onto dev and built: Release clean, 0 warnings, tests 1820/1821 (the one failure is SpotlightMathTests "7/2", a decimal separator issue on my machine, unrelated). Not verified in-game. I re-checked the data claims and they hold: sizeFactor is byte identical, all 1064 mob and 91 aetheryte POIs survive, all 1063 zonePoiIds still resolve, and every POI in the old data sits inside its zone's crop rect, so normalizing by 2048 is the right call. Two blocking comments inline.

PR #143, opened 2026-08-23
fix(icontile): use the draw-list CenterIcon overload, not the cursor one

BluntEXE:
Split out of #(venue-manager-app), per review feedback there: the fix touches every screen with an IconTile, not just Venue Sync, so it gets its own PR and its own visual pass.

## What

`IconTile.Draw` called `ProgressRing.CenterIcon`'s cursor-based overload, which calls `SetCursorScreenPos` plus `ImGui.TextUnformatted` internally and moves ImGui's real layout cursor. `IconTile.Draw` is used across screens built entirely on manual Rect-based row layout (Activity, Collections, Dailies, Fishing, Health, Muster, Timers, Venues, Yellow Pages, CoinStreakCard, Venue Sync), so every tile drawn left the cursor corrupted for whatever ImGui call came next.

Switched to the sibling draw-list overload. Same centering math (measured glyph size, `GlyphPen` offset), rendered through `dl.AddText` instead of `SetCursorScreenPos` plus a text widget, so it draws identically without the cursor side effect.

## Test plan

- [x] `dotnet build Aetherphone.sln -c Release`: 0 warnings, 0 errors
- [x] `dotnet test Aetherphone.sln -c Release`: 1557/1557 passing
- [ ] Visual pass across the 11 affected screens (not done here, that's the reason for the separate PR)

### DIFF
--- a/src/Aetherphone/Windows/Components/Chrome/IconTile.cs
+++ b/src/Aetherphone/Windows/Components/Chrome/IconTile.cs
@@ -12,7 +12,7 @@ internal static class IconTile
         var half = size * 0.5f;
         Squircle.Fill(drawList, center - new Vector2(half, half), center + new Vector2(half, half),
             size * Metrics.Radius.TileFactor, ImGui.GetColorU32(tint));
-        ProgressRing.CenterIcon(center, icon, AccentRing.Ink, size * 0.50f);
+        ProgressRing.CenterIcon(drawList, center, icon, AccentRing.Ink, size * 0.50f);
     }
 
     public static void DrawApp(ImDrawListPtr drawList, string appId, Vector2 center, float size, Vector4 surface)


src/Aetherphone/Windows/Components/IconTile.cs
```diff
@@ -12,7 +12,7 @@ public static void Draw(Vector2 center, float size, Vector4 tint, FontAwesomeIco
         var half = size * 0.5f;
         Squircle.Fill(drawList, center - new Vector2(half, half), center + new Vector2(half, half),
             size * Metrics.Radius.TileFactor, ImGui.GetColorU32(tint));
-        ProgressRing.CenterIcon(center, icon, AccentRing.Ink, size * 0.50f);
+        ProgressRing.CenterIcon(drawList, center, icon, AccentRing.Ink, size * 0.50f);
```
XeldarAlz, line 15: Blocking: this conflicts because dev moved the toolkit in 0c8f4e8c (182 renames in one commit), so the file now lives at `src/Aetherphone/Windows/Components/Chrome/IconTile.cs`. GitHub's rename detection gives up at that size and reports it as modify/delete. Rebase onto dev and re-apply this one line at the new path, the namespace and the body are otherwise unchanged.

XeldarAlz, line 15: Non-blocking: the two overloads do not center identically, so "draws identically" in the description is a little optimistic. The cursor one centers the `CalcTextSize` line box, the draw-list one centers the glyph ink box via `GlyphPen` (X0/X1, Y0/Y1), so a tile's glyph can shift a pixel or two. Right move, it matches the other 40-odd call sites, but that shift is what the visual pass has to check.

XeldarAlz (changes requested):
Verified on a local merge of this branch into dev: `dotnet build -c Release` 0 warnings, 0 errors, `dotnet test` 1820/1821 (the single failure is SpotlightMath's decimal separator under a non-invariant machine locale, unrelated to this change). Not run in-game. The fix itself is correct and matches the house pattern; two notes inline, one blocking on the path.

src/Aetherphone/Windows/Components/IconTile.cs
BluntEXE: Fixed. Rebased onto current dev, reapplied the one-line change at the new path. Build `-c Release` clean (0 warnings) and `dotnet test` 1959/1959 pass at `032aa8a2`.

BluntEXE: Acknowledged, noted for the visual pass.

PR #144, opened 2026-08-24
feat(shortcuts): allow custom images as shortcut icons

Deldee:
## What

Added the ability to use custom images as shortcut icons,

## Why

Feature requested in discord

Closes #
https://discord.com/channels/1527213298087493732/1541458772764004362
## How I tested it
- Create a new shortcut
- Use a custom image as an icon
- Show on home screen, confirm the image is there
- (fallback) delete the image from the pluginConfig/ShortcutIcons folder, then reload the plugin
- Icon falls back to its glyph

## Checklist

Strings:

Quality:


src/Aetherphone/Core/Shortcuts/ShortcutIconLibrary.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    public IDalamudTextureWrap? Icon(string id)
+    {
+        if (id.Length == 0)
+        {
+            return null;
+        }
+
+        var record = Find(id);
+        return record is null ? null : images.Get(Path.Combine(directory.FullName, record.FileName));
```
XeldarAlz: Blocking: Path.Combine allocates a string every frame for every visible custom icon tile (HomeGridRenderer.cs:139 and :309, FolderOverlay.cs:153, ShortcutsApp.cs:287), and Find is a linear scan beside it. The plugin branch below allocates nothing. Cache the resolved path per id.

src/Aetherphone/Apps/Shortcuts/ShortcutsApp.Editor.cs
```diff
@@ -648,27 +662,36 @@ private void DrawGlyphSection(ShortcutEntry entry, float scale)
             {
                 entry.Glyph = glyph;
                 entry.IconPlugin = string.Empty;
+                entry.IconImage = string.Empty;
```
XeldarAlz, line 676: Blocking: this orphans a committed icon. SetCustomIcon already wrote the JPEG and the config record, so blanking IconImage here drops the only reference and DiscardUnsavedIcon can no longer see it. Release UnsavedIconOf(entry) before clearing.

src/Aetherphone/Apps/Shortcuts/ShortcutsApp.Plugins.cs
```diff
@@ -390,6 +390,7 @@ private void UsePluginIcon(PluginEntry entry)
         if (draft is not null)
         {
             draft.IconPlugin = entry.InternalName;
+            draft.IconImage = string.Empty;
```
XeldarAlz, line 394: Blocking: same orphan as the glyph grid. If the draft holds an unsaved custom icon, clearing IconImage here leaves the file and its CustomShortcutIcons entry unreferenced forever. Release UnsavedIconOf(draft) first.

src/Aetherphone/Apps/Shortcuts/ShortcutsApp.Editor.cs
```diff
@@ -452,6 +459,11 @@ private void DuplicateDraft()
         var copy = draft.Copy();
         copy.Name = CopyName(draft.Name);
         PruneEmptySteps(copy);
+        if (copy.IconImage.Length > 0)
+        {
+            copy.IconImage = store.DuplicateIcon(copy.IconImage) ?? string.Empty;
```
XeldarAlz, line 463: Blocking: on a draft holding an unsaved icon this duplicates it, then draft = copy.Copy() drops the original id with nothing left to release it. Release the pre duplicate IconImage after store.Add(copy).

src/Aetherphone/Core/Shortcuts/CustomShortcutIcon.cs
```diff
@@ -0,0 +1,7 @@
+namespace Aetherphone.Core.Shortcuts;
+
+internal sealed class CustomShortcutIcon
+{
+    public string Id { get; set; } = string.Empty;
+    public string FileName { get; set; } = string.Empty;
```
XeldarAlz: Non-blocking: FileName is always Id plus .jpg, and Duplicate reads Path.GetExtension on it to rediscover a constant. Storing ids alone drops the field and lets the lookup become a set.

src/Aetherphone/Core/Shortcuts/ShortcutIconLibrary.cs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public static byte[] Bake(string sourcePath, WallpaperCrop crop) =>
+        ImageProcessor.BakeSquareJpeg(sourcePath, crop, IconSize).Bytes;
+
+    public string Commit(byte[] bakedBytes)
+    {
+        var id = Guid.NewGuid().ToString("N");
+        var fileName = id + ".jpg";
+        File.WriteAllBytes(Path.Combine(directory.FullName, fileName), bakedBytes);
```
XeldarAlz: Non-blocking: Commit runs from ConsumeBakedIcon, which is called from Draw, so this write plus configuration.Save land on the UI thread. The bake was moved off thread on purpose. Small at 256px, but worth knowing.

src/Aetherphone/Apps/Shortcuts/ShortcutsApp.Editor.cs
```diff
@@ -626,7 +640,7 @@ private void DrawGlyphSection(ShortcutEntry entry, float scale)
             var min = center - new Vector2(tile * 0.5f);
             var max = center + new Vector2(tile * 0.5f);
             var glyph = index == 0 ? 0 : (int)ShortcutPalette.Glyphs[index - 1];
-            var active = !usesPluginIcon && entry.Glyph == glyph;
+            var active = !usesPluginIcon && !usesCustomIcon && entry.Glyph == glyph;
```
XeldarAlz: Non-blocking: with a custom icon set no tile reads as active, so the glyph that shows when the file goes missing is invisible in the editor. Since that fallback is part of the feature, consider still marking the glyph.

src/Aetherphone/Apps/Shortcuts/ShortcutsApp.CustomIcon.cs
```diff
@@ -0,0 +1,128 @@
...
+            try
+            {
+                var bytes = ShortcutIconLibrary.Bake(sourcePath, crop);
+                Interlocked.Exchange(ref pendingIconBytes, bytes);
+            }
+            catch (Exception exception)
+            {
+                AepLog.Warning(exception, "[Shortcuts] failed to bake a custom icon");
+                iconSaving = false;
+                iconFailed = true;
```
XeldarAlz, line 70: Nit: the failure path leaves iconBakeDraft holding the draft reference until the next bake overwrites it. Clear it here so the field stops living past the operation.

src/Aetherphone/Apps/Shortcuts/ShortcutsApp.CustomIcon.cs
```diff
@@ -0,0 +1,128 @@
...
+    private ShortcutEntry? iconBakeDraft;
+
+    private void OpenCustomIconPicker()
+    {
+        iconFailed = false;
+        iconPicker.Open();
+        router.Push(ShortcutsScreen.CustomIcon);
+    }
+
+    private void DrawCustomIconPicker(Rect content, float scale)
```
XeldarAlz: Nit: the scale parameter is never used in this method.

src/Aetherphone/Core/Localization/L.cs
```diff
@@ -375,6 +375,14 @@ internal static class Shortcuts
         public static readonly LocString Symbol = new("shortcuts.symbol", "Symbol");
         public static readonly LocString PluginIcon = new("shortcuts.pluginIcon", "Plugin icon");
         public static readonly LocString PluginIconNone = new("shortcuts.pluginIconNone", "None");
+        public static readonly LocString CustomIcon = new("shortcuts.customIcon", "Custom image");
+        public static readonly LocString CustomIconNone = new("shortcuts.customIconNone", "None");
```
XeldarAlz: Nit: this duplicates shortcuts.pluginIconNone (None) across all nine catalogs. Reusing the existing key keeps the translator surface smaller.

XeldarAlz (changes requested):
Merged onto dev locally: Release build clean, 0 warnings. Tests 1794/1795, the one failure is SpotlightMathTests hitting a decimal separator on my machine, unrelated to this. Localization lockstep is complete, 8 keys in all nine catalogs. The conflict is ShortcutsApp.cs only, from 4a78b953 moving fields above the constructor, so a rebase keeping both sides resolves it. Four blocking findings inline. Not verified in game.

Deldee:
@XeldarAlz CI failed due to issue on the tip of the dev branch, unrelated to this (EncryptionHelp and Muster changes that cannot build right now)

src/Aetherphone/Core/Shortcuts/ShortcutIconLibrary.cs
```diff
@@ -0,0 +1,94 @@
...
+    private readonly Dictionary<string, string> pathsById = new();
+
+    public ShortcutIconLibrary(DirectoryInfo directory, Configuration configuration, WallpaperImageCache images)
+    {
+        this.directory = directory;
+        this.configuration = configuration;
+        this.images = images;
+        directory.Create();
+
+        var ids = configuration.CustomShortcutIconIds;
```
XeldarAlz, line 25: Non-blocking: nothing sweeps unreferenced ids. Pick an image, back out of the editor, then have the plugin unload without OnClosed running, and the file plus its id stay forever. Dropping ids that no shortcut references here would make the release paths self-healing.

XeldarAlz (approved):
Merged onto dev locally at 69857c7d: Release build clean, 0 warnings. Tests 1824/1825, the one failure is SpotlightMathTests hitting a decimal separator on my machine, unrelated. All four blocking findings are fixed, and the icon lifecycle holds across save, cancel, duplicate, glyph swap and delete. CI is red on a stale merge ref: it is pinned to base 149ab85f, where AppRegistry still handed EncryptionHelp to MusterApp, and a re-run reuses that same merge commit, so merge current dev in and push to get a green run. One non-blocking note inline. Not verified in game.

PR #145, opened 2026-08-26
Feat(kindkupo)

Paukky:
## What                                                                                                                                                                                                                                                                                                                                                                                                                       

    Adds KindKupo, an anonymous confession and supportive listening app for Aetherphone, featuring feed browsing, composing confessions with expiry options, thread responses, moderation reporting, conduct gate enforcement, and onboarding tour integration.                                                                                                                                                                   

## Why                                                                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                  
    Provides players with a supportive, anonymous space to share thoughts, vent daily reflections, and receive kind words within Aetherphone.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
Closes # 
## How I tested it                                                                                                                                                                                                                                                                                                                                                                                                            

- **Release & Test Suite**: Executed `dotnet test` and verified all 1,366 unit, conduct gate resolution, and 9-language localization parity tests pass with zero warnings or errors.                                                                                                                                                                                                                                          
- **Debug & Dev Build (`/phonedev`)**:                                                                                                                                                                                                                                                                                                                                                                                        
  - Tested signed-out behavior: shows standard lock empty state (`L.KindKupo.SignInTitle` / `SignInHint`) when no session is active.                                                                                                                                                                                                                                                                                          
  - Tested first-launch conduct gate: verified the ConductGate overlay enforces scrolling to the bottom and a 3-second countdown timer before enabling "I Agree".                                                                                                                                                                                                                                                             
  - Tested home screen navigation: Write and Respond buttons correctly route to compose and response feeds.                                                                                                                                                                                                                                                                                                                   
  - Tested moderation reporting: tapping the flag icon on confessions or reply cards opens `ReportPrompt` and submits reports to `net.Safety`.                                                                                                                                                                                                                                                                                
  - Tested inbox and reply viewing: verified responses populate in thread view and clear the `SocialNotificationService` badge.                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                    
## Backend coordination                                                                                                                                                                                                                                                                                                                                                                                                       

Integrates with `net.Kupo` client endpoints for feed, submission, and responses, as well as `net.Safety.ReportAsync` target types (`kupo_confession` / `kupo_response`). Includes local mock fallbacks during development when endpoints are unreachable.                                                                                                                                                                     

## Checklist                                                                                                                                                                                                                                                                                                                                                                                                                  

Strings:                                                                                                                                                                                                                                                                                                                                                                                                                      

Quality:                                                                                                                                                                                                                                                                                                                                                                                                                      


src/Aetherphone/Apps/KindKupo/KindKupoStore.cs
```diff
@@ -0,0 +1,129 @@
...
+        work.Run("compose confession", async token =>
+        {
+            var created = await client.CreateConfessionAsync(text, expiryDays, token).ConfigureAwait(false);
+
+            var newConfession = created ?? KindKupoMockData.CreateMockConfession(text, authorId: "me");
+
+            confessions = [newConfession, ..confessions];
+            userConfessions = [newConfession, ..userConfessions];
+
+            return true;
```
XeldarAlz: Blocking: this always returns true. When `CreateConfessionAsync` returns null (offline, rate limited, 500), `created ?? CreateMockConfession(...)` inserts a fake and reports success, so Compose.cs:88 clears the draft and pops the router. The user's text is destroyed and they are told it posted.

src/Aetherphone/Apps/KindKupo/KindKupoStore.cs
```diff
@@ -0,0 +1,129 @@
...
+            var created = await client.RespondAsync(confessionId, text, token).ConfigureAwait(false);
+            var newResponse = created ?? KindKupoMockData.CreateMockResponse(confessionId, text, "me");
+
+            // Find the confession in local memory and append the reply
+            var target = confessions.FirstOrDefault(c => c.Id == confessionId)
+                ?? KindKupoMockData.GetConfessions().FirstOrDefault(c => c.Id == confessionId);
+
+            if (target is not null)
+            {
+                target.Responses.Add(newResponse);
```
XeldarAlz: Blocking: `confessions` is always empty, so this falls back to `KindKupoMockData.GetConfessions()` and `target.Responses.Add` mutates a throwaway that is garbage on return. The reply never reaches the feed, the inbox, or the captured confession. It also returns true when `RespondAsync` returned null.

src/Aetherphone/Apps/KindKupo/KindKupoStore.cs
```diff
@@ -0,0 +1,129 @@
...
+                if (page is not null)
+                {
+                    confessions = page.Items;
+                    cursor = page.NextCursor; // Save bookmark for next page
+                }
+            });
+        }
+    public void LoadMore()
+        {
+            // 2. If cursor is null, we're at the end — don't make unnecessary network requests
```
XeldarAlz: Blocking: em dash (U+2014) in this comment fails CI. `.github/workflows/ci.yml:53` greps every tracked file and exits 1. The second hit is `conduct.kindkupo.intro` in ru.json:5081. The numbered explanatory comments in this file also break the no-comments rule.

src/Aetherphone/Apps/KindKupo/KindKupoStore.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            if (page is not null)
+            {
+                userConfessions = page.Items;
+                userCursor = page.NextCursor;
+            }
+            else
+            {
+                // Mock fallback during local development
+                userConfessions = KindKupoMockData.GetConfessions()
```
XeldarAlz: Non-blocking: `Where(...).ToArray()` here and `FirstOrDefault` at 122 are LINQ. `ci.yml:80` only greps for `using System.Linq;` and ImplicitUsings is on, so the guard never sees them. Dead in the same change: `userCursor`, `activeUserId`, `userLoading`, `UserConfessions`, `Refresh`, `LoadMore`, `FetchUserConfessions`, plus two unused usings.

src/Aetherphone/Apps/KindKupo/KindKupoApp.Compose.cs
```diff
@@ -0,0 +1,190 @@
...
+                if (ui.HeaderAction(area, Loc.T(L.KindKupo.Post), canSubmit))
+                {
+                    composeBusy = true;
+                    store.ComposeConfession(draft, activeExpiryTab, success =>
+                    {
+                        composeBusy = false;
+                        if (success)
+                        {
+                            draft = string.Empty;
+                            router.Pop();
```
XeldarAlz: Blocking: `onComplete` runs on StoreWork's background thread, so `router.Pop()` mutates ViewRouter's unsynchronized stack while the frame thread is inside `router.Draw`. `draft = string.Empty` also races the `ref draft` held by ImGui. Set an outcome flag and act on it next frame, as ChirperApp.Compose.cs:35-51 does.

src/Aetherphone/Apps/KindKupo/KindKupoApp.Compose.cs
```diff
@@ -0,0 +1,190 @@
...
+                        }
+                    });
+                }
+
+            ui.Field(String.Empty, "##confessionText", ref draft, maxPostLength, true, area.Max.Y);
```
XeldarAlz: Blocking: `ui.Field` expects an unscaled height but gets `area.Max.Y`, an absolute screen coordinate. `AppSkin.Field` computes `heightUnscaled * scale`, so the field renders hundreds of px past the bottom of the phone. Line 157 divides by scale correctly; do the same here.

src/Aetherphone/Apps/KindKupo/KindKupoApp.Compose.cs
```diff
@@ -0,0 +1,190 @@
...
+                    AppPalettes.KindKupo.MutedInk,
+                    1.0f);
+            }
+        }
+    }
+    private void DrawResponseFeed(Rect area)
+    {
+        var scale = UiScale.Current;
+        var padding = 16f * scale;
+        var feed = KindKupoMockData.GetConfessions();
```
XeldarAlz: Blocking: the feed renders five hardcoded lorem ipsum entries, rebuilt from scratch every frame. `store.Refresh`, `LoadMore` and `FetchUserConfessions` have no callers, so `KupoClient.FeedAsync` is never reached. Each frame also allocates a fresh List plus 5 ConfessionDto and 7 ResponseDto. Profile.cs:65 does the same for the Inbox.

src/Aetherphone/Apps/KindKupo/KindKupoApp.Profile.cs
```diff
@@ -0,0 +1,99 @@
...
+                AppPalettes.KindKupo.Accent, AppSkin.Transparent, 1.1f, Loc.T(L.KindKupo.Inbox),
+                HoverLabelSide.Below))
+        {
+            router.Push(KindKupoRoute.Inbox);
+        }
+    }
+
+    private void DrawInbox(Rect area, string userId)
+    {
+        social.MarkSeen(Id);
```
XeldarAlz, line 54: Non-blocking: `MarkSeen` runs every frame the Inbox is open. It scans the notification array, calls `RemoveSocial` and `ClearServerUnread`, and on advance hits `configuration.Save()` (disk) plus an EnqueueAck. Every other call site fires it once from a lifecycle handler, for example ChirperApp.cs:666.

src/Aetherphone/Apps/KindKupo/KindKupoApp.cs
```diff
@@ -0,0 +1,360 @@
...
+    private readonly KindKupoStore store;
+    private readonly ViewRouter<KindKupoRoute> router;
+    private readonly RouterDraw<KindKupoRoute> drawView;
+    private readonly AppSkin ui = new(AppPalettes.KindKupo);
+    private readonly AethernetSession session;
+    private readonly AethernetApi net;
+    private readonly ConductGateService conduct;
+    private readonly SocialNotificationService social;
+    private readonly ReportService report;
+    RichTextLayout? bodyLayout = null;
```
XeldarAlz: Non-blocking: `bodyLayout` is never assigned, so `bodyLayout?.Size.Y ?? Measure...` (150, 242) always measures and `bodyLayout is null` (161, 251) is always true: a full CalcTextSize wrap per card per frame. Assigning it is not the fix either, since one field shared across all cards mis-sizes every card after the first. It also needs an accessibility keyword.

src/Aetherphone/Apps/KindKupo/KindKupoApp.cs
```diff
@@ -0,0 +1,360 @@
...
+
+        var stamp = TimeText.Ago(confession.CreatedAt);
+        var stampSize = Typography.Measure(stamp, 0.85f, FontWeight.Regular);
+        var stampPos = new Vector2(left, centerY - stampSize.Y * 0.5f);
+        Typography.Draw(drawList, stampPos, stamp, AppPalettes.KindKupo.MutedInk, 0.85f, FontWeight.Regular);
+
+        var iconSize = 16f * scale;
+        var transparent = Vector4.Zero;
+
+        switch (router.Current.Screen)
```
XeldarAlz: Non-blocking: this switches on the global `router.Current.Screen` rather than the route being drawn. ViewRouter renders both layers during a transition, so pushing Respond strips the Respond and Report buttons off every outgoing card, and popping ResponseList draws a live button mid-slide. Pass the route in.

src/Aetherphone/Core/Aethernet/Clients/KupoClient.cs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+
+
+    public Task<ConfessionDto?> CreateConfessionAsync(string text, int expiryDays, CancellationToken token,
+        Action<int>? statusSink = null)
+    {
+        var now = DateTime.UtcNow;
+        DateTime? expiresAt;
+        switch (expiryDays)
```
XeldarAlz: Blocking: `expiryDays` is really a SegmentStrip tab index, so 2 means 3 days and 3 means 7. A caller trusting the name and passing 7 hits `default` and posts a confession that never expires. Take a real duration here and map the tab index at the call site.

src/Aetherphone/Core/Aethernet/Clients/KupoClient.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            path += $"?cursor={Uri.EscapeDataString(cursor)}";
+        }
+
+        return net.GetAsync(path, AethernetJsonContext.Default.ConfessionPage, token);
+    }
+
+    public Task<ConfessionPage?> MyConfessionsAsync(string userId, string? cursor, CancellationToken token)
+    {
+        var path = $"/kupo/{userId}/confessions";
```
XeldarAlz: Non-blocking: `userId` is interpolated unescaped, unlike every other method in this class. A '/', '?' or '#' rewrites the request target. The shape also disagrees with `InboxAsync`'s `/kupo/users/{id}/confessions` for the same resource, so one of the two cannot match the backend.

src/Aetherphone/Core/Apps/AppRegistry.cs
```diff
@@ -66,6 +67,7 @@ public static AppBundle BuildDefault(PhoneServices services, VideoPlayer video,
...
         apps.Add(new VelvetShell(services.AethernetSession, new AethernetApi(services.Http, services.AethernetSession, "velvet"), services.Lodestone, services.Configuration, photoLibrary, services.Http, services.RemoteImages, services.Notifications, services.VelvetLauncher, services.SocialLauncher, services.GameData, services.SocialNotifications, services.KeyVault, services.ConversationKeys, services.Visibility, services.RealtimeSignals, services.WallpaperImages, services.Confirm, services.Report, services.Conduct, services.Installer));
+        apps.Add(new KindKupoApp(services.AethernetSession, new AethernetApi(services.Http, services.AethernetSession, "kindkupo"), services.Confirm, services.Report, services.Conduct, services.SocialNotifications));
```
XeldarAlz: Blocking: 'kindkupo' is missing from `SuspensionGate.GatedAppIds` (Core/Moderation/SuspensionGate.cs:10), so a banned user can still open this app and post confessions and replies. Anonymous UGC with report flows is exactly the surface that lockout exists to close.

src/Aetherphone/Core/Conduct/ConductRules.cs
```diff
@@ -223,7 +223,31 @@ internal static class ConductRules
         },
     };
 
-    private static readonly ConductGate[] All = { Chirper, Aethergram, Velvet, Muster, YellowPages, Casino, Coin };
+    public static readonly ConductGate KindKupo = new()
+    {
+        AppId = "kindkupo",
```
XeldarAlz, line 230: Non-blocking: two sibling registries were missed. `AppStoreCatalog` has no kindkupo entry, so the app lists under Tools with the fallback subtitle instead of Social. `NotificationChannels` has none either, while `BadgeCount` reads `social.UnseenCount(Id)`, so the badge can never light and the app cannot be muted.

src/Aetherphone/Localization/de.json
```diff
@@ -3232,6 +3232,17 @@
   "onboarding.aetherStreamPartyBody": "Starte eine Party, und Leute in der Nähe können um Zutritt bitten. Du entscheidest, wer reinkommt, und alle bleiben auf derselben Sekunde desselben Videos.",
   "onboarding.aetherStreamSettingsTitle": "Auf deinen Rechner abstimmen",
   "onboarding.aetherStreamSettingsBody": "Maximale Qualität, Hardware-Dekodierung, ob andere deinen Stream finden, und die Komponenten, die der Player braucht, alles hinter diesem Zahnrad.",
+  "onboarding.kindkupoBody": "A place to share anonymous confessions and receive kind, supportive responses from the community.",
+  "onboarding.kindkupoStatsTitle": "Your activity",
+  "onboarding.kindkupoStatsBody": "Track how many confessions you've posted, the responses you've received, and kudos given by others.",
+  "onboarding.kindkupoWriteTitle": "Share what's on your mind",
+  "onboarding.kindkupoWriteBody": "Post an anonymous confession and choose how long before it expires, from 24 hours to forever.",
```
XeldarAlz: Blocking: the eleven `onboarding.kindkupo*` strings are literal English in all nine catalogs, while the neighbouring `kindkupo.*` block is translated. Parity tests only check key presence, so CI stays green and an English tour ships to every locale. Separately `conduct.kindkupo.title` is 'KindKupo Guidelines' in L.cs:5891 but 'Community Rules' in en.json and all eight translations.

XeldarAlz (changes requested):
Build: clean, 0 warnings. Tests: 1366 passed. Not tested in game.

Nine blocking and six non-blocking findings inline. The headline is that the app does not work end to end yet: the feed and inbox render mock data and the store's network paths have no callers. CI is also red on an em dash.

Paukky:
Testing update:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       • Verified core app flows (posting confessions, viewing responses, composing replies, feed navigation, and draft reset lifecycle). Everything works as expected.                                                                                                                                                                                                                                                             
• Tested the reporting UI/dialog for both confessions and replies. The prompts trigger and submit properly, though backend receipt was not directly verifiable from the client side.

src/Aetherphone/Core/Apps/AppRegistry.cs
```diff
@@ -68,6 +69,7 @@ public static AppBundle BuildDefault(PhoneServices services, VideoPlayer video,
...
         apps.Add(new VelvetShell(services.AethernetSession, new AethernetApi(services.Http, services.AethernetSession, "velvet"), services.Lodestone, services.Configuration, photoLibrary, services.Http, services.RemoteImages, services.Notifications, services.VelvetLauncher, services.SocialLauncher, services.GameData, services.SocialNotifications, services.KeyVault, services.ConversationKeys, services.ChatHistory, services.Visibility, services.RealtimeSignals, services.WallpaperImages, services.Confirm, services.Translation, services.Report, services.Conduct, services.Installer, services.EncryptionHelp));
+        apps.Add(new KindKupoApp(services.AethernetSession, new AethernetApi(services.Http, services.AethernetSession, "kindkupo"), services.Report, services.Conduct, services.SocialNotifications));
```
XeldarAlz, line 72: Blocking: the server knows this app as `kupo` (AppCatalog.cs:20, DefaultEnabled false) and every `/kupo/*` route sits behind `RequireAppEnabled("kupo")`. AppAvailability.cs:52 looks up `kindkupo`, misses, and falls back to enabled. So the app ships visible on every phone while all of its endpoints 503, and the admin kill switch can never hide it. Align the two ids.

src/Aetherphone/Core/Notifications/NotificationChannels.cs
```diff
@@ -16,6 +16,7 @@ internal static class NotificationChannels
         new("chirper", L.Apps.Chirper, AppAccents.For("chirper")),
         new("aethergram", L.Apps.Aethergram, AppAccents.For("aethergram")),
         new("velvet", L.Apps.Velvet, AppAccents.For("velvet")),
+        new("kindkupo", L.Apps.KindKupo, AppAccents.For("kindkupo")),
```
XeldarAlz, line 19: Blocking: the server sends `KupoResponse = 22`, but SocialActivity.cs stops at 21, so `Body()` returns empty and SocialNotificationService.Present bails on the empty body. A reply lights this badge and produces no banner and no notification center entry. Add the type with an anonymous title, plus a `kindkupo` branch in NotificationRouter.Open for the inbox.

src/Aetherphone/Apps/KindKupo/KindKupoApp.cs
```diff
@@ -0,0 +1,371 @@
...
+                break;
+            case KindKupoScreen.Inbox:
+                DrawInbox(area);
+                break;
+            case KindKupoScreen.Respond:
+                DrawResponseFeed(area);
+                break;
+            case KindKupoScreen.ResponseList:
+            {
+                var confession = ResolveConfession(route);
```
XeldarAlz, line 116: Non-blocking: when `FindConfession` misses, for example after `Refresh` replaced the feed array, this route draws only `ui.Body`: no header, no back button, a blank screen the app's own navigation cannot leave. Draw the header plus an empty state, or pop the route.

src/Aetherphone/Apps/KindKupo/KindKupoApp.Compose.cs
```diff
@@ -0,0 +1,263 @@
...
+    {
+        1 => DateTimeOffset.UtcNow.AddDays(1).ToUnixTimeSeconds(),
+        2 => DateTimeOffset.UtcNow.AddDays(3).ToUnixTimeSeconds(),
+        3 => DateTimeOffset.UtcNow.AddDays(7).ToUnixTimeSeconds(),
+        _ => 0L,
+    };
+
+    private void DrawWriteScreen(Rect area)
+    {
+        if (composeOutcome == 1)
```
XeldarAlz, line 44: Non-blocking: `composeOutcome` is only consumed while this screen draws, and neither OnOpened nor OnClosed clears it. Press Post then Back before the request lands and the flag survives, so the next Write pops itself on the frame it opens. Reset both outcome flags next to `draft`.

src/Aetherphone/Apps/KindKupo/KindKupoStore.cs
```diff
@@ -0,0 +1,195 @@
...
+    {
+        work.Run("compose confession", async token =>
+        {
+            var created = await client.CreateConfessionAsync(text, expiresAtUnix, token).ConfigureAwait(false);
+            if (created is null)
+            {
+                return false;
+            }
+
+            confessions = [created, ..confessions];
```
XeldarAlz, line 125: Non-blocking: the browse feed deliberately excludes your own confessions (the server filters `AuthorId != me.Id`), so prepending here puts your own post into Respond until the next Refresh, where tapping Respond 400s with "You cannot reply to your own confession". Only `userConfessions` should take it.

src/Aetherphone/Core/Aethernet/Clients/KupoClient.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    public Task<ConfessionDto?> CreateConfessionAsync(string text, long expiresAtUnix, CancellationToken token,
+        Action<int>? statusSink = null)
+    {
+        return net.PostAsync("/kupo/confessions", new CreateConfessionRequest(text, expiresAtUnix),
+            AethernetJsonContext.Default.CreateConfessionRequest,
+            AethernetJsonContext.Default.ConfessionDto, token, statusSink);
+    }
+
+    public Task<ConfessionDto?> GetConfessionAsync(string confessionId, CancellationToken token)
```
XeldarAlz, line 22: Non-blocking: eight of the twelve methods here have no caller (GetConfession, DeleteConfession, Responses, DeleteResponse, SendKudos, LikeResponse, Stats, UserStats). Same for `HasMoreUserConfessions` and nine LocStrings shipped in all nine catalogs: kindkupo.feed, .rules, .expiryDate, .written, .kudos, .rulesTitle, .rulesBody, onboarding.kindkupoStatsTitle and StatsBody.

src/Aetherphone/Apps/KindKupo/KindKupoApp.Profile.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            return;
+        }
+
+        var body = new Rect(new Vector2(area.Min.X + padding, top), new Vector2(area.Max.X - padding, area.Max.Y));
+        using (AppSurface.Begin(body))
+        {
+            ImGui.Dummy(new Vector2(0f, 4f * scale));
+            for (var index = 0; index < confessions.Length; index++)
```
XeldarAlz, line 88: Non-blocking: the inbox stops at the server's 30 row page. `store.HasMoreUserConfessions` has no reader and there is no load-more here, so once a user passes 30 confessions the older ones and their replies are unreachable.

src/Aetherphone/Apps/KindKupo/KindKupoApp.cs
```diff
@@ -0,0 +1,371 @@
...
+                        AppPalettes.KindKupo.MutedInk, transparent, 1f, Loc.T(L.KindKupo.ViewReplies, confession.ResponseCount)))
+                {
+                    router.Push(KindKupoRoute.ViewResponse(confession.Id));
+                }
+                break;
+            }
+        }
+    }
+
+    private void DrawResponseCard(ConfessionResponseDto response)
```
XeldarAlz, line 235: Non-blocking: this repeats DrawConfessionCard almost line for line (pad, rounding, wrap measure, footer center, Dummy advance). One card body with the footer passed in stops the two from drifting apart.

src/Aetherphone/Apps/KindKupo/KindKupoApp.cs
```diff
@@ -0,0 +1,371 @@
...
+        router = new ViewRouter<KindKupoRoute>(KindKupoRoute.Home);
+        back = () => router.Pop();
+    }
+    public void Dispose() => store.Dispose();
+    public void OnOpened()
+    {
+        router.Reset();
+        draft = string.Empty;
+        replyDraft = string.Empty;
+        conduct.NotifyAppOpened(Id);
```
XeldarAlz, line 64: Nit: PhoneShell.cs:86 already wires `navigation.AppOpened` to `NotifyAppOpened`, so the gate is asked to open twice. No other app does this locally.

src/Aetherphone/Apps/KindKupo/KindKupoApp.Compose.cs
```diff
@@ -0,0 +1,263 @@
...
+
+namespace Aetherphone.Apps.KindKupo;
+
+internal sealed partial class KindKupoApp
+{
+    private const int MaxPostLength = 1000;
+
+    private static readonly List<LocString> ExpiryTabOptions =
+    [
+        L.KindKupo.ExpiryNever,
```
XeldarAlz, line 20: Nit: English shortens this to "Nvr" to fit the 22px segment, but the catalogs put much longer words in the same slot, where SegmentStrip hands them to Marquee and they read as truncated. Widen the strip or drop the Never tab.

XeldarAlz (changes requested):
Build: clean, 0 warnings. Tests: 1940 passed. Not tested in game.

All fifteen findings from the last pass are fixed, localization included: 68 KindKupo keys, present in all nine catalogs, no L.cs to en.json drift. Two new blocking findings inline, both in the plumbing around the app rather than in the app itself.

src/Aetherphone/Localization/zh.json
```diff
@@ -5634,6 +5677,32 @@
   "conduct.coin.scams.lead": "硬币永远不会离开你的账号，所以任何交易提议都是诈骗。请远离：",
   "conduct.coin.scams.1": "任何出售硬币、充值或余额翻倍服务的人",
   "conduct.coin.scams.2": "要求你到别处登录或分享验证码的链接或私信",
+  "conduct.coin.scams.3": "先要金币或物品的抽奖活动",
```
XeldarAlz, line 5680: Blocking: `conduct.coin.scams.3` already exists at line 5706 with a different Chinese string, so this branch adds a second copy of a key it does not own. The JSON parser keeps one and silently drops the other, and LocalizationParityTests only checks presence, so CI stays green either way. Looks like merge debris, delete this line.

src/Aetherphone/Apps/KindKupo/KindKupoApp.cs
```diff
@@ -0,0 +1,371 @@
...
+            case KindKupoScreen.Respond:
+            {
+                var respondPos = new Vector2(right - iconSize * 0.8f, centerY);
+                if (ui.IconButton(respondPos, iconSize, FontAwesomeIcon.Pen.ToIconString(),
+                        AppPalettes.KindKupo.MutedInk, transparent, 1f, Loc.T(L.KindKupo.Respond)))
+                {
+                    router.Push(KindKupoRoute.ComposeResponse(confession.Id));
+                }
+
+                var reportPos = new Vector2(respondPos.X - 24f * scale, centerY);
```
XeldarAlz, line 213: Non-blocking: `AppSkin.IconButton` treats its second argument as a hit radius, so both buttons are 32*scale wide but their centers are only 24*scale apart. In the 8*scale overlap the report button registers its pendingTap last and wins the release, so tapping the left edge of the pen opens the report prompt. The respond box also reaches 3.2*scale past the card edge.

src/Aetherphone/Apps/KindKupo/KindKupoApp.Compose.cs
```diff
@@ -0,0 +1,263 @@
...
+            store.ComposeConfession(draft, ExpiryFromTab(activeExpiryTab), success =>
+            {
+                composeBusy = false;
+                composeOutcome = success ? 1 : 2;
+            });
+        }
+
+        var fieldTop = ImGui.GetCursorScreenPos().Y;
+        var fieldHeight = MathF.Max(120f, (area.Max.Y - fieldTop - 44f * scale) / scale);
+        ui.Field(string.Empty, "##confessionText", ref draft, MaxPostLength, true, fieldHeight);
```
XeldarAlz, line 121: Non-blocking: `AppSkin.Field` runs `Typography.Plain(label)` unconditionally, and an empty string still advances a full line plus spacing. `fieldHeight` was measured from the cursor before that line, so the box starts one line lower and its bottom runs into the 44px counter reserve. Pass a real label, or reserve the extra line. Line 209 has the same shape inside the AppSurface child.

src/Aetherphone/Apps/KindKupo/KindKupoStore.cs
```diff
@@ -0,0 +1,195 @@
...
+        }, () =>
+        {
+            if (string.Equals(activeUserId, userId, StringComparison.Ordinal))
+            {
+                userLoading = false;
+            }
+        });
+    }
+
+    public void ComposeConfession(string text, long expiresAtUnix, Action<bool> onComplete)
```
XeldarAlz, line 115: Non-blocking: `CreateConfessionAsync` and `RespondAsync` both take a `statusSink` that is never passed, so 429, 403 from SuspensionGate and a moderation rejection all surface as `L.Account.CannotReach`. A rate limited or suspended user is told they are offline and retries. Thread the status through the way Chirper threads its compose failure reason.

XeldarAlz (commented):
Second pass on the same head, five more findings inline. The first is a duplicate JSON key this branch adds to zh.json, which is why that catalog is +69 while the other eight are +68.

src/Aetherphone/Core/Notifications/NotificationChannels.cs
XeldarAlz, line 19: Adding to the type 22 comment above: `SocialNotificationService.ServedApps` (line 19) is still the five ids chirper, aethergram, velvet, yellowpages, message. `OnFrameworkTick` returns early unless one of those is installed, so on a phone with KindKupo but none of the five, nothing polls at all and this channel controls a badge that can never light. Add `kindkupo` there too.

PR #146, opened 2026-08-27
feat(localization): polish Simplified Chinese translation

Nero0421:
This PR polishes the Simplified Chinese (zh.json) translation on top of the
current dev tree. 143 values change; no keys are added, removed, or renamed,
and no other locale is touched.

What changed

- Phone app names: Message becomes 以太信, MogCast becomes 莫古影院, Yellow
  Pages becomes 人才市场, Aether Coin becomes 以太积分, and Gamba becomes
  以太赌场. These follow the transliteration precedent set by Chirper (叽叽)
  and Aethergram (以太图集). Linkpearl, Velvet, and Muster keep their English
  names, per docs/localization.md.
- Eorzean units: the Health app's unit picker now transliterates the remaining
  units the way yalm (码尔姆) already was: ilm 伊尔姆, fulm 法尔姆, malm 玛尔姆,
  ponz 庞兹.
- Terminology: 支援 becomes 支持, real money is 现实货币 with in-game gil
  spelled out as 游戏内金币, hunt train 狩猎车, treasure hunt 挖宝,
  portrait/landscape 竖向/横向, Reference 错误代码, Venue nights 场馆之夜,
  FC chest 部队柜, and the chocobo saddlebag 陆行鸟鞍囊.
- Wording: touch-ups across the conduct rules, moderation removal notices,
  barkeep grades (完美/棒/粗糙/失手), and assorted app strings.

Copy rules

- No em dashes were introduced, and every placeholder in the English source is
  preserved in each changed string (both checked programmatically).

Happy to adjust wording, revert any of the renames, or split the PR if you
would rather review the app renames separately.


src/Aetherphone/Localization/zh.json
```diff
@@ -221,12 +221,12 @@
   "common.live": "LIVE",
   "common.hq": "HQ",
   "common.nq": "NQ",
-  "common.openInBrowser": "点击在浏览器中打开",
+  "common.openInBrowser": "在浏览器中打开它",
```
XeldarAlz: Blocking: the source is "Click to open in browser" (L.cs:29). The new text drops the click affordance, which is the whole point of the string. 点击在浏览器中打开 was right.

src/Aetherphone/Localization/zh.json
```diff
@@ -4106,18 +4106,18 @@
...
   "settings.importScreenshots": "导入截图",
   "settings.importScreenshotsHint": "把你拍下的截图复制到“照片”图库，包括 ReShade 和 GShade 的截图。只复制手机运行期间拍摄的截图，原文件保持不动。",
-  "social.roleManagement": "管理层",
+  "social.roleManagement": "管理组",
   "social.roleDeveloper": "开发者",
   "social.roleModerator": "版主",
   "social.roleVerified": "已认证",
   "social.rolePatreon": "Patreon 会员",
-  "social.roleSupport": "支援",
+  "social.roleSupport": "支持",
```
XeldarAlz: Non-blocking: this is a staff badge worn next to a name, where 支持 reads as "supporter". It also collides with 支持 in settings.supportHint, which means the opposite thing. 支援 kept the two apart.

src/Aetherphone/Localization/zh.json
```diff
@@ -483,7 +483,7 @@
   "message.tapToRemove": "点击移除",
   "message.draftPrefix": "草稿:",
   "message.presenceOnline": "在线",
-  "message.presenceLastSeen": "最后上线 {0}",
+  "message.presenceLastSeen": "最后在线 {0}",
```
XeldarAlz: Non-blocking: the presence label becomes 最后在线 but settings.lastSeenOnline and settings.chatPrivacyHint still say 最后上线, and those are the switch that controls this label. Please pick one and apply it to all three.

XeldarAlz (commented):
Disregard my earlier notes here, I have removed the line comments. They were nits and not worth your time. This is good work: the mechanics are clean (same 5816 keys in the same order, every placeholder preserved, no em dashes) and the terminology pass reads better than what we ship today.

XeldarAlz (commented):
One real question before I merge, and I would rather have your read than mine.

以太币 is also the common Chinese name for Ether, Ethereum's coin. Inside an FFXIV phone app I would expect most players to parse 以太 as aether first, so this may well be a non-issue. But Aether Coin ships with a conduct page whose whole point is that it has no real world value and can never be cashed out, and I do not want the name pulling against that.

Does 以太币 read as crypto to you, or does the context settle it? If it does read that way, is there a form you would prefer?

Everything else looks good to me.

Nero0421:
Fair point, and after sitting with it I think you are right. 以太币 is exactly how Chinese crypto communities write ETH (it sits right next to 比特币), so even with the in-app aether context, the wallet UI and the no-real-value disclaimer would keep pulling toward a crypto read. Not worth it for a currency whose whole point is that it is not that.

I'll switch the name to 以太积分 (aether points): "+{0} 以太积分", "以太积分余额有变动", and the conduct line becomes "以太积分是没有现实价值的装饰性货币". Points fits the earn-by-playing, spend-on-cosmetics loop better than coin, and it carries zero crypto echo.

One more naming note while we are here: I translated Yellow Pages as 人才市场 rather than 黄页. 黄页 is the literal rendering of "yellow pages", the old physical phone directories, which younger players mostly do not recognize anymore; 人才市场 reads as a familiar local listings market to that audience. Happy to reconsider it too if you see it differently.

I'll update the branch and push the revision in a moment. Thanks for the close read.@XeldarAlz

PR #152, opened 2026-09-01
feat(UI) Add globally managed badge setting to apps

Deldee:
## What
Updated the badge architecture to be manageable from the setting page : 
- Apps can opt into this management by setting hasBadge => true in their declaration (default false)
- The Notification setting page was renamed to "Notification and Badges" and now has a section to disable badges
- If only notification, or only badges are enabled, the setting list shows it directly
- Apps with no badge, or disabled from the setting page will not show any badges on the homeView
- The three app with their own badge management (activities, wallet, dailies) all had their config migrated once, then rely on the new configuration
- All other app with badges has hasBagde => true set

## Why
3 apps currently had their own management of badges, and badges in general were detached from notification,this aims to bring them more in line, without merging the two distinct behavior into one, less modular one

Closes #

## How I tested it
- Build an old version of the plugin
- Set the configuration in the wallet app for badge
- Rebuild the plugin under the new version
- Ensure that the wallet settings are still the same

## Screenshots

image

## Checklist

Strings:

Quality:


src/Aetherphone/Apps/Notifications/NotificationsApp.cs
```diff
@@ -19,6 +19,7 @@ internal sealed class NotificationsApp : IPhoneApp
     public string DisplayName => Loc.T(L.Apps.Notifications);
     public string Glyph => "N";
     public int BadgeCount => notifications.UnreadCount;
+    public bool HasBadge => true;
```
XeldarAlz, line 22: Blocking: only the home tile honors this toggle. MinimizedPhone.RefreshBadge (MinimizedPhone.cs:716) and Plugin.UpdateDtrBadge (Plugin.cs:465) still read notifications.UnreadCount directly, so the count stays on the capsule and in the DTR bar after the user turns the badge off.

src/Aetherphone/Apps/Wallet/WalletApp.cs
```diff
@@ -25,7 +25,7 @@ public int BadgeCount
     {
         get
         {
-            if (!configuration.ShowWalletBadge)
+            if (!configuration.IsAppBadgeEnabled(Id))
```
XeldarAlz: Blocking: Wallet is double-gated now. HasBadge is true, so HomeTileView.IsBadgeVisible already applies the preference, and docs/app-framework.md:82 (added in this PR) says to return the raw count. Activity and Dailies dropped their gate; drop this one too, or the folder sum reads 0 for Wallet alone.

src/Aetherphone.Tests/ConfigurationBadgeMigrationTests.cs
```diff
@@ -0,0 +1,33 @@
...
+    public void LegacyBadgeFlagsDeserializeFromTheirPreRenameJsonKeys()
+    {
+        const string json = """{"ShowWalletBadge":false,"ShowDailiesBadge":false,"ShowActivityBadge":true}""";
+
+        var configuration = JsonConvert.DeserializeObject<Configuration>(json);
+
+        Assert.NotNull(configuration);
+        Assert.False(configuration!.LegacyShowWalletBadge);
+        Assert.False(configuration.LegacyShowDailiesBadge);
+        Assert.True(configuration.LegacyShowActivityBadge);
```
XeldarAlz, line 18: Blocking: these asserts stop at deserialization and never call MigrateBadgeSettings, so the three id literals it writes (`wallet`, `dailies`, `character`) are untested. `character` for ActivityApp is the easy typo, and getting it wrong silently restores a hidden badge with BadgeSettingsMigrated already true.

src/Aetherphone/Core/Apps/AppRegistry.cs
```diff
@@ -101,7 +101,7 @@ public static AppBundle BuildDefault(PhoneServices services, VideoPlayer video,
         apps.Add(new GamesApp(services.GameStats, services.GameData, services.Textures, services.Coins,
             services.CoinSessions, services.GameRooms));
         apps.Add(new NotificationsApp(services.Notifications, services.SocialNotifications, services.LinkpearlLauncher, services.VelvetLauncher, services.DmLauncher, services.GramDmLauncher, services.SocialLauncher, services.MusterLauncher, services.YellowPagesLauncher, services.AnnouncementsLauncher, services.SafetyLauncher, services.EncryptionSetup, services.RadioLauncher, services.CasinoLauncher, services.AetherStreamLauncher, services.HuntsLauncher));
-        apps.Add(new SettingsApp(services, photoLibrary));
+        apps.Add(new SettingsApp(services, photoLibrary, apps));
```
XeldarAlz: Non-blocking: SettingsApp receives `apps` while it is still being filled, since Calendar, Coin, Casino, AppStore, Housing and Hunts are added after this line. It works only because the List reference stays live. Swapping in a snapshot later drops those apps from Settings > Notifications and Badges with no compile error.

src/Aetherphone/Windows/Components/Chrome/HomeTileView.cs
```diff
@@ -31,12 +31,14 @@ public static void DrawApp(Vector2 center, float size, IPhoneApp app, PhoneTheme
...
 
         DrawLabel(center, size, app.DisplayName, theme, scale, labelAlpha, showLabels, labelWidth, zoom);
-        if (app.BadgeCount > 0)
+        if (app.BadgeCount > 0 && IsBadgeVisible(app))
         {
             DrawBadge(center, size, app.BadgeCount, app.BadgeAsDot, theme, scale);
         }
     }
 
+    private static bool IsBadgeVisible(IPhoneApp app) => !app.HasBadge || Plugin.Cfg.IsAppBadgeEnabled(app.Id);
```
XeldarAlz: Non-blocking: this is the first Plugin dependency in HomeTileView, and Plugin.Cfg is `null!` until the constructor assigns it. GameComposer.cs:238 and GameEmojiComposer.cs:34 both guard it with `Plugin.Cfg?.`. Threading configuration through DrawApp keeps the gate non-null and testable.

src/Aetherphone/Apps/Settings/Pages/AppNotificationPage.cs
```diff
@@ -11,67 +12,106 @@ namespace Aetherphone.Apps.Settings.Pages;
 
 internal sealed class AppNotificationPage : ISettingsPage
 {
-    public string Title => Loc.T(channel.Name);
+    public string Title => entry.Name;
```
XeldarAlz: Non-blocking: entry.Name is a Loc.T result resolved once in NotificationsPage.EnsureEntries and copied here by Show(). NotificationsPage rebuilds on a language change but this copy is not refreshed, so the header keeps the old language until the user navigates out and back in.

src/Aetherphone/Apps/Settings/Pages/AppNotificationPage.cs
```diff
@@ -11,67 +12,106 @@ namespace Aetherphone.Apps.Settings.Pages;
...
     public string Summary => string.Empty;
     public FontAwesomeIcon Icon => FontAwesomeIcon.Bell;
-    public Vector4 Tint => channel.Accent;
+    public Vector4 Tint => entry.Accent;
     private readonly Configuration configuration;
     private readonly SoundService sound;
-    private NotificationChannel channel = NotificationChannels.All[0];
+    private AppSettingsEntry entry = new(string.Empty, string.Empty, default, false, false);
```
XeldarAlz: Nit: this seed gives Title an empty string and Tint a fully transparent Vector4. The old code seeded from NotificationChannels.All[0], so it always had a real name and accent if it ever drew before Show().

docs/app-framework.md
```diff
@@ -237,7 +239,7 @@ Accent colors have exactly one source: `AppAccents.For(id)` (src/Aetherphone/Cor
 
 ## Badges
 
-`HomeTileView.DrawApp` reads `BadgeCount` (and `BadgeAsDot`) every frame the home screen is visible and draws `AppBadge` in the tile corner when the count is positive. Folder tiles sum the numeric counts of the apps inside; a dot is shown only if no numeric badge exists but some member wants a dot.
+`HomeTileView.DrawApp` (src/Aetherphone/Windows/Components/Chrome/HomeTileView.cs) reads `BadgeCount` (and `BadgeAsDot`) every frame the home screen is visible and draws `AppBadge` in the tile corner when the count is positive and the app's badge is not turned off (`HasBadge`/`Configuration.IsAppBadgeEnabled`, see [Notifications](notifications.md#hiding-a-badge)). Folder tiles sum the numeric counts of the apps inside; a dot is shown only if no numeric badge exists but some member wants a dot.
```
XeldarAlz, line 242: Non-blocking: line 244 just below still says `DailiesApp` gates its count behind a configuration toggle. This PR removed that gate (DailiesApp.cs:44 is the bare count now), and it contradicts the new rule on line 82. Worth swapping the example.

docs/app-framework.md
```diff
@@ -80,7 +81,8 @@ Members with a `=>` body are C# default interface implementations: you only over
...
 | `Glyph` | One- or two-character fallback text drawn on the tile when no icon texture exists for `Id`. `HomeTileView.DrawApp` (src/Aetherphone/Windows/Components/HomeTileView.cs) tries `AppIconArt.TryDraw` first and falls back to the glyph. |
 | `Accent` | Tile and highlight color. The default delegates to `AppAccents.For(Id)`; keep it that way and add your color to the `AppAccents` table instead of hardcoding one. Unknown ids get a gray fallback. |
-| `BadgeCount` | Unread count shown on the home tile. Read every frame; return a cached field, never compute or allocate here. `0` means no badge. |
+| `BadgeCount` | Unread count shown on the home tile. Read every frame; return a cached field, never compute or allocate here. `0` means no badge. Always return the raw count here; the user's on/off preference is applied centrally, not by this getter. |
+| `HasBadge` | Opts the app into the shared, user-toggleable badge switch (`Configuration.BadgeSettings`, default on) and into a "Show badge" row on that app's Settings > Notifications and Badges page. Default `false`, so `BadgeCount => 0` apps need not override it. See [Notifications](notifications.md#hiding-a-badge). |
```
XeldarAlz, line 85: Non-blocking: docs/creating-an-app.md was not updated to match. Its defaults table (line 69) and its Badge count bullet (line 220) never mention HasBadge, so an author following that guide ships an app whose badge cannot be turned off and never shows up in Settings.

src/Aetherphone/Configuration.cs
```diff
@@ -722,6 +725,40 @@ public bool ShouldShowNotificationBanner(string appId) =>
 
     public string ResolveNotificationToken(string appId) => AppSoundOverride(appId) ?? NotificationSound;
 
+    public bool IsAppBadgeEnabled(string appId) =>
+        !BadgeSettings.TryGetValue(appId, out var enabled) || enabled;
+
+    public void SetAppBadgeEnabled(string appId, bool enabled)
```
XeldarAlz, line 740: Non-blocking: unlike AcknowledgeConduct and MarkChangelogSeen nearby, this setter does not Save(), so every caller has to remember it. AppNotificationPage already pairs the two calls by hand, and the next caller that forgets loses the preference on restart.

src/Aetherphone/Configuration.cs
```diff
@@ -96,9 +97,11 @@ internal sealed class Configuration : IPluginConfiguration, IHomeConfiguration,
     public bool NotifyWeeklyReset { get; set; }
     public bool NotifyGrandCompanyReset { get; set; }
     public bool NotifyRetainerVentures { get; set; }
-    public bool ShowWalletBadge { get; set; } = true;
-    public bool ShowDailiesBadge { get; set; } = true;
-    public bool ShowActivityBadge { get; set; } = true;
+    [JsonProperty("ShowWalletBadge")] public bool LegacyShowWalletBadge { get; set; } = true;
```
XeldarAlz: Nit: attributes go on their own line per CLAUDE.md. These three are the only attribute-decorated properties in the file, so the single-line form has no precedent here.

src/Aetherphone/Apps/Settings/Pages/NotificationsPage.cs
```diff
@@ -60,63 +62,78 @@ public void Draw(in PhoneContext context, Rect body)
...
+        var count = 0;
+        for (var index = 0; index < apps.Count; index++)
         {
-            var primary = Loc.Culture.CompareInfo.Compare(Loc.T(left.Name), Loc.T(right.Name),
-                CompareOptions.IgnoreCase);
-            return primary != 0 ? primary : string.CompareOrdinal(left.AppId, right.AppId);
-        });
-        sortedLanguage = Loc.Current;
-    }
+            if (Qualifies(apps[index]))
```
XeldarAlz: Nit: Qualifies() runs over the whole app list twice, each call doing a linear NotificationChannels.Contains scan, and line 116 does a third lookup for HasChannel. One pass into a List that keeps the Contains result removes both.

src/Aetherphone/Apps/Settings/Pages/NotificationsPage.cs
```diff
@@ -125,6 +142,35 @@ private int CountInstalled(IReadOnlyList<NotificationChannel> channels)
         return count;
     }
 
-    private string Summarize(string appId) =>
-        configuration.IsAppNotificationEnabled(appId) ? string.Empty : Loc.T(L.Settings.NotificationsOff);
+    private string Summarize(AppSettingsEntry entry)
```
XeldarAlz, line 132: Nit: the two tail branches re-derive the same on/off decision the first branch already makes, and both return NotificationsOff. Reading the two flags once, treating an absent capability as on, collapses this to a single four-way decision.

XeldarAlz (changes requested):
Release build clean (0 warnings, 0 errors) and 1961/1961 tests pass on 37c5afa. Not verified in-game.

Three blocking findings inline: the DTR and minimized-phone badges ignore the new toggle, Wallet is still double-gated, and the migration's app-id literals are untested.

XeldarAlz (approved):
All twelve findings from the first pass are resolved. Merged dev in and pushed the remaining fixes: the capsule and DTR badges now refresh on a Configuration.BadgeSettingsChanged event instead of waiting for the next notification, SettingsApp opts into HasBadge, and DailiesApp drops the field the old badge card left behind.

Release build clean (0 warnings, 0 errors), 2016/2016 tests pass. Not verified in-game.

PR #153, opened 2026-09-02
fix(home): stop widget-removal from soft-locking the home screen

jambalong:
## What

Fixes a home screen soft-lock when removing a widget via the size menu's "Remove" row. Added `WidgetSizeMenu.CloseImmediate()`, called from `HomeScreen.DrawSizeMenu` when the menu's tile can no longer be found.

## Why

`WidgetSizeMenu.Close()` only sets `closing = true`; the shrink animation in `Draw` is what actually clears `tile` back to null. Remove detaches the tile from the layout first, so on the next frame the menu's anchor lookup fails, `DrawSizeMenu` returns early, and `Draw` (and its cleanup) never runs again. `tile` never clears, so home input stays blocked until reload. Resize doesn't hit this because the tile stays in the layout.

Ref: https://discord.com/channels/1527213298087493732/1543792025244274738

## How I tested it

Release build, `/phone`:
- Add a widget, open its size menu, tap Remove: closes cleanly, home stays clickable. Repeated a few times.
- Pick a size, tap outside the menu: both still animate out normally.
- No exceptions in `/xllog` through the repro.

## Screenshots

widget-remove-fix

## Checklist

Strings:

Quality:

### DIFF
--- a/src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
+++ b/src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
@@ -44,6 +44,7 @@ internal sealed class WidgetSizeMenu
     {
         tile = null;
         closing = false;
+        pop.SnapTo(0f);
     }
 
     public void Draw(Rect content, Rect anchor, PhoneTheme theme, float delta, float scale)
--- a/src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
+++ b/src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
@@ -40,6 +40,12 @@ internal sealed class WidgetSizeMenu
         closing = true;
     }
 
+    public void CloseImmediate()
+    {
+        tile = null;
+        closing = false;
+    }
+
     public void Draw(Rect content, Rect anchor, PhoneTheme theme, float delta, float scale)
     {
         if (tile is null)
--- a/src/Aetherphone/Core/Shell/HomeScreen.cs
+++ b/src/Aetherphone/Core/Shell/HomeScreen.cs
@@ -192,7 +192,7 @@ internal sealed class HomeScreen
         var anchor = interaction.CommittedRect(metrics, tile);
         if (anchor is not { } rect)
         {
-            sizeMenu.Close();
+            sizeMenu.CloseImmediate();
             return;
         }
 
--- a/src/Aetherphone/Core/Shell/HomeScreen.cs
+++ b/src/Aetherphone/Core/Shell/HomeScreen.cs
@@ -99,6 +99,7 @@ internal sealed class HomeScreen
     {
         gallery.CloseImmediate();
         spotlight.CloseImmediate();
+        sizeMenu.CloseImmediate();
         interaction.ResetForReveal();
         var page = PageContaining(appId);
         if (page >= 0)


src/Aetherphone/Core/Shell/HomeScreen.cs
```diff
@@ -192,7 +192,7 @@ private void DrawSizeMenu(Rect content, in HomeMetrics metrics, PhoneTheme theme
         var anchor = interaction.CommittedRect(metrics, tile);
         if (anchor is not { } rect)
         {
-            sizeMenu.Close();
+            sizeMenu.CloseImmediate();
```
XeldarAlz, line 196: Blocking: there is one more spot that needs the same fix. HomeScreen.cs:100 closes the widget gallery and the search overlay when an app opens, but not this menu. So if an app opens without tapping its icon (a notification link, `/phone <app>`, the app switcher), the home screen still locks up. Adding `sizeMenu.CloseImmediate()` there too should cover it.

src/Aetherphone/Core/Shell/HomeScreen.cs
```diff
@@ -192,7 +192,7 @@ private void DrawSizeMenu(Rect content, in HomeMetrics metrics, PhoneTheme theme
         var anchor = interaction.CommittedRect(metrics, tile);
         if (anchor is not { } rect)
```
XeldarAlz, line 194: Non-blocking: this line also runs while the menu is playing its normal close animation, so the menu can disappear instantly instead of shrinking. FolderOverlay.cs:63 avoids that by only doing this when it is not already closing. Nice to copy, fine to leave.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
```diff
@@ -40,6 +40,12 @@ public void Close()
         closing = true;
     }
 
+    public void CloseImmediate()
```
XeldarAlz, line 43: Non-blocking, just something to know for next time: the menu is what puts itself in the broken state (line 95 removes the widget, then closes), so this cleanup could also live inside this class instead of in HomeScreen. FolderOverlay does it that way. Your fix works either way.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
```diff
@@ -40,6 +40,12 @@ public void Close()
...
+    {
+        tile = null;
```
XeldarAlz, line 45: Nit: the two other CloseImmediate methods in the codebase (WidgetGallery.cs:42 and SpotlightOverlay.cs:57) also reset their animation value. Adding `pop.SnapTo(0f)` here matches them and avoids a surprise later. Not a problem today.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
```diff
@@ -40,6 +40,12 @@ public void Close()
...
+        closing = false;
```
XeldarAlz, line 46: Non-blocking, this was already the case before your PR: when a widget is removed at line 95, nothing tells the input handler to forget it, so it keeps tracking a widget that is gone for a moment. HomeGridRenderer.cs:130 shows the call that is missing. Fine as a follow-up.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
```diff
@@ -40,6 +40,12 @@ public void Close()
...
+    }
```
XeldarAlz, line 48: Non-blocking: if you want a nice small next step, this is easy to test. Nothing in Open, Close or CloseImmediate touches the UI, so a test can open the menu, remove the widget, call CloseImmediate and check the menu is closed. There are ready made helpers in HomeFakes.cs.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
```diff
@@ -40,6 +40,12 @@ public void Close()
...
+
     public void Draw(Rect content, Rect anchor, PhoneTheme theme, float delta, float scale)
```
XeldarAlz, line 50: Non-blocking, also older than this PR: the menu does not tell the rest of the UI that it is on top, so widgets underneath still react to the mouse through it. Other popups handle that with `UiInteract.HoverOverlay` (see MentionPopup.cs:69). Just noting it, not asking you to fix it.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
```diff
@@ -40,6 +40,12 @@ public void Close()
...
     {
```
XeldarAlz, line 51: Non-blocking, also pre-existing: the menu is kept inside the screen on the left and right (line 73) but not on the top, so a tall menu on a widget near the top can draw outside the phone. Writing it down so it does not get forgotten.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
```diff
@@ -40,6 +40,12 @@ public void Close()
...
         if (tile is null)
```
XeldarAlz, line 52: Non-blocking, pre-existing: Remove happens the moment the mouse button goes down, so you cannot slide off the row to change your mind. Everywhere else on the home screen waits for the button to come back up. Separate cleanup, not yours to do here.

XeldarAlz (changes requested):
Really solid first PR, thanks for digging into this one. Build is clean (Release, 0 warnings) and all 1959 tests pass. Not tested in-game yet.

Your fix is the right approach. Only the blocking comment needs a change, one more line in the same spirit as yours. Everything else is background or older code, so please do not feel you have to touch it. No changelog entry needed, we handle those at release time.

src/Aetherphone/Core/Shell/HomeScreen.cs
jambalong: Fixed: `PrepareReveal` now calls `sizeMenu.CloseImmediate()` too.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
jambalong: Added `pop.SnapTo(0f)` to `CloseImmediate()`.

src/Aetherphone/Core/Shell/HomeScreen.cs
jambalong: I agree, this would be a nice follow-up.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
jambalong: Same follow-up as above will move this into `WidgetSizeMenu` itself.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
jambalong: Noted, thanks for writing it down.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
jambalong: Just noticed it after your mention, noted.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
jambalong: Noted.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
jambalong: Good idea, will follow-up.

src/Aetherphone/Core/Shell/Home/WidgetSizeMenu.cs
jambalong: Noted.

XeldarAlz (approved):
Approving. Both requested changes landed:

- `PrepareReveal` now calls `sizeMenu.CloseImmediate()` (1f0b6b4a). That was the blocking one: opening an app without tapping its icon (notification link, `/phone <app>`, the app switcher) no longer leaves the menu holding a detached tile.
- `CloseImmediate()` resets the animation with `pop.SnapTo(0f)` (f64f51d1), matching `WidgetGallery.cs:42` and `SpotlightOverlay.cs:57`.

On the missing checks: CI was never failing, it was queued behind the fork approval gate since your first push, so nothing had run. I approved it and both jobs are green, Guards and Build (Windows), tests included.

The `FolderOverlay`-style `!closing` guard at `HomeScreen.cs:196` stays as the follow-up you offered, together with the unit test and the input handler cleanup. None of that blocks this.

Thanks for tracing the root cause properly rather than papering over the symptom, the writeup made this quick to review.

PR #154, opened 2026-09-02
feat(hunts) add native map markers, and sighting support

Deldee:
## What

- Added the ability to view landmines on the in game map and minimap (For instanced area, default to 1 if the player isnt standing in it, else show their current instance)
- Native marker show all ZonePOI for this map, including A/B ranks location, but those default to "unavailable for S rank spawns"
- Add support for sighting, to show which spots were already claimed by another mark (Requires reporter permissions on faloop)
- Fixed some window % math issue and sniping handling
- Editing the notification tab for a proper setting tab

## Screenshots
image
image

## Checklist

Strings:

Quality:


src/Aetherphone/Core/Hunts/HuntCandidateCache.cs
```diff
@@ -0,0 +1,54 @@
...
+        this.zoneCatalog = zoneCatalog;
+        this.hunts = hunts;
+    }
+
+    public (IReadOnlyList<HuntPoiState> States, int? ReportedPoiId) ResolveFor(HuntMobDefinition mob,
+        string worldId, int zoneInstance, string zoneId, bool includeLandmineOnlySpots)
+    {
+        var token = hunts.CandidateStateToken;
+        var key = (mob.Id, worldId, zoneInstance, zoneId);
+        if (!entries.TryGetValue(key, out var entry) || !entry.Token.Equals(token))
```
XeldarAlz: **Blocking: `entries` is a plain `Dictionary` mutated from two different threads.**

`HuntsApp.DrawDetailBody` calls `ResolveFor` from the ImGui **render** thread every frame, while `HuntsMapMarkers.OnFrameworkUpdate` calls `ResolveZoneMarkers` -> `ResolveFor` from the **framework/game** thread every 2 seconds. `entries[key] = entry` on line 42 can then resize the dictionary while the other thread is walking its buckets.

Every other piece of shared hunt state in this PR is deliberately thread safe (`HuntsService` uses `volatile` copy-on-write plus `realtimeStateGate`), so this cache is the odd one out.

Scenario: player opens the Hunts detail screen and the FFXIV map at the same time in a zone with active windows. A framework tick inserts a new entry and triggers a resize exactly while the draw thread is inside `TryGetValue` -> `InvalidOperationException`, a bogus hit, or an infinite loop inside the game process.

Either guard the dictionary with a lock, or give the map-marker path its own cache instance.

src/Aetherphone/Apps/Hunts/ProgressBar.cs
```diff
@@ -7,16 +7,18 @@ namespace Aetherphone.Apps.Hunts;
...
+            ? ((int)Math.Round(known, MidpointRounding.AwayFromZero)) + "%"
+            : "?";
         var labelSize = Typography.Measure(percentageLabel, labelStyle);
         var radius = height * 0.5f;
         var trackMax = new Vector2(right - labelSize.X - labelGap, top + height);
         var trackMin = new Vector2(left, top);
         Squircle.Fill(drawList, trackMin, trackMax, radius, ImGui.GetColorU32(Palette.WithAlpha(color, trackAlpha)));
 
-        var fraction = (float)Math.Clamp(percentage / 100.0, 0.0, 1.0);
+        var fraction = percentage is { } value ? (float)Math.Clamp(value / 100.0, 0.0, 1.0) : 1f;
```
XeldarAlz: **An unknown percentage now renders a completely full bar.**

`percentage is { } value ? ... : 1f` means a `null` percentage fills the track 100%. `HuntWindowMath.Percentage` returns `null` not only past the phantom-cycle degeneracy, but also whenever timing cannot be resolved at all: `ResolveCore` bails when `mob is null`, `mob.Windows.Length == 0`, `timing.Cap is null`, or `span <= 0`.

Scenario: Faloop reports a window for a mob that is not yet in the bundled `HuntMob.json` (a new patch's marks). `mobCatalog.Find` returns null -> `Percentage` is null -> the list row and the detail hero bar both paint a **full** bar labelled `?`, telling the user the window is maxed out. Before this PR the same case rendered `percentage ?? 0d`, an empty bar.

An indeterminate value should not read as 100%; consider `0f` fill (or a striped/indeterminate track) for the unknown case.

src/Aetherphone/Core/Hunts/HuntsService.cs
```diff
@@ -523,6 +604,8 @@ private void OnMobReportReceived(HuntsSocketMobReport report)
             return;
         }
 
+        FlushPendingSightingReplace(identity.MobId, identity.WorldId, identity.ZoneInstance);
```
XeldarAlz: **The `sighting_replace` deferral is defeated by the very next mob report.**

`ApplyOrDeferSightingReplace` parks the replace while the mob is spawned, but `FlushPendingSightingReplace` is called unconditionally at the top of `OnMobReportReceived`, for **every** action on that identity, not only the ones that end the spawn.

Scenario: `sighting_replace` arrives while `ixtab@mateus` is up -> deferred. A second later a `spawn_claim` (or `spawn_progress`, or `spawn_location`) frame lands for the same identity -> the pending replace is flushed and applied while the mob is still spawned, which is exactly the state the deferral existed to avoid. Claim frames fire constantly during a live spawn, so in practice the deferral almost never holds.

Flush only from the paths that actually clear the spawn (`death`, `spawn_false`), or gate the flush on `!IsSpawned(...)`.

src/Aetherphone/Core/Hunts/HuntCandidateResolver.cs
```diff
@@ -0,0 +1,400 @@
...
+            if (hunts.IsPoiSighted(mob.Id, worldId, zoneInstance, points[index].Id))
+            {
+                continue;
+            }
+
+            unsightedCount++;
+            soleUnsightedPoiId = points[index].Id;
+        }
+
+        var confirmedPoiId = reportedPoiId ?? (unsightedCount == 1 ? soleUnsightedPoiId : (int?)null);
```
XeldarAlz, line 200: **A zone with a single candidate POI is always reported as `Confirmed`, even with zero sightings and no spawn.**

The deduction "one unsighted spot left, so that is where it is" only holds if sightings actually eliminated the others. When `points.Count == 1` and nothing has been sighted, `unsightedCount` is still `1`, so `confirmedPoiId` becomes that lone point and line 198 paints it `HuntsMapMarkerState.Confirmed`.

Scenario: any mob whose active phase resolves to exactly one POI in the target zone, sitting closed with no reports at all. The native map draws icon `60403` and the legend says "Confirmed" - a hunter is told the mark is standing there.

Require evidence before deducing, e.g. `points.Count > 1 && unsightedCount == 1`, and preferably also `spawned`.

src/Aetherphone/Core/Maps/HuntsMapMarkers.cs
```diff
@@ -0,0 +1,209 @@
...
+        this.hunts = hunts;
+        this.mobCatalog = mobCatalog;
+        this.zoneCatalog = zoneCatalog;
+        this.candidateCache = candidateCache;
+        Plugin.Framework.Update += OnFrameworkUpdate;
+        Plugin.AddonLifecycle.RegisterListener(AddonEvent.PostSetup, AreaMapAddonName, OnAreaMapOpenedOrChanged);
+        Plugin.AddonLifecycle.RegisterListener(AddonEvent.PostRefresh, AreaMapAddonName, OnAreaMapOpenedOrChanged);
+    }
+
+    public void Dispose()
```
XeldarAlz: **`Dispose` leaves the placed markers on the game map.**

The listeners are unregistered but `ResetMapMarkers()` / `ResetMiniMapMarkers()` are never called, and `hasPlacedMarkers` stays true on a dead object.

Scenario: user has markers on, then disables or updates the plugin while standing in a hunt zone. Every candidate icon stays burned onto the map and minimap until the game itself resets them (zone change / relog), with nothing left alive to clear them.

Calling `ClearNativeMarkersIfNeeded` here when `AgentMap.Instance()` is non-null would close it out.

src/Aetherphone/Core/Maps/HuntsMapMarkers.cs
```diff
@@ -0,0 +1,209 @@
...
+            return;
+        }
+
+        if (!mustRedraw && hasPlacedMarkers && points.Count == lastPlacedPoints.Count &&
+            lastPlacedPoints.SetEquals(points))
+        {
+            return;
+        }
+
+        agentMap->ResetMapMarkers();
```
XeldarAlz, line 128: **Non-blocking: `ResetMapMarkers()` / `ResetMiniMapMarkers()` wipe every other plugin's custom markers.**

`AgentMap`'s custom marker list is shared process-wide. This code clears the whole list before each redraw, and it redraws on `AreaMap` PostSetup/PostRefresh and whenever the point set changes.

Scenario: a user runs Aetherphone alongside any plugin that places custom map markers (gathering-node, treasure-map or navigation helpers). Opening the map fires our PostSetup handler, we `ResetMapMarkers()`, and their icons vanish until that plugin happens to re-add them - a hard-to-diagnose cross-plugin conflict that will get reported against us.

Worth tracking the handles we add and removing only those if the API allows it; at minimum this deserves a line in the docs so support can recognise the symptom.

src/Aetherphone/Core/Maps/HuntsMapMarkers.cs
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        if (!configuration.HuntsNativeMapMarkers)
+        {
+            ClearNativeMarkersIfNeeded(agentMap);
+            return;
+        }
+
+        var territoryId = agentMap->SelectedTerritoryId;
+        var worldId = HuntDataCenterWorlds.SlugFor(LocationShare.CurrentWorldId());
```
XeldarAlz: **Two string allocations plus an Excel row lookup on every framework tick, above the throttle.**

`SlugFor` runs `world.Name.ExtractText()` (allocates) and `.ToLowerInvariant()` (allocates again), and it sits **before** the `RefreshInterval` early-out on line 100, so it executes on every single framework tick while `HuntsNativeMapMarkers` is on - not once every 2 seconds like the rest of the body.

Cache the slug against the `LocationShare.CurrentWorldId()` row id and recompute only when that id changes; the world cannot change without the id changing.

src/Aetherphone/Windows/HuntsMapMarkersIndicatorWindow.cs
```diff
@@ -0,0 +1,257 @@
...
+    public override void PreDraw()
+    {
+        TryGetAreaMapBounds(out var mapPosition, out var mapSize);
+        var scale = UiScale.Global;
+        var label = Loc.T(L.Hunts.NativeMapMarkersIndicator);
+        var labelSize = Typography.Measure(label, TextScale, FontWeight.SemiBold);
+        var headerRowWidth = (IconWidth + IconGap) * scale + labelSize.X;
+
+        instanceLabel = markers.ShownInstance is { } instance
+            ? string.Format(Loc.T(L.Hunts.NativeMapMarkersInstanceIndicator), instance)
```
XeldarAlz: **Per-frame allocation in draw code.**

`string.Format(Loc.T(...), instance)` boxes `instance` into an `object[]` and allocates a fresh result string on **every frame** the AreaMap is open, since `PreDraw` runs every frame.

CLAUDE.md (repo root): "This is an immediate mode UI (Dear ImGui): Draw code runs every frame. No LINQ, no per-frame allocations, no reflection in anything reachable from Draw." The PR checklist repeats it: "Draw-path code allocates nothing per frame".

Cache `instanceLabel` and rebuild it only when `markers.ShownInstance` or the language actually changes.

src/Aetherphone/Core/Hunts/HuntCandidateCache.cs
```diff
@@ -0,0 +1,54 @@
...
+            entry = new Entry(token, states, reportedPoiId);
+            entries[key] = entry;
+        }
+
+        if (!includeLandmineOnlySpots)
+        {
+            return (entry.States, entry.ReportedPoiId);
+        }
+
+        var withLandmines = new List<HuntPoiState>(entry.States);
```
XeldarAlz, line 69: **Non-blocking: the cache grows without bound, and the landmine path allocates on every tick regardless.**

Two problems on this path:

1. `entries` is never evicted. `ResolveZoneMarkers` calls `ResolveFor` for **every window on the current world**, so an entry is created per `(mob, world, instance, zone)` - including the overwhelming majority that resolve to an empty state list. As the player moves between zones the key space multiplies by every zone visited, and nothing is ever removed for the life of the session.

2. `includeLandmineOnlySpots: true` skips the cache entirely: a fresh `List<HuntPoiState>` is copied per window per 2-second tick (hundreds of allocations), and `AppendLandmineOnlyStates` then runs `results.TrueForAll(existing => ...)` inside a loop over every POI in the zone - a new closure object per iteration on top of the O(n*m) scan.

The landmine set depends only on `zoneId`, so it could be precomputed once per zone and merged without the per-window copy; `entries` should also be bounded or cleared on zone change.

src/Aetherphone/Apps/Hunts/HuntsApp.Detail.cs
```diff
@@ -698,32 +602,32 @@ private bool DrawNavigateButton(HuntsView view, float scale, int? confirmedPoiId
...
     {
         ArmPendingInstanceSync(worldId, territoryId, zoneInstance);
 
         var destination = TravelPlanner.ResolveNearestAetheryteTo(territoryId, worldId, LocationShare.CurrentWorldId(),
-            Plugin.ClientState.TerritoryType, coordinate);
+            Plugin.ClientState.TerritoryType, targetCoordinate);
         if (destination.Kind == TravelKind.AlreadyThere)
         {
-            if (coordinate is { } here)
+            if (flagCoordinate is { } here)
```
XeldarAlz, line 621: **Tapping an aetheryte dot is now a dead click whenever there is no reported spawn location.**

`DrawAetheryteDot` (line 556) passes `flagCoordinate = confirmedPoiId is { } confirmed ? ... : null`, and `confirmedPoiId` comes from `entry.ReportedPoiId`, which `ResolveMobZoneStates` only populates when the mob is **spawned**. Previously this call site passed the aetheryte's own coordinate, so the `TravelKind.AlreadyThere` branch dropped a flag on it.

Scenario: the player is already standing in the hunt zone (the common case while camping a window), the mark has not popped yet, and they tap an aetheryte dot on the detail map. `ResolveNearestAetheryteTo` returns `AlreadyThere`, `flagCoordinate` is null, and the method returns having done nothing at all - no flag, no travel, no feedback.

Fall back to the tapped aetheryte's own coordinate when there is no confirmed POI.

src/Aetherphone/Apps/Hunts/HuntsApp.cs
```diff
@@ -217,16 +227,14 @@ private void DrawFooter(Rect bar, HuntsRoute route, float scale)
             authLocked ? Loc.T(L.Hunts.HistoryRequiresLoginTooltip) : Loc.T(L.Hunts.HistoryTab),
             Disabled: authLocked);
         footerTabs[1] = new NavTab(FontAwesomeIcon.Book, Loc.T(L.Hunts.ListTab));
-        footerTabs[2] = new NavTab(FontAwesomeIcon.Bell,
-            authLocked ? Loc.T(L.Hunts.NotificationSettingsRequiresLoginTooltip) : Loc.T(L.Hunts.NotificationSettingsTab),
-            Disabled: authLocked);
-        footerTabs[3] = new NavTab(FontAwesomeIcon.Question, Loc.T(L.Hunts.GuideTab), AnchorKey: "hunts.guide");
+        footerTabs[2] = new NavTab(FontAwesomeIcon.Question, Loc.T(L.Hunts.GuideTab), AnchorKey: "hunts.guide");
+        footerTabs[3] = new NavTab(FontAwesomeIcon.Cog, Loc.T(L.Hunts.SettingsTab), AnchorKey: "hunts.settings");
```
XeldarAlz, line 231: **Non-blocking: the notification settings lost their login gate.**

The old Notifications tab was `Disabled: authLocked` with `NotificationSettingsRequiresLoginTooltip`; that string is deleted from `L.cs` and all nine catalogs in this PR. The new Settings tab is never disabled, and `DrawSettings` renders the rank chips, world picker, per-mark overrides and "Reset to Default" unconditionally.

Scenario: a logged-out user opens Settings, spends time configuring which ranks and worlds to be notified about, and nothing will ever fire because live spawn notifications require the Faloop session. The onboarding copy added in this PR even says "once you sign in, manage which notifications you get" - but the UI never tells them that.

The map-markers and reset-tutorial rows genuinely do not need auth, so keeping the tab itself enabled is right; gate the notifications section instead (disabled rows plus a sign-in hint).

src/Aetherphone/Core/Hunts/HuntCandidateResolver.cs
```diff
@@ -0,0 +1,400 @@
...
+        if (mob.Rank == "SS" && !zoneConfirmed)
+        {
+            return;
+        }
+
+        var activePhase = hunts.PhaseFor(mob.Id, worldId, zoneInstance);
+        var poiIds = ResolveCandidatePoiIds(mob, activePhase, out var finalPhase);
+        if (mob.Rank != "SS")
+        {
+            poiIds.RemoveWhere(mobCatalog.IsSsRankPoi);
```
XeldarAlz: **Non-blocking: this strips POIs from non-SS mobs globally, not per zone or per mob.**

`IsSsRankPoi` is true for any POI id that appears in **any** SS-rank mob's phase list anywhere in the catalog. Every non-SS mob then loses those ids from its own candidate set.

Scenario: an S rank shares a spawn point with the SS rank that roams the same zone (exactly the kind of overlap `LandmineAppendNeverOverwritesAnSRankSOwnPoiState` is guarding against elsewhere). The S rank silently loses a legitimate candidate spot - it is never drawn on the detail map or the native map, and it also skews `ResolveBestZoneId`'s per-zone counts.

Also note that `RemoveWhere(mobCatalog.IsSsRankPoi)` allocates a `Predicate<int>` delegate from the method group on every cache miss.

The filter probably wants to be "POIs that belong to an SS rank **in this zone** and to no non-SS mob".

src/Aetherphone/Core/Hunts/HuntCandidateResolver.cs
```diff
@@ -0,0 +1,400 @@
...
+        }
+
+        var confirmedPoiId = reportedPoiId ?? (unsightedCount == 1 ? soleUnsightedPoiId : (int?)null);
+        for (var index = 0; index < points.Count; index++)
+        {
+            var poi = points[index];
+            HuntsMapMarkerState state;
+            if (confirmedPoiId is { } confirmed)
+            {
+                state = poi.Id == confirmed ? HuntsMapMarkerState.Confirmed : HuntsMapMarkerState.Sighted;
```
XeldarAlz, line 207: **Non-blocking: `Sighted` now carries three different meanings and the legend explains only one.**

On this line every POI that is not the confirmed one is labelled `Sighted`, whether or not anyone ever sighted it. `AppendLandmineOnlyStates` (line 261) then labels A/B-rank spawn points `Sighted` too.

So icon `60444` / legend row "Sighted" simultaneously means (a) a spot genuinely reported as sighted, (b) a candidate eliminated because the mark was found elsewhere, and (c) an A/B-rank spot that was never an S-rank candidate at all.

Scenario: a hunter opens the map, counts six "Sighted" pins in the zone, and reads it as six sighting reports. Note this is also a behaviour change: the previous detail-map code *hid* non-confirmed dots once a location was known (`if (confirmedKnown && poi.Id != confirmedPoiId) continue;`).

A distinct `Eliminated` / `OtherRank` state with its own icon and legend row would say what is actually meant.

src/Aetherphone/Core/Maps/HuntsMapMarkers.cs
```diff
@@ -0,0 +1,209 @@
...
+            return;
+        }
+
+        agentMap->ResetMapMarkers();
+        agentMap->ResetMiniMapMarkers();
+        hasPlacedMarkers = false;
+        lastPlacedPoints.Clear();
+    }
+
+    private static unsafe int? ResolveExplicitInstance(uint territoryId)
```
XeldarAlz, line 190: **Non-blocking: an instanced zone whose `InstanceId` reads 0 lands in a bucket that never has data.**

When the player is standing in `territoryId` this returns `(int)uiState->PublicInstance.InstanceId` even when that is `0`, and `ResolveZoneMarkers` then takes the `explicitInstance is { } known` branch: `targetInstance = 0`, and `shownInstance = 0` because `zoneIsInstanced` is true. Faloop numbers instances from 1, so `statesByInstance` has no key `0`, the method returns with no markers at all, and the indicator chip announces "Currently showing instance 0".

Scenario: zoning into an instanced hunt zone during the window where `PublicInstance` has not been populated yet.

Returning `null` for a zero instance id would fall through to the documented "default to 1" behaviour instead.

src/Aetherphone/Windows/HuntsMapMarkersIndicatorWindow.cs
```diff
@@ -0,0 +1,257 @@
...
+{
+    private const ImGuiWindowFlags ChipFlags = ImGuiWindowFlags.NoTitleBar | ImGuiWindowFlags.NoScrollbar |
+                                               ImGuiWindowFlags.NoScrollWithMouse | ImGuiWindowFlags.NoCollapse |
+                                               ImGuiWindowFlags.NoResize | ImGuiWindowFlags.NoBackground |
+                                               ImGuiWindowFlags.NoMove | ImGuiWindowFlags.NoFocusOnAppearing |
+                                               ImGuiWindowFlags.NoNavInputs | ImGuiWindowFlags.NoNavFocus |
+                                               ImGuiWindowFlags.NoSavedSettings;
+
+    private const string AreaMapAddonName = "AreaMap";
+    private const uint CandidateIconId = 60557u;
```
XeldarAlz: **Non-blocking: the seven marker icon ids are duplicated verbatim from `HuntsMapMarkers`.**

`CandidateIconId` through `FateActiveIconId` (lines 24-30) are byte-for-byte copies of the constants in `src/Aetherphone/Core/Maps/HuntsMapMarkers.cs` (lines 13-19). This legend is the key for those exact markers, so the two copies must agree or the legend lies about what is on the map - and nothing enforces that.

Lift the ids next to `HuntsMapMarkerState`, or expose `HuntsMapMarkers.IconFor` as an internal static, and have both the marker placement and the legend read the one source.

src/Aetherphone/Core/Hunts/HuntCandidateResolver.cs
Deldee: This is the expected behavior
"confirmed" does not mean that the mark has spawned on a landmine, but that the S mark has claimed this landmine, the distinction is important because some S rank require to do actions specifically to their claimed landmine (Nandi, Gamma, Armstrong, Kirlinger,Forecaster, as examples), finding those landmine is the main goal of sightings

If there is only one POI (a non existing case,as this would softlock S ranks, but in theory), the expected behavior would be for this spot to be confirmed instantly
I'll review the legend, annoting "Confirmed Landmine" might make the distinction clearer

Worth nothing that sighting are only available to reporters account, who are familiar with this terminology to begin with

src/Aetherphone/Core/Maps/HuntsMapMarkers.cs
Deldee: The api sadly doesnt allow us to track our own handle, it's simply a shared array
I've implemented a workaround where we keep track of how many markers we added and then clear only that amount, but it can still overlap if plugins were to add markers after us

This is a fairly low effect issues, markers get redrawn on every map open and wiped everytime the map is closed, so the risk is fairly minimal to begin with

src/Aetherphone/Core/Hunts/HuntCandidateResolver.cs
Deldee: This is working as intended
We want landmines to render on the map even if they don't belong to the S rank, because the map is zone-scoped. Meanwhile the detail page is mark-scoped (which is also why landmines disappear on spawn on the detail page, as there is nothing else using them in that scope)
The sighting data, on the other hand, is always explicitly S-scoped, which is why B/A landmines are opted out of it

As mentioned in another comment, the goal of sighting is only to say "this landmine is not currently claimed by the S rank"
All three of those behavior carry that exact meaning

The distinction between Eliminated/Sighted is, for all intent and purpose, non-existent

src/Aetherphone/Core/Maps/HuntsMapMarkers.cs
Deldee: Reverting this change, the Marker array is too unpredictable to assume the marker we place will remain at the tails end (fates updating, questing being completed etc, all update the array)
Adding documentation to game_interaction.md instead, and leaving it as is

XeldarAlz (approved):
All 15 review threads are closed and verified at the head commit.

Twelve are fixed in code: the candidate cache is now lock guarded, an unresolvable percentage paints an empty bar instead of a full one, the deferred sighting_replace flushes only from the spawn clearing path, Dispose resets the markers it placed, the world slug is cached by row id, the instance chip label is rebuilt only on an instance or language change, the cache is bounded and the landmine spots precomputed per zone, tapping an aetheryte dot routes to that aetheryte, the notification rows are gated behind a sign in hint, candidate resolution no longer reads a companion's phases, an unpopulated instance id falls through to the documented default, and the legend and the placement code share one set of icon ids in HuntsMapMarkerIcons.

Three are answered as intended behavior and I accept the reasoning: the single candidate Confirmed state, the Sighted semantics, and the full marker array reset. The relabelled legend (Claimed by S Rank, Unclaimed by S Rank) resolves the ambiguity behind the first two, and the reset tradeoff is now written up in the gotchas in docs/game-integration.md.

Conventions check is clean: no em dashes, localization lockstep holds across all nine catalogs with L.cs and en.json in sync and real translations throughout, no comments, no async void, no raw GlobalScale, no LINQ outside the tests.

I resolved the ChipRail conflict against dev in d69e3164, keeping both the centered parameter from dev and interactive from this branch. Release build clean and 2010 tests pass on the merged tree.

src/Aetherphone/Core/Hunts/HuntCandidateResolver.cs
```diff
@@ -0,0 +1,417 @@
...
+            if (hunts.IsPoiSighted(mob.Id, worldId, zoneInstance, points[index].Id))
+            {
+                continue;
+            }
+
+            unsightedCount++;
+            soleUnsightedPoiId = points[index].Id;
+        }
+
+        var confirmedPoiId = reportedPoiId ?? (unsightedCount == 1 ? soleUnsightedPoiId : (int?)null);
```
XeldarAlz: Blocking: this elimination runs for every rank and regardless of `spawned`, so any zone that resolves to a single POI is painted Confirmed with nothing behind it. A and B ranks never receive sightings, so `unsightedCount == 1` is trivially true for them. Gate it on `spawned`, the way `finalLocationResolved` above it is.

src/Aetherphone/Core/Hunts/HuntCandidateResolver.cs
```diff
@@ -0,0 +1,417 @@
...
+            {
+                statesByPoiId = new Dictionary<int, (HuntsMapMarkerState, int)>();
+                statesByInstance[window.ZoneInstance] = statesByPoiId;
+            }
+
+            var rankTier = RankTier(mob.Rank);
+            for (var stateIndex = 0; stateIndex < states.Count; stateIndex++)
+            {
+                var (poi, state) = states[stateIndex];
+                var priority = rankTier * TrackedRankPriorityBoost + Priority(state);
```
XeldarAlz: Blocking: `AppendLandmineOnlyStates` is S rank only, so a landmine-only POI scores RankTier(1) * 10 + Priority(Sighted)(2) = 12, while another tracked mob's real Candidate at the same POI scores 10 (Candidate falls through to `_ => 0`). The grey unclaimed icon wins over a live candidate.

src/Aetherphone/Core/Maps/HuntsMapMarkers.cs
```diff
@@ -0,0 +1,213 @@
...
+        {
+            var point = points[index];
+            lastPlacedPoints.Add(point);
+            var worldX = MapPixelMath.ToWorldCoordinate(point.RawX, map.SizeFactor, map.OffsetX);
+            var worldZ = MapPixelMath.ToWorldCoordinate(point.RawY, map.SizeFactor, map.OffsetY);
+            var worldPosition = new Vector3(worldX, 0f, worldZ);
+            var iconId = HuntsMapMarkerIcons.IconFor(point.State);
+            var scale = ScaleFor(point.State);
+            agentMap->AddMapMarker(worldPosition, iconId, scale);
+            agentMap->AddMiniMapMarker(worldPosition, iconId, scale);
```
XeldarAlz: Blocking: `territoryId` here is `agentMap->SelectedTerritoryId`, so browsing another zone's map paints that zone's points onto the minimap, which always renders the current territory, at coordinates translated with the wrong map's SizeFactor. Gate `AddMiniMapMarker` on the player's actual territory.

src/Aetherphone/Core/Hunts/HuntsService.cs
```diff
@@ -350,6 +430,7 @@ private void Refresh()
             }
             SeedActiveSpawnsFromPoll(status?.Spawns);
             EnsureWindowEntriesForActiveSpawns();
+            sightedPoiIds = BuildSightedPoiIds(status?.Sightings);
```
XeldarAlz: Blocking: this assignment is outside `realtimeStateGate`, while `AddSighting`, `RemoveSighting` and `ReplaceSightings` all read-copy-write under it. A socket frame landing during the poll loses either the new sighting or the whole polled set. Line 965 in `Reset` has the same gap. Move both inside the lock.

src/Aetherphone/Apps/Hunts/HuntsApp.Detail.cs
```diff
@@ -267,16 +189,19 @@ private void DrawDetailBody(Rect body, float scale, HuntsView view)
...
                 detailMapActivePhase = livePhase;
                 detailMapConfirmedZoneId = liveZoneId;
-                ResolveDetailMap(def, livePhase, liveZoneId);
+                ResolveDetailMap(def, view.WorldId, view.ZoneInstance);
             }
 
-            var spawned = hunts.IsSpawned(view.MobId, view.WorldId, view.ZoneInstance);
-            var confirmedPoiId = spawned ? ResolveConfirmedPoiId(view) : null;
-            if (DrawDetailZoneMap(scale, confirmedPoiId, view))
+            var (states, confirmedPoiId) = def is not null
```
XeldarAlz: Blocking: `confirmedPoiId` now comes from `ResolveMobZoneStates`, which returns before setting `reportedPoiId` for rank F (HuntCandidateResolver.cs:148) and for SS with an unconfirmed zone (:154). The removed `ResolveConfirmedPoiId` consulted `hunts.ConfirmedPoiIdFor` first, so the navigate button loses the reported spot in those two cases.

src/Aetherphone/Core/Maps/HuntsMapMarkers.cs
```diff
@@ -0,0 +1,213 @@
...
+        forceRedraw = false;
+
+        if (!TryResolveTarget(territoryId, worldId, out var map))
+        {
+            ClearNativeMarkersIfNeeded(agentMap);
+            return;
+        }
+
+        if (!mustRedraw && hasPlacedMarkers && points.Count == lastPlacedPoints.Count &&
+            lastPlacedPoints.SetEquals(points))
```
XeldarAlz: Non-blocking: `points` is a List and `lastPlacedPoints` is a HashSet, and `HuntsMapMarkerPoint` carries no POI id, so two POIs sharing coordinates and state make the counts unequal forever. The guard then never short-circuits and each 2s redraw wipes other plugins' markers. Compare one shape consistently.

src/Aetherphone/Windows/HuntsMapMarkersIndicatorWindow.cs
```diff
@@ -0,0 +1,260 @@
...
+            top += InstanceRowGap * scale;
+            Typography.Draw(drawList, new Vector2(instanceLeft, top), instanceLabel,
+                new Vector4(ink.X, ink.Y, ink.Z, 0.75f), InstanceTextScale, FontWeight.Regular);
+            top += instanceSize.Y;
+        }
+
+        top += ToggleGap * scale;
+        var toggleMin = new Vector2(min.X, top);
+        var toggleMax = new Vector2(max.X, top + ToggleRowHeight * scale);
+        var toggleHovered = ImGui.IsMouseHoveringRect(toggleMin, toggleMax);
```
XeldarAlz: Non-blocking: raw `IsMouseHoveringRect` plus `IsMouseClicked` with no window-hover or claim check, and the chip draws into `GetForegroundDrawList()`. A click on the phone where it overlaps this corner also toggles the legend. The rest of the codebase routes this through `UiInteract.Hover` and `UiInteract.Click`.

src/Aetherphone/Core/Hunts/HuntsService.cs
```diff
@@ -567,6 +662,143 @@ private void OnMobReportReceived(HuntsSocketMobReport report)
...
+    {
+        if (!IsSpawned(mobId, worldId, zoneInstance))
+        {
+            ReplaceSightings(mobId, worldId, zoneInstance, entries);
+            return;
+        }
+
+        lock (realtimeStateGate)
+        {
+            pendingSightingReplaces[new HuntSightingScopeKey(mobId, worldId, zoneInstance)] = entries;
```
XeldarAlz: Non-blocking: `pendingSightingReplaces` is drained only by `ClearSpawnRelease`, so a spawn that ends through the poll rather than a socket release strands its replace forever and leaks the entry. Nothing else sweeps the dictionary. Dropping it on the next poll that shows the mob unspawned would close it.

src/Aetherphone/Windows/HuntsMapMarkersIndicatorWindow.cs
```diff
@@ -0,0 +1,260 @@
...
+                                               ImGuiWindowFlags.NoSavedSettings;
+
+    private const string AreaMapAddonName = "AreaMap";
+    private const float CornerInset = 10f;
+    private const float ChipHeight = 28f;
+    private const float SidePadding = 12f;
+    private const float IconGap = 7f;
+    private const float IconScale = 0.66f;
+    private const float IconWidth = 11f;
+    private const float TextScale = 0.8f;
```
XeldarAlz: Nit: this file builds its own type ladder and spacing scale from literals (TextScale 0.8, LegendTextScale 0.9, ChipHeight 28, LegendRowHeight 30) fed straight into Typography. Per conventions.md text goes through a TextStyles style and spacing through Metrics tokens, or this overlay stops tracking them.

src/Aetherphone/Core/Hunts/HuntWindowMath.cs
```diff
@@ -65,12 +97,65 @@ public static HuntWindowStatus Status(HuntWindowDto window, HuntMobDefinition? m
...
+        cycles += hoursSinceCap < minEdge ? (hoursSinceCap - previousCapEdge) / (minEdge - previousCapEdge) : 1d;
+        if (hoursSinceCap > minEdge)
+        {
+            cycles += (hoursSinceCap - minEdge) / (capEdge - minEdge);
+        }
+
+        return cycles * 100d;
+    }
+
+    private static double RawOvertimePercentage(double min, double cap, double hoursSinceCap)
```
XeldarAlz: Nit: `RawOvertimePercentage` repeats `OvertimePercentage` line for line with the null branch swapped for a clamp, so it reduces to `OvertimePercentage(min, cap, hoursSinceCap) ?? fallback`. Duplicated, the label and the bar fill can drift apart, which is the class of bug this PR fixes elsewhere.

XeldarAlz (changes requested):
Done

PR #157, opened 2026-09-05
fix(aetherstream): prevent audio dropout on long streams

Farroness:
## What

Fixes audio stall detection in VideoEngine that prevents recovery when audio stops reporting on long streams. The watchdog timer logic was broken, causing audio dropouts after 20-30 minutes to go undetected.

## Why

Audio dropout detection had an impossible condition: it required audio position to be both present and absent simultaneously. When libmpv stopped reporting audio-pts, the code reset the stall timer on every frame instead of letting it count down, so the 10-second timeout never accumulated and recovery never triggered.

Closes #

## How I tested it

Release build: video playback with audio monitoring. Build verified clean with dotnet build Aetherphone.sln -c Release.

## Checklist

- [x] `dotnet test` passes
- [x] No `async void`, no new LINQ outside the allowlisted files, no raw `ImGuiHelpers.GlobalScale`, no hand-formatted clock text

Strings:

- [x] No new user-visible strings changed

Quality:

- [x] Verified build on Release variant
- [x] No AI attribution trailers in commits or PR body

### DIFF
--- a/src/Aetherphone/Core/Video/VideoEngine.cs
+++ b/src/Aetherphone/Core/Video/VideoEngine.cs
@@ -342,17 +342,20 @@ internal sealed class VideoEngine : IDisposable
             framesProgressAtTicks = now;
         }
 
-        if (info.CoreIdle || !info.HasAudioPosition)
+        if (info.CoreIdle)
         {
             lastObservedAudioPosition = info.AudioPositionSeconds;
             audioProgressAtTicks = now;
         }
-        else if (!audioPositionSeen
-                 || Math.Abs(info.AudioPositionSeconds - lastObservedAudioPosition) > StallPositionTolerance)
+        else if (info.HasAudioPosition)
         {
-            audioPositionSeen = true;
-            lastObservedAudioPosition = info.AudioPositionSeconds;
-            audioProgressAtTicks = now;
+            if (!audioPositionSeen
+                 || Math.Abs(info.AudioPositionSeconds - lastObservedAudioPosition) > StallPositionTolerance)
+            {
+                audioPositionSeen = true;
+                lastObservedAudioPosition = info.AudioPositionSeconds;
+                audioProgressAtTicks = now;
+            }
         }
 
         var nearEnd = info.DurationSeconds > 0d


Farroness:
This is more of a bandaid than fixing the audio dropping issues completely, but still better that our dropout detection is working.

I tested a 45 minute stream and did not experience any drops.

XeldarAlz:
@Farroness I also made a fix for this, can you test again?

src/Aetherphone/Core/Video/VideoEngine.cs
```diff
@@ -342,17 +342,20 @@ internal void ObserveForStall(in MpvPlaybackInfo info)
             framesProgressAtTicks = now;
         }
 
-        if (info.CoreIdle || !info.HasAudioPosition)
+        if (info.CoreIdle)
```
XeldarAlz, line 345: Blocking: the tracker now ages while audio-pts is missing, but `audioStalled` at VideoEngine.cs:367 still requires `info.HasAudioPosition`, so the vanished-track case never trips. A 45-minute run with no drop cannot show that either way.

src/Aetherphone/Core/Video/VideoEngine.cs
```diff
@@ -342,17 +342,20 @@ internal void ObserveForStall(in MpvPlaybackInfo info)
...
         {
             lastObservedAudioPosition = info.AudioPositionSeconds;
             audioProgressAtTicks = now;
         }
-        else if (!audioPositionSeen
-                 || Math.Abs(info.AudioPositionSeconds - lastObservedAudioPosition) > StallPositionTolerance)
+        else if (info.HasAudioPosition)
```
XeldarAlz, line 350: Blocking: e41a0989 fixed this on dev and shipped in v1.0.2.0 (the gate drops `HasAudioPosition`, and a pause or seek no longer clears `audioPositionSeen`), and this branch conflicts with it. Please retest on 1.0.3.2: a drop should log `Playback audio (track ended) stalled`. Close this if it holds, otherwise rebase and say what dev misses.

XeldarAlz (changes requested):
CI build, tests and guards green; reviewed against current dev, not built locally or run in-game. dev already carries a working version of this fix, shipped in v1.0.2.0; two notes inline.

PR #161, opened 2026-09-06
feat(uno): house rules

Haydooners:
## What

Adds a optional "House Rules" to the game Uno. Stack +2's & +4's, Swap Hands on 7's, Rotate hands on 0's

## Why

Everybody loves Uno house rules! And thought it'd be a fun first thing to do. 

## How I tested it

- Ran `dotnet build Aetherphone.sln -c Release`, clean build with 0 errors.
- Tested logic in `UnoRulesTests.cs` using `dotnet test`, returns all with 1967 successes. 
- Tested the host / lobby menu to ensure Default and House Rules can be selected properly.

## Screenshots

before
after

## Backend coordination

Introduces optional House Rules parameter support `ruleSet` for online Uno Games.

**DTO Changes:**

-`UnoRoomStateDto`
(`src/Aetherphone/Core/Aethernet/Contracts/GamesDtos.cs`):
 - Added `RuleSet` (default rules == 0, house rules == 1)
 - Added `PendingDrawCount` (default 0)
   - Indicates how much the next player in the turn order must draw if theyre unable to stack a +2 or +4. 

**Request Parameters Sent:**

-`ActionStart`
 - Passes `Card = selectedRuleSetIndex` when room host starts the game
- `ActionPlay`
 - When a 7 Card is played under House Rules, passes "Color = targetSeatIndex` to indicate the hand swap target seat. 

**Current behavior against un-updated server:**

- If an older server does not return `RuleSet` or `PendingDrawCount`, It should default to the "Default" rule set and play as normal.
- Unable to totally finish/verify everything works until the backend can receive new rule parameters. 

## Checklist

Strings:

Quality:


src/Aetherphone/Core/Localization/L.cs
```diff
@@ -5181,6 +5181,11 @@ internal static class Games
         public static readonly LocString OnlineEightWin = new("games.onlineEightWin", "{0} sinks the eight and wins!");
         public static readonly LocString OnlineEightEarlyLoss = new("games.onlineEightEarlyLoss", "{0} wins, the eight went down too early");
         public static readonly LocString OnlineEightScratchLoss = new("games.onlineEightScratchLoss", "{0} wins, the eight fell on a foul");
+        public static readonly LocString OnlineRuleDefault = new("games.onlineRuleDefault", "Default");
+        public static readonly LocString OnlineRuleHouse = new("games.onlineRuleHouse", "House Rules");
+        public static readonly LocString OnlineHouseBadge = new("games.onlineHouseBadge", "⚡ House Rules");
```
XeldarAlz: Blocking: the emoji have to come out of these three strings. The phone's font atlas cannot draw Unicode emoji, which is why emoji are ASCII :shortcode: everywhere else and only render through RichText. Typography.Draw and ShowBanner do not take that path, so these ship as tofu boxes.

src/Aetherphone/Localization/de.json
```diff
@@ -3140,6 +3140,11 @@
   "games.onlineRoundVoid": "Die Runde endete, weil niemand mehr da war",
   "games.onlineDeck": "Stapel",
   "games.onlineUnoCall": "Uno!",
+  "games.onlineRuleDefault": "Default",
+  "games.onlineRuleHouse": "House Rules",
```
XeldarAlz: Blocking: the five new keys carry English text in all nine catalogs. LocalizationParityTests only checks that the keys exist, so this passes CI while breaking the lockstep rule. Each locale needs a real translation in this same PR.

src/Aetherphone/Apps/Games/Online/OnlineRoomView.cs
```diff
@@ -46,6 +49,11 @@ public void Enter()
         poolTable.Reset();
     }
 
+    private static readonly DropdownMenu.Item[] RuleSetOptions = [
```
XeldarAlz: Blocking: a static initializer resolves Loc.T once at type init, so these labels freeze in whatever language loaded first. Build the items at draw time the way HousingApp.cs:357 and HuntsApp.Filters.cs:206 do, and set Selected on the current rule set while you are in there.

src/Aetherphone/Apps/Games/Online/OnlineRoomView.cs
```diff
@@ -233,6 +244,26 @@ private void DrawLobby(Rect body, PhoneTheme theme, float scale, GameRoomState h
             DrawInlineNotice(theme, scale);
         }
 
+        if (isHost && held.Snapshot.GameKind == GameRoomWire.UnoKind) {
+            var pickOrigin = ImGui.GetCursorScreenPos();
+            var pickWidth = ScrollLayout.StableContentWidth();
+            var dropdownRect = new Rect(pickOrigin, new Vector2(pickOrigin.X + pickWidth, pickOrigin.Y + 36f * scale));
+
+            var label = $"Rules: {RuleSetOptions[selectedRuleSetIndex].Label}";
```
XeldarAlz: Blocking: "Rules: " is hardcoded English and needs a LocString. The interpolation also allocates a string every frame in draw code, so fold the prefix into the two rule set strings rather than building the label per frame.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -654,14 +672,20 @@ private void DrawSeats(ImDrawListPtr drawList, PhoneTheme theme, float scale, Un
...
-        for (var seat = 0; seat < players.Length; seat++)
+        for (var seat = 0; seat < players.Length && seat < seatCounts.Length; seat++)
         {
             if (seat == mySeat)
             {
                 continue;
             }
 
             var player = players[seat];
+            if (sevenPendingCard >= 0 && ImGui.IsItemClicked(ImGuiMouseButton.Left))
```
XeldarAlz: Blocking: DrawSeats submits no ImGui item per seat, it paints on the draw list, so IsItemClicked resolves to the body Dummy from the top of Draw. Any left click anywhere plays the seven at the first non self seat. This needs a real hit test against each seat anchor, plus something on screen telling the player to pick a target.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -1022,6 +1048,20 @@ private void DrawHand(ImDrawListPtr drawList, Rect body, PhoneTheme theme, float
...
+        else if (board.RuleSet == GameRoomWire.RuleSetHouse && GameRoomWire.IsSeven(picked.Card))
+        {
+            var playerCount = board.Players?.Length ?? 0;
+            if (playerCount > 2)
+            {
+                sevenPendingCard = picked.Card;
+            }
+            else
+            {
+                var mySeatIndex = SeatOf(board.Players ?? Array.Empty<UnoPlayerDto>(), store.AccountId);
```
XeldarAlz: Non-blocking: mySeatIndex == 0 ? 1 : 0 only names the right opponent when the two seats happen to be 0 and 1. Walk the players array for the other seat that is not Away instead. The server refuses a bad target, so this currently fails the play rather than swapping the wrong hand.

src/Aetherphone/Apps/Games/Online/OnlineHub.cs
```diff
@@ -45,7 +45,8 @@ public OnlineHub(GameRoomsStore store, Action<string, string> openRoom)
 
     public void Enter(string preferredKind)
     {
-        this.preferredKind = preferredKind;
+        AepLog.Debug($"[OnlineHub] Enter: preferredKind={preferredKind}");
```
XeldarAlz: Blocking: please drop the AepLog.Debug scaffolding before merge, here and at 120, 125, 133, 211, 248, 282 and 297, plus GameRoomsStore.cs:240 and OnlineUnoTable.cs:269. The OnlineHub ones have nothing to do with house rules, and the house style is one concern per PR.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -147,6 +148,7 @@ public void Draw(Rect body, PhoneTheme theme, float scale, GameRoomSnapshotDto s
     {
         using var surface = AppSurface.Begin(body, true);
         ImGui.Dummy(new Vector2(MathF.Max(1f, body.Width - 32f * scale), body.Height - 16f * scale));
+        var ruleSet = store.ActiveRuleSet != 0 ? store.ActiveRuleSet : board.RuleSet;
```
XeldarAlz: Non-blocking: ruleSet is computed and never read. It would also be wrong: ActiveRuleSet is set only on the host and never reset in Exit, so a stale 1 leaks into the next room. board.RuleSet is the only source that holds for every seat.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -879,7 +903,8 @@ private static bool AnyPlayable(int[] hand, UnoRoomStateDto board)
     {
         for (var index = 0; index < hand.Length; index++)
         {
-            if (GameRoomWire.IsPlayable(hand[index], board.ActiveColor, board.DiscardTop))
+            if (GameRoomWire.IsPlayable(hand[index], board.ActiveColor, board.DiscardTop,
```
XeldarAlz, line 1016: Non-blocking: the hand now greys correctly against a stack, but the deck tap still reads as a one card draw. On the server one draw eats the whole count, so the deck needs a draw N affordance or players will not understand the six cards that just landed.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -294,7 +302,7 @@ private void ObserveBoard(UnoRoomStateDto board, UnoPlayerDto[] players, int myS
             }
         }
 
-        for (var seat = 0; seat < players.Length; seat++)
+        for (var seat = 0; seat < players.Length && seat < previousCounts.Length; seat++)
```
XeldarAlz: Non-blocking: this guard and the one at 675 hide the fault instead of fixing it. previousCounts and seatCounts are allocated to players.Length in Layout, so if a seat index can outrun them then Layout is running after its reader. Worth finding that ordering bug.

src/Aetherphone.Tests/UnoRulesTests.cs
```diff
@@ -0,0 +1,184 @@
...
+    private const int Wild = 52;        
+    private const int WildDrawFour = 53;
+
+    [Fact]
+    public void HouseRulesAllowsStackingDrawTwoOnDrawTwo(){
+
+        int topCard = 12; // Red Draw 2
+        int myCard = 25; // Blue Draw 2
+
+        bool playableDefault = GameRoomWire.IsPlayable(
```
XeldarAlz: Nit: declaring bool and int where the repo uses var is what produces the 28 IDE0007 warnings. The file also has no trailing newline, and neither do the nine JSONs any more. HouseRulesAllowsStackingDrawTwoOnDrawTwo and PendingDraw_HouseRulesAllowsDrawTwoStacking assert the same thing twice.

XeldarAlz (changes requested):
Thanks for this, house rules are a good first thing to pick up and the wire shape you chose was the right one.

Release build: 0 errors, but 28 new IDE0007 warnings, all in UnoRulesTests.cs. dotnet test: 1969 passing. Not verified in game, because no server implements the three rules yet.

The server half now exists on a feat/uno-house-rules branch in the Aethernet repo, built to exactly the contract this PR assumes (rule set in the start action's card slot, target seat in the play action's color slot), 1935 tests green. Findings inline.

src/Aetherphone/Apps/Games/Online/OnlineHub.cs
Haydooners: Apologies forgot about these, thought I got rid of most of them. Usually shove them in places to see when the front end hits certain parts of code, Ill make sure to remove them in the future.

Haydooners:
made the requested changes, builds cleanly and all tests ran with all successes and usual warnings.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -913,6 +977,19 @@ private void DrawHand(ImDrawListPtr drawList, Rect body, PhoneTheme theme, float
         }
 
         var count = slotCount;
+        
+        if (count > 0)
+        {
+            var badgeCenter = Absolute(new Vector2(handAnchor.X + halfSpan + cardWidth * 0.5f + 16f * scale, handAnchor.Y - 12f * scale));
```
XeldarAlz: Blocking: this does not compile. `halfSpan` is declared at 997, below this use, so the Release build fails with CS0841 and nothing in the solution builds. dev builds clean, so the break is this commit.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -913,6 +977,19 @@ private void DrawHand(ImDrawListPtr drawList, Rect body, PhoneTheme theme, float
         }
 
         var count = slotCount;
```
XeldarAlz, line 1050: Blocking: a count badge on my own hand is not a house rule, it changes the base Uno table for every ruleset. Please pull it out into its own PR. `count > 0` is also dead, `slotCount == 0` already returned two lines up.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -741,6 +780,23 @@ private void DrawCenter(ImDrawListPtr drawList, PhoneTheme theme, float scale, U
         }
 
         UnoCardArt.DrawBack(drawList, topRect, scale, canDraw ? 1f : 0.8f);
+        if (board.PendingDrawCount > 1)
+        {
+            var badgeText = $"+{board.PendingDrawCount}";
```
XeldarAlz: Blocking: three new interpolations in the draw path, here plus the deck label at 823 and the no playable line at 932. They allocate every frame, which is the same note as the old "Rules: " one. Compose them once when `PendingDrawCount` changes and hold the result in a field.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -654,6 +671,16 @@ private void DrawSeats(ImDrawListPtr drawList, PhoneTheme theme, float scale, Un
             return;
         }
 
+        var pickingTarget = sevenPendingCard >= 0;
```
XeldarAlz, line 746: Blocking: `sevenPendingCard` is never cleared. Reset() clears `wildPendingCard` and not this one, and neither does the round change in ObserveBoard, so arming a seven and leaving the room carries the picker into the next round where any seat tap fires a stale play. `interactive` at 1001 also gates on `wildPendingCard` alone, so the hand and the deck stay live while the picker is armed, and there is no cancel the way the wild picker has one at 1237.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -663,6 +690,18 @@ private void DrawSeats(ImDrawListPtr drawList, PhoneTheme theme, float scale, Un
 
             var player = players[seat];
             var anchor = Absolute(seatAnchors[seat]);
+
+            if (pickingTarget)
+            {
+                var hitRadius = HitRadius * scale;
+                ProgressRing.Glow(anchor, hitRadius, accent, 0.3f + 0.25f * Pulse.Wave(Pulse.Calm));
```
XeldarAlz, line 767: Away seats are clickable targets here, and the server validates the target is still in the round, so tapping one loses the play with nothing on screen. The two player path below already filters `Away`, do the same here.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -376,6 +383,16 @@ private void ReactToCardEffect(UnoRoomStateDto board, UnoPlayerDto[] players, in
...
+        {
+            ShowBanner(Loc.T(L.Games.OnlineHandsRotated), accent);
+        }
+
+        if (board.RuleSet == GameRoomWire.RuleSetHouse && GameRoomWire.IsSeven(board.LastCard))
+        {
+            ShowBanner(Loc.T(L.Games.OnlineHandSwapped), accent);
+        }
+
         if (penalty == 0 || seat < 0 || players.Length < 2)
```
XeldarAlz: Under house rules the server banks the penalty into `PendingDrawCount` instead of serving it, so this flight deals two or four cards onto a seat that received nothing. Gate it on `RuleSetDefault`, and fly the stack when someone actually eats it.

src/Aetherphone/Apps/Games/Online/OnlineUnoTable.cs
```diff
@@ -1022,6 +1100,49 @@ private void DrawHand(ImDrawListPtr drawList, Rect body, PhoneTheme theme, float
...
+            {
+                sevenPendingCard = picked.Card;
+            }
+            else
+            {
+                var mySeatIndex = SeatOf(board.Players ?? Array.Empty<UnoPlayerDto>(), store.AccountId);
+                var opponentSeat = -1;
+                if (board.Players != null)
+                {
+                    for (var i = 0; i < board.Players.Length; i++)
```
XeldarAlz: `i` here and at 1128. No abbreviations in names, even loop variables: `seatIndex`.

src/Aetherphone/Core/Games/GameRoomWire.cs
```diff
@@ -119,8 +123,36 @@ public static bool IsWild(int card)
...
+    public static bool IsSeven(int card)
+    {
+        return card >= 0 && card < WildCard && RankOf(card) == 7;
+    }
+
+    public static bool IsPlayable(int card, int activeColor, int topCard,
+     int ruleSet = RuleSetDefault, int pendingDrawCount = 0)
     {
+
+        if (pendingDrawCount > 0){
```
XeldarAlz: The rule is right, the shape is not: braces belong on their own line, the body is indented three levels deep for one branch, there is a blank line after the opening brace and trailing whitespace at 148. Same for the five space continuation at 137. Invert the house rules check and return early.

src/Aetherphone/Core/Games/GameRoomsStore.cs
```diff
@@ -90,6 +90,8 @@ public GameRoomsStore(AethernetSession session, GamesClient games, PhoneVisibili
 
     public bool ActInFlight => actInFlight;
 
+    public int ActiveRuleSet { get; set; } = GameRoomWire.RuleSetDefault;
```
XeldarAlz: `ActiveRuleSet` is written and never read, drop it. `board.RuleSet` is the only source that holds for every seat, which is why the reader went away. `SendStart(int ruleSet = 0)` should also default to `GameRoomWire.RuleSetDefault`.

src/Aetherphone/Apps/Games/Online/OnlineRoomView.cs
```diff
@@ -221,6 +227,10 @@ private void DrawLobby(Rect body, PhoneTheme theme, float scale, GameRoomState h
         var roster = held.Roster!;
         var players = roster.Players;
         var isHost = IsHost(roster);
+        rulesMenu.Gate();
+
+        var picked = rulesMenuItems.Count > 0 ? rulesMenu.Draw(body, theme, System.Runtime.InteropServices.CollectionsMarshal
```
XeldarAlz: The guard misses `rulesMenu.Open`, the items are never cleared after a pick and the menu is never closed in Reset, so an open picker comes back the next time a lobby draws. HousingApp.cs:314 is the shape to copy. Please also put `CollectionsMarshal` in a using rather than a fully qualified name split across the ternary.

src/Aetherphone/Apps/Games/Online/OnlineRoomView.cs
```diff
@@ -261,7 +292,7 @@ private void DrawLobby(Rect body, PhoneTheme theme, float scale, GameRoomState h
             if (GameHud.Button(primaryCenter, buttonSize, label, enough ? accent : theme.TextMuted, theme)
                 && enough && !store.ActInFlight)
             {
-                store.SendStart();
+                store.SendStart(selectedRuleSetIndex);
```
XeldarAlz: `selectedRuleSetIndex` is sent for chess and pool too, and Reset never clears it, so hosting a house rules Uno room then hosting chess ships a 1 in the start action's card slot. Send `RuleSetDefault` unless the room is Uno, and reset the index with the rest of the state.

src/Aetherphone/Localization/zh.json
```diff
@@ -3140,6 +3140,12 @@
   "games.onlineRoundVoid": "桌上没有玩家了，对局结束",
   "games.onlineDeck": "牌堆",
   "games.onlineUnoCall": "Uno！",
+  "games.onlineRuleDefault": "Default",
```
XeldarAlz: Still English. 默认.

src/Aetherphone/Localization/ja.json
```diff
@@ -3140,6 +3140,12 @@
   "games.onlineRoundVoid": "誰もいなくなったのでラウンドが終了しました",
   "games.onlineDeck": "山札",
   "games.onlineUnoCall": "ウノ！",
+  "games.onlineRuleDefault": "標準",
+  "games.onlineRuleHouse": "ハウスルール",
+  "games.onlineHouseBadge": "ハウスルール",
+  "games.onlineHandsRotated": "手札を交代！",
```
XeldarAlz: 交代 is a change of shift, not a rotation. 手札が回りました or 手札をローテーション. ru.json:3146 has the same kind of slip, обменяны should be обменялись.

src/Aetherphone/Core/Localization/L.cs
```diff
@@ -5181,6 +5181,12 @@ internal static class Games
         public static readonly LocString OnlineEightWin = new("games.onlineEightWin", "{0} sinks the eight and wins!");
         public static readonly LocString OnlineEightEarlyLoss = new("games.onlineEightEarlyLoss", "{0} wins, the eight went down too early");
         public static readonly LocString OnlineEightScratchLoss = new("games.onlineEightScratchLoss", "{0} wins, the eight fell on a foul");
+        public static readonly LocString OnlineRuleDefault = new("games.onlineRuleDefault", "Default");
+        public static readonly LocString OnlineRuleHouse = new("games.onlineRuleHouse", "House Rules");
+        public static readonly LocString OnlineHouseBadge = new("games.onlineHouseBadge", "House Rules");
```
XeldarAlz: `OnlineHouseBadge` and `OnlineRuleHouse` carry identical text in all nine catalogs. One key is enough.

src/Aetherphone/Apps/Games/Online/OnlineHub.cs
```diff
@@ -45,7 +45,7 @@ public OnlineHub(GameRoomsStore store, Action<string, string> openRoom)
 
     public void Enter(string preferredKind)
     {
-        this.preferredKind = preferredKind;
+        this.preferredKind = preferredKind;
```
XeldarAlz: Please revert this file, the only change left in it is a trailing space.

src/Aetherphone/Apps/Games/Online/OnlineRoomView.cs
```diff
@@ -9,6 +9,7 @@
 using Dalamud.Bindings.ImGui;
 using Dalamud.Interface;
 
+
```
XeldarAlz: Nit: stray blank line. UnoRulesTests.cs also still has no trailing newline.

XeldarAlz (changes requested):
Second pass. Every thread from the first round is genuinely fixed: the emoji are gone, the nine catalogs carry real translations, the rule set items are built at draw time, the seat picker is a real hit test through UiInteract.HoverClickCircle, the debug logging is out, the seat buffers are checked and cleared together, and the deck now shows the stack it owes. Thanks for turning that around.

The head does not compile though. Release build on dev is clean, on 377a79f7 it fails with CS0841 in OnlineUnoTable.cs:983, so `dotnet test` never ran either and the checklist claims do not hold for this commit. The break comes from the last commit, the hand count badge, which is also not a house rule and belongs in its own PR.

Beyond that: `sevenPendingCard` is never cleared, three per frame interpolations landed in the draw path, the +2 penalty flight still deals cards the house ruleset banks instead, and zh.json kept English for the default rule set. Findings inline.

The backend branch is unchanged and still matches the contract this PR assumes.

src/Aetherphone/Localization/ja.json
XeldarAlz: Correction on the Russian half: обменяны is a fine short participle of обменять, so ru.json stands as you wrote it. Japanese is now 手札ローテーション！, pushed.

XeldarAlz:
Pushed the review fixes to this branch as three commits rather than leaving you a list, since the build break blocked everything else:

- `df6fa691` fix(uno): repair the table draw path and the seven target picker. Drops the hand count badge (it read `halfSpan` before it existed, and a count on your own fan is not a house rule, so it can land on its own PR), clears the picker with the round, the turn and the room, makes the hand and the deck inert while it is up, cancels it on a click that misses every seat, skips away seats as targets, auto targets whenever exactly one opponent is live, and stops flying penalty cards that house rules bank instead of serving. The stack now flies when a seat actually eats it, as many cards as it owed. Stacked draw labels are composed on change instead of interpolated per frame.
- `0887bad9` refactor(games): tidy the rule set seam. `IsPlayable` reads as early returns, `IsZero` and `IsSeven` lean on `RankOf` with named rank constants, `ActiveRuleSet` is gone, the lobby picker follows the shape the rest of the app uses and resets on room entry, and `OnlineHub` is back to untouched. Adds the test for the case the server enforces and the client had none for: a draw two cannot land on a wild draw four's stack.
- `5afbece1` fix(l10n): finish the house rules strings. The badge draws the rule set string and the duplicate key is gone from all nine catalogs, zh gets 默认, ja gets 手札ローテーション！.

Release build is clean with 0 warnings. The test suite cannot run on macOS (`net10.0-windows`), so I approved the CI run on this push for the Windows build and test job.

Two things left before this merges: the server half needs to go out from the Aethernet branch first, and it wants a real in game pass on all three rules with three or more seats, which nobody can do until then.

PR #167, opened 2026-09-14
feat(wallet): Add weekly cap for tomestones

Deldee:
## What
Added the weekly capped tomestone limit to the wallet display

(As a note, the wallet app currently does game read on the draw thread with a delta poll time, I have anothe pr in the work to fix this, this is pre-exisiting)

## Why
Currently wallet did not show the weekly cap

Closes #
https://discord.com/channels/1527213298087493732/1548905787860983808

## How I tested it
- Open the wallet app
- Check the weekly cap appears under tome of mnemonics
- If not already capped, obtained tomes
- Confirm the value updates

## Screenshots

image

## AI involvement

Copilot
No new asset

## Checklist

Strings:

Quality:


src/Aetherphone/Apps/Wallet/CurrencyRow.cs
```diff
@@ -95,7 +99,11 @@ public static void Draw(Rect band, Rect content, WalletEntry entry, ITextureProv
         var capped = entry.Amount >= entry.Cap;
         var accent = capped ? CappedTint : palette.Accent;
         var barTop = content.Max.Y - 15f * scale;
-        var lineCenterY = (content.Min.Y + barTop) * 0.5f;
+        var hasWeekly = entry.HasWeeklyCap;
+        var weeklyText = hasWeekly ? Loc.T(L.Wallet.WeeklyCap, Format(entry.WeeklyAmount), Format(entry.WeeklyCap)) : string.Empty;
```
XeldarAlz: Blocking: `Loc.T(entry, params object[])` allocates the args array plus a fresh string through `string.Format` every frame, and Draw has to stay allocation-free. Cache the text and rebuild it only when `WeeklyAmount`, `WeeklyCap` or `Loc.Current` change, the way `NumberText.Group` keys its cache.

src/Aetherphone/Apps/Wallet/WalletApp.cs
```diff
@@ -145,18 +145,25 @@ private Rect DrawSectionCard(WalletSection section, float scale)
...
         var origin = ImGui.GetCursorScreenPos();
-        var card = GroupCard.Begin(ui, rowCount, CurrencyRow.Height);
+        var totalHeight = 0f;
         for (var entryIndex = 0; entryIndex < rowCount; entryIndex++)
         {
-            var contentRect = card.NextRow();
+            totalHeight += CurrencyRow.HeightFor(section.Entries[entryIndex]);
+        }
+
+        var card = GroupCard.Begin(ui, (int)totalHeight, 1f);
```
XeldarAlz: Non-blocking: `rowHeight: 1f` turns `rowCount` and `rowSpan` into pixel counts, and the `(int)` casts truncate silently the moment a row height stops being whole. A float overload on `GroupCard` (total height in `Begin`, row height in `NextRow`) keeps the units honest.

src/Aetherphone/Core/Game/GameData.cs
```diff
@@ -440,40 +441,49 @@ public bool TryGetItem(uint itemId, out string name, out uint iconId, out int it
...
+
+        if (previous != 0)
+        {
+            into.Add(previous);
+        }
+
+        into.Add(PoeticsItemId);
+    }
+
+    public uint LimitedTomestoneItemId()
```
XeldarAlz: Non-blocking: "highest item id" is a proxy, and this walks the whole sheet again right after `CollectTomestoneItemIds` did. The sheet names the capped slot directly: Mnemonics (49) maps to `Tomestones` row 3 with `WeeklyLimit` 900, every other slot is 0. Pick on `Tomestones.Value.WeeklyLimit > 0` inside the existing scan.

src/Aetherphone/Core/Wallet/WalletReader.cs
```diff
@@ -216,6 +224,7 @@ private static long ReadAmount(InventoryManager* manager, WalletEntry entry)
         {
             CurrencyKind.Gil => (long)manager->GetGil(),
             CurrencyKind.Tomestone => (long)manager->GetTomestoneCount(entry.ItemId),
+            CurrencyKind.LimitedTomestone => (long)manager->GetTomestoneCount(entry.ItemId),
```
XeldarAlz: Nit: same body as the `Tomestone` arm above; `CurrencyKind.Tomestone or CurrencyKind.LimitedTomestone => ...` keeps a single read path.

src/Aetherphone/Apps/Wallet/CurrencyRow.cs
```diff
@@ -12,14 +12,18 @@ namespace Aetherphone.Apps.Wallet;
 internal static class CurrencyRow
 {
     public const float Height = 62f;
+    public const float WeeklyRowExtra = 18f;
```
XeldarAlz: Nit: with `HeightFor` as the entry point, nothing outside `CurrencyRow` reads `Height` or `WeeklyRowExtra` any more; both can be private.

XeldarAlz (changes requested):
Release build clean (0 warnings) and 2281/2281 tests pass on cf3b4fa; not run in-game. One blocker (per-frame allocation in the row draw), the rest are small. Details inline.

XeldarAlz (approved):
Thanks, all five points are addressed cleanly: the weekly-cap text is cached off the draw path, GroupCard has honest float units, and the limited tome comes straight from Tomestones.WeeklyLimit.

Non-blocking follow-up if you feel like it: ClearAmounts resets WeeklyAmount but not WeeklyCap, so after logout the row would read "This week: 0 / 900" until the next refresh.

PR #168, opened 2026-09-14
refactor(wallet): read wallet data on a framework tick instead of in Draw

Deldee:
## What
Move the logic outside of the draw thread ,using an FrameworkTicker instead, as requested of  docs/game-integration.md, 
Draw only reads the cached data now

## Why
Aligning the Wallet app to our current conventions

## How I tested it
- Open the wallet app
- Check a currency
- Spend a currency and check that the tracking is valid 
- 

## AI involvement

Copilot/No asset

## Checklist

Quality:


XeldarAlz (commented):
Verified on a0accf3: Release build clean (0 warnings) and `dotnet test` 2281/2281 passed, CI green; not run in-game. Two non-blocking notes inline.

src/Aetherphone/Apps/Wallet/WalletApp.cs
```diff
@@ -40,28 +41,36 @@ public int BadgeCount
...
         this.gameData = gameData;
         this.textures = textures;
+        this.framework = framework;
     }
 
-    public void OnOpened() => Rebuild();
+    public void OnOpened()
+    {
+        OnTick();
+        ticker = new FrameworkTicker(framework, RefreshIntervalMilliseconds, OnTick);
```
XeldarAlz, line 62: Non-blocking: `OnOpened` is not idempotent. `NavigationStack.OpenApp` can re-fire it without `OnClosed` (app already current, or reversed mid-dismiss: NavigationStack.cs:99-109), and this assignment then orphans the old ticker on `Framework.Update`. Latent for wallet today; `ticker ??=` makes it safe.

src/Aetherphone/Apps/Wallet/WalletApp.cs
```diff
@@ -13,7 +14,7 @@ namespace Aetherphone.Apps.Wallet;
 
 internal sealed class WalletApp : IPhoneApp
 {
-    private const float RefreshIntervalSeconds = 1.5f;
+    private const long RefreshIntervalMilliseconds = 1500;
     private const float SectionGap = 12f;
     private const float BadgeRefreshMillis = 1500f;
```
XeldarAlz, line 19: Non-blocking: the badge still reads game memory on the draw path. `HomeTileView.cs:34` checks `BadgeCount` before `IsBadgeVisible`, so `CountCapped` (L33) runs even with the badge off. One ctor-owned ticker (`InventoryCaptureService` shape) could own `cappedBadge` and refresh amounts only while open, retiring this constant and the per-open ticker.

XeldarAlz (approved):
Approved. The two inline notes can land as a follow-up.

src/Aetherphone/Apps/Wallet/WalletApp.cs
XeldarAlz: Fixed on dev in fe8c030b, no follow-up needed for this one.

mauriciospark:
Thanks for the feedback! 

- The `OnOpened` ticker idempotency issue was already fixed in commit `fe8c030`.
- For the badge, I plan to follow the `InventoryCaptureService` pattern to avoid memory reads on the draw-path:

```csharp
private readonly FrameworkTicker badgeTicker;
private int cachedBadgeCount;

// In the constructor:
badgeTicker = new FrameworkTicker(framework, 1500, () => 
{
    if (IsBadgeVisible)
    {
        cachedBadgeCount = walletService.GetCappedBadgeCount();
    }
});

PR #170, opened 2026-09-16
feat(skywatcher): Add ability to view other areas and favorite them

Deldee:
## What
- Allow the ability to view the weather of other areas (By default current weather is previewed, detail page is the same as the preexisting one/6 windows)
- Allow the ability to favorite region to always see their forecast (5 window)
- Optimize refresh rate to only hit on weather change (Predictably from EZT clock), zone change, EnvManager mismatch, or login for both widget and app
- Small backend changes to weather functions to allow hunt to use it more effectively (Also so that getting weather data is easier in the future, for fishing, foray and hunts)
- Current area is still pulled via the env manager instead of computed, with all the behavior that implies (control changing the widget and data, handling story weather, etc), those are the previous behaviors and I did not touch them

## Why

The current weather app only showed the current area, which was quite restrictive

Closes #

## How I tested it
- Open the home screen
- check that the weather is valid in the widget
- Change area
- Check that the widget has updated
- Open the app
- View another area
- Wait for a weather change
- Confirm that widget, current area and the viewed area all updated

## Screenshots
image

## AI involvement

Copilot, code reviewed and optimized by both AI and manually

## Checklist

Strings:

Quality:


src/Aetherphone/Apps/Skywatcher/SkywatcherApp.Browse.cs
```diff
@@ -0,0 +1,485 @@
...
+        for (var regionIndex = 0; regionIndex < regions.Count; regionIndex++)
+        {
+            var region = regions[regionIndex];
+            var visibleZones = CountOtherZones(region.Zones);
+            if (visibleZones == 0)
+            {
+                continue;
+            }
+
+            SectionLabel(region.Region, palette, scale);
```
XeldarAlz: Blocking: SectionLabel calls TextInfo.ToUpper every call, so each region header allocates a new string every frame (15+ per frame here). Store the uppercased region name in WeatherRegionGroup when ZonesByRegion builds it.

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.Browse.cs
```diff
@@ -0,0 +1,485 @@
...
+        var glyphBottom = labelTop - 4f * scale;
+        for (var index = 0; index < count; index++)
+        {
+            var window = windows[index];
+            var columnCenterX = inner.Min.X + columnWidth * (index + 0.5f);
+            var radius = (index == 0 ? currentGlyphRadius : PreviewGlyphRadius) * scale;
+            var glyphCenter = new Vector2(columnCenterX, glyphBottom - radius);
+            DrawMini(window, glyphCenter, radius);
+            var columnMaxWidth = MathF.Max(1f, columnWidth - 4f * scale);
+            Marquee.DrawCentered(new MarqueeId(marqueePrefix, marqueeBase + index), ShortWhen(window), columnCenterX,
```
XeldarAlz: Blocking: ShortWhen formats a new string per column per frame, 5 for the preview plus 5 per favorite card. Cache the labels on refresh (they only change once a minute).

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.cs
```diff
@@ -22,50 +22,87 @@ internal sealed partial class SkywatcherApp : IPhoneApp
...
     private readonly WeatherService weather;
     private readonly WeatherControl control;
+    private readonly Configuration configuration;
     private readonly List<WeatherWindow> forecast = new();
     private string zone = string.Empty;
-    private float sinceRefresh;
+    private long lastWindowStartUnix = -1;
     private SkywatcherTab activeTab;
     private bool scrubbing;
+    private bool showingBrowse = true;
```
XeldarAlz: Blocking: sub-screens go through ViewRouter (see ChirperApp) for the slide transition and a single back path, not a bool swap. Each layer can paint its own WeatherSky so the router layers stay opaque.

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.cs
```diff
@@ -106,6 +157,20 @@ public void Draw(in PhoneContext context)
         DrawBottomNav(navRect, palette, scale);
     }
 
+    private void DrawDetailBackChevron(Rect content, Vector4 ink, float scale)
```
XeldarAlz: Blocking: this is a line-for-line copy of SceneChrome.BackChevron with only the callback changed. Goes away with ViewRouter; otherwise give BackChevron an Action overload instead of duplicating it.

src/Aetherphone/Apps/Skywatcher/WeatherCard.cs
```diff
@@ -7,6 +7,7 @@ namespace Aetherphone.Apps.Skywatcher;
 internal static class WeatherCard
 {
     private const int ShadowLayers = 3;
+    private static readonly Vector4 ChipBorder = new(1f, 1f, 1f, 0.45f);
```
XeldarAlz: Blocking (scope): 0.14 to 0.45 restyles every weather chip app-wide (hourly strip, forecast list, Control grid). Please split this, the Hunts refactor and the MapGlyphs move out, and retitle to a lowercase summary with one concern.

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.Browse.cs
```diff
@@ -0,0 +1,485 @@
...
+    {
+        var origin = ImGui.GetCursorScreenPos();
+        var width = ImGui.GetContentRegionAvail().X;
+        var rowHeight = ZoneRowHeight * scale;
+        var card = new Rect(origin, origin + new Vector2(width, visibleCount * rowHeight + 10f * scale));
+        DrawGlass(card, palette, scale);
+        var inner = card.Inset(5f * scale);
+        var drawList = ImGui.GetWindowDrawList();
+        var rowIndex = 0;
+        for (var index = 0; index < zones.Count; index++)
```
XeldarAlz: Non-blocking: every zone row (marquee, gradient chip, star) draws every frame with no culling. Skip rows whose rect is outside the clip rect (ImGui.IsRectVisible).

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.Browse.cs
```diff
@@ -0,0 +1,485 @@
...
+        {
+            zoneName = string.Empty;
+            return false;
+        }
+
+        zoneName = weather.ZoneName(territoryId);
+        return zoneName.Length > 0;
+    }
+
+    private static float FavoritesSectionHeightFor(int visibleCount, float scale)
```
XeldarAlz: Non-blocking: this hardcodes SectionLabel's 12f/6f spacing and ItemSpacing, so it drifts silently if the label layout changes. Measure the real section height from the cursor on the previous frame instead.

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.Browse.cs
```diff
@@ -0,0 +1,485 @@
...
+            FavoriteForecast(territoryId);
+        }
+
+        configuration.Save();
+
+        var visibleAfter = VisibleFavoriteCount();
+        var delta = FavoritesSectionHeightFor(visibleAfter, scale) - FavoritesSectionHeightFor(visibleBefore, scale);
+        if (delta != 0f)
+        {
+            ImGui.SetScrollY(ImGui.GetScrollY() + delta);
```
XeldarAlz: Non-blocking: raw SetScrollY bypasses DragScrollHost, whose scroller can overwrite it mid-fling. Use surface.JumpTo(...) like pendingScrollReset does.

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.Browse.cs
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    private List<WeatherWindow> FavoriteForecast(uint territoryId)
+    {
+        if (favoriteForecasts.TryGetValue(territoryId, out var cached))
+        {
+            return cached;
+        }
+
+        var windows = new List<WeatherWindow>();
```
XeldarAlz: Non-blocking: RefreshFavoriteForecasts clears the dictionary and this allocates a fresh List per favorite on every refresh. Keep and reuse the lists, Forecast already clears its target.

src/Aetherphone/Core/Game/GameSheetLanguage.cs
```diff
@@ -0,0 +1,16 @@
...
+
+internal static class GameSheetLanguage
+{
+    public static ClientLanguage Current() => Loc.Current.Code switch
+    {
+        "de" => ClientLanguage.German,
+        "en" => ClientLanguage.English,
+        "fr" => ClientLanguage.French,
+        "ja" => ClientLanguage.Japanese,
+        _ => ClientLanguage.English,
```
XeldarAlz: Non-blocking: the CN client ships no English sheets, so zh maps here to an UnsupportedLanguageException in ZoneName/ZonesByRegion (HuntZoneCatalogue.cs:107 guards the same case). Fall back to the client language instead.

src/Aetherphone/Core/Game/WeatherService.cs
```diff
@@ -9,43 +10,164 @@ namespace Aetherphone.Core.Game;
...
+            if (zoneName.Length == 0)
+            {
+                continue;
+            }
+
+            var region = RegionName(territory.RowId);
+            if (region.Length == 0)
+            {
+                region = FieldOperations.IsFieldOperationZone(territory.TerritoryIntendedUse.RowId)
+                    ? Loc.T(L.Skywatcher.FieldOpsRegion)
```
XeldarAlz: Non-blocking: Core now depends on Skywatcher UI strings. Return an empty region and let the app substitute the Field Ops / Miscellaneous labels.

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.Browse.cs
```diff
@@ -0,0 +1,485 @@
...
+        var stored = configuration.SkywatcherFavorites;
+        for (var index = 0; index < stored.Count; index++)
+        {
+            favorites.Add(stored[index]);
+        }
+
+        favoritesSynced = true;
+    }
+
+    private void EnsureFavoritesSynced()
```
XeldarAlz: Nit: OnOpened already calls SyncFavorites, and the HashSet mirrors the config list. The sync flag and the set can both go.

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.Browse.cs
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    private void DrawBrowse(in SkyPalette palette, float scale)
+    {
+        EnsureFavoritesSynced();
+        DrawCurrentAreaPreview(palette, scale);
+        DrawFavoritesSection(palette, scale);
+        DrawSearchField(palette, scale);
+
+        var regions = RefreshedFilteredRegions();
```
XeldarAlz: Nit: a search with no matches renders an empty list. Show an empty-state line.

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.Browse.cs
```diff
@@ -0,0 +1,485 @@
...
+            }
+        }
+
+        return count;
+    }
+
+    private static void DrawFavoriteStar(ImDrawListPtr drawList, Vector2 center, float radius, in SkyPalette palette,
+        bool filled)
+    {
+        MapGlyphs.Star(drawList, center, radius, filled, MapGlyphs.FavoriteFill, palette.InkFaint, UiScale.Current);
```
XeldarAlz: Nit: Skywatcher now depends on Apps.Maps. Move Star and FavoriteFill into Windows/Components.

XeldarAlz (changes requested):
CI green (build + guards); not run in-game on my side. Three blocking items: per-frame allocations, ViewRouter, and PR scope/title. Details inline.

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.cs
```diff
@@ -95,7 +153,21 @@ public void Draw(in PhoneContext context)
...
+                    pendingScrollReset = false;
+                }
+
+                if (activeTab == SkywatcherTab.Control)
+                {
+                    DrawControl(palette, scale);
+                }
+                else
+                {
+                    router.Draw(body, AppSkin.Transparent, ImGui.GetIO().DeltaTime, drawView);
```
XeldarAlz: Blocking: router.Draw runs inside the ##sky scroll child, but each router layer is a fixed body-sized child with NoScrollbar | NoScrollWithMouse. ##sky has nothing to scroll, so Browse and Detail clip at the nav bar. Call router.Draw at top level and open the scroll surface inside DrawView, like ChirperApp.

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.cs
```diff
@@ -22,69 +22,127 @@ internal sealed partial class SkywatcherApp : IPhoneApp
...
-        if (sinceRefresh >= RefreshIntervalSeconds)
+        var windowStart = WeatherService.CurrentWindowStartUnix();
+        var route = router.Current;
+        var onBrowse = route.Screen == SkywatcherScreen.Browse;
+        var activeForecast = onBrowse ? previewForecast : detailForecast;
+        var activeTerritoryId = onBrowse ? weather.CurrentTerritoryId : route.TerritoryId;
+        var zoneChanged = onBrowse && weather.CurrentTerritoryId != lastPreviewTerritoryId;
+        var liveDiverged = activeTerritoryId == weather.CurrentTerritoryId && activeForecast.Count > 0 &&
+            weather.LiveRenderedWeather() is { } live && live.Id != activeForecast[0].Weather.Id;
+        if (zoneChanged || liveDiverged || windowStart != lastWindowStartUnix)
```
XeldarAlz: Blocking: refresh now only fires per weather window (1400s), and MinutesFromNow is computed once in Forecast. The "in Nm" strips, hourly row, forecast list and Summary freeze for up to 23 minutes. Keep the window refresh, but rebuild the minute labels when the minute changes.

src/Aetherphone/Apps/Skywatcher/SkywatcherWidget.cs
```diff
@@ -71,17 +71,22 @@ public void Draw(in WidgetContext context)
...
+    private void Advance()
     {
-        sinceRefresh += delta;
-        if (sinceRefresh < RefreshIntervalSeconds)
+        var windowStart = WeatherService.CurrentWindowStartUnix();
+        var territoryId = weather.CurrentTerritoryId;
+        var liveDiverged = territoryId == viewedTerritoryId && forecast.Count > 0 &&
+            weather.LiveRenderedWeather() is { } live && live.Id != forecast[0].Weather.Id;
+
+        if (!liveDiverged && territoryId == viewedTerritoryId && windowStart == lastWindowStartUnix)
```
XeldarAlz: Blocking: same freeze on the home widget. When() reads MinutesFromNow from a forecast that is now only rebuilt per window, so the countdown sits still for up to 23 minutes (dev refreshed every 5s).

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.cs
```diff
@@ -22,69 +22,127 @@ internal sealed partial class SkywatcherApp : IPhoneApp
...
-        var kind = hasData ? WeatherSky.Classify(forecast[0].Weather.EnglishKey) : WeatherKind.Clouds;
-        var palette = WeatherSky.Blend(kind, hasData ? daylight : 0f);
+        var (kind, isDay, palette, _) = SkyDataFor(activeForecast);
         WeatherSky.Paint(screen, theme.ScreenRounding * scale, palette, kind, isDay);
         WeatherAmbience.Draw(ImGui.GetWindowDrawList(), screen, theme.ScreenRounding * scale, kind, isDay, palette,
             scale, 1f, false);
-        SceneChrome.BackChevron(content, context.Navigation, palette.Ink, scale);
+        if (activeTab == SkywatcherTab.Forecast && !onBrowse)
+        {
+            SceneChrome.BackChevron(content, CloseDetail, palette.Ink, scale);
```
XeldarAlz: Blocking: CloseDetail is an instance method group, so this allocates a new Action every frame. Cache it in a readonly field in the constructor, like ChirperApp's back.

src/Aetherphone/Windows/Components/Chrome/SceneChrome.cs
```diff
@@ -22,7 +22,10 @@ public static Rect AppAreaFrom(Rect content, PhoneTheme theme, float scale)
         return new Rect(new Vector2(screen.Min.X, content.Min.Y), new Vector2(screen.Max.X, content.Max.Y));
     }
 
-    public static void BackChevron(Rect content, INavigator navigation, Vector4 ink, float scale)
+    public static void BackChevron(Rect content, INavigator navigation, Vector4 ink, float scale) =>
+        BackChevron(content, navigation.Back, ink, scale);
```
XeldarAlz: Blocking: navigation.Back is an interface instance method group, so this overload allocates a delegate on every call, every frame. Keep the INavigator overload calling navigation.Back() directly and share the body through a private helper instead.

src/Aetherphone/Apps/Skywatcher/SkywatcherApp.Browse.cs
```diff
@@ -0,0 +1,492 @@
...
+                favoriteZones.Add(new WeatherZoneEntry(territoryId, zoneName));
+            }
+        }
+
+        if (favoriteZones.Count == 0)
+        {
+            return;
+        }
+
+        SectionLabel(Loc.T(L.Skywatcher.Favorites), palette, scale);
```
XeldarAlz: Non-blocking: SectionLabel still runs ToUpper on this label every frame while favorites are shown. Cache the uppercased string per locale, like RegionUpper.

src/Aetherphone/Core/Hunts/HuntSpawnConditionResolver.cs
```diff
@@ -9,17 +9,11 @@ internal readonly record struct HuntConditionWindow(DateTimeOffset Start, DateTi
 
 internal static class HuntSpawnConditionResolver
 {
-    private const double EorzeaSecondsPerRealSecond = 144.0 / 7.0;
-    private const long EorzeaSecondsPerHour = 3600;
-    private const long EorzeaSecondsPerDay = 86400;
-
-    private const long MoonPhaseEorzeaSeconds = 4 * EorzeaSecondsPerDay;
+    private const long MoonPhaseEorzeaSeconds = 4 * EorzeaTime.SecondsPerDay;
```
XeldarAlz: Blocking (scope): the Hunts refactor, the MapGlyphs/MapsApp move, PopoutPresence and the EorzeaTime constants are still in this PR. Please split them into their own PR so this one stays the Skywatcher feature.

XeldarAlz (changes requested):
CI green (build + guards); not run in-game on my side. Previous blockers mostly addressed; new blockers inline. The earlier non-blocking comments are still open. Title still needs a lowercase summary.

Deldee:
There were a lot of pushes due to rebasing and moving stuff out of this branch, I think this should now be accurate and limited to skywatcher (I did include the map star glyph component change, since it's a direct result of adding this featurE)

PR #171, opened 2026-09-17
fix(hunts): use maintenance/last know time instead of epoch when startedAt is undefined

Deldee:
## What
If faloop didnt return startedAt, the default would be to use the epoch time, showing windows that were six digits days old

## Why

Closes #

## AI involvement
Assist, no asset

## Checklist

Strings:

Quality:


XeldarAlz (approved):
Thanks, this is the right fix, and reusing `TryResolveSnipedFallback` keeps the poll and realtime paths consistent. I retitled the PR to match your commit headline, since it anchors to the maintenance restart rather than `prevStartedAt`.

Non-blocking follow-up: if `maintenanceStartAt` has not loaded yet, these windows still fall back to the epoch. Having `HuntWindowMath` treat a window with no start time as `Unknown` would close that gap.

PR #172, opened 2026-09-18
feat(core): Add option to read game data in the phone's locale

Deldee:
## What
Currently, the game data reads always default to the client's language, and some API default to EN (ffxivcollect)
Added an option to set those primarily to the phone's locale, and if the locale doesnt exist, to the client language
This also adds checks on applications that cache game data to force a recache if the language changes

## Why
Some data would not match the phone's locale, no matter what, this aligns the locales as the first source of language turth, followed bythe game client as a fallback

Closes #

## How I tested it
- Set the phone to French
- Review the following apps : Market, Games (trivia), Collections, Hunts and ensure all game data is in French
- Swap the phone locale to Turkish
- Test the same apps, ensure the data is in the client's language (as turkish is not supported)

## AI involvement

Copilot, no new asset
## Checklist

Strings:

Quality:


src/Aetherphone/Core/Game/GameSheetLanguage.cs
```diff
@@ -0,0 +1,73 @@
...
+        if (Resolve(overrideLanguage) is not { } language)
+        {
+            return data.GetExcelSheet<T>();
+        }
+
+        try
+        {
+            return data.GetExcelSheet<T>(language);
+        }
+        catch (UnsupportedLanguageException)
```
XeldarAlz: Blocking: CN/KR clients ship only their own language. Phone de/fr/ja with the toggle on throws here, then the English fallback throws again uncaught into Draw callers (job, item, zone names). Fall back to `data.GetExcelSheet<T>()` instead.

src/Aetherphone/Core/Game/GameSheetLanguage.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
```
XeldarAlz: Blocking: even when this catch recovers, Lumina throws from the sheet constructor and nothing caches the failure, so it throws on every call, every frame. Probe each language once and cache the result.

src/Aetherphone/Core/Game/GameSheetLanguage.cs
```diff
@@ -0,0 +1,73 @@
...
+        var phoneLanguage = Loc.Current.Code switch
+        {
+            "de" => ClientLanguage.German,
+            "en" => ClientLanguage.English,
+            "fr" => ClientLanguage.French,
+            "ja" => ClientLanguage.Japanese,
+            _ => (ClientLanguage?)null,
+        };
+
+        return phoneLanguage ?? Plugin.DataManager.Language;
```
XeldarAlz: Nit: the client language is already the default, so return `null` here and keep the plain `GetExcelSheet<T>()` path. Also, the settings hint promises phone, then client, then English, but a failed phone language skips straight to English.

src/Aetherphone/Core/Game/GameSheetLanguage.cs
```diff
@@ -0,0 +1,73 @@
...
+    German,
+    French,
+    Japanese,
+}
+
+internal readonly record struct SheetLanguageGate(bool PreferPhoneLocale, string LocaleCode);
+
+internal static class GameSheetLanguage
+{
+    public static SheetLanguageGate CurrentGate() =>
```
XeldarAlz: Non-blocking: the gate keys on the raw locale even with the toggle off, so a phone language change clears Collections and rebuilds the full Market index for identical output. Gate on the resolved `ClientLanguage?` instead. Hunts can add the locale separately.

src/Aetherphone/Core/Game/GameSheetLanguage.cs
```diff
@@ -0,0 +1,73 @@
+using Aetherphone.Core.Localization;
+using Dalamud.Game;
+using Dalamud.Plugin.Services;
+using Lumina.Excel;
+using Lumina.Excel.Exceptions;
+
+namespace Aetherphone.Core.Game;
+
+internal enum SheetLanguageOverride
```
XeldarAlz: Non-blocking: only `English` is ever passed, so this enum and its switch are speculative. Take a plain `ClientLanguage` for the forced case, or keep the literal at the two call sites.

src/Aetherphone/Apps/Hunts/HuntsApp.Detail.cs
```diff
@@ -429,11 +437,19 @@ private string ResolveZoneLabel(string zoneId, uint territoryId)
         return label;
     }
 
-    private static string? ResolveLiveZoneName(uint territoryId) =>
-        territoryId != 0 && Plugin.DataManager.GetExcelSheet<TerritoryType>(HuntUiLanguage.SheetLanguage())
-            .TryGetRow(territoryId, out var territory) && territory.PlaceName.RowId != 0
-            ? territory.PlaceName.Value.Name.ExtractText()
+    private static string? ResolveLiveZoneName(uint territoryId)
```
XeldarAlz: Blocking: this used to always follow the phone language. With the toggle off by default, a French phone on an English client now gets an English zone header, while the rest of Hunts stays French. Keep Hunts on the phone language.

src/Aetherphone/Core/Market/MarketItemIndex.cs
```diff
@@ -156,11 +159,15 @@ private void Build()
             vendorPrices = localVendor;
             lowerNames = localLower;
             indexById = localIndex;
+            builtGate = GameSheetLanguage.CurrentGate();
```
XeldarAlz: Blocking: this stamps the gate current at build end, not the one the build ran with. A language change mid-build marks old-language names as up to date. Capture the gate in `EnsureBuilt`, pass it to `Build`, and stamp that.

src/Aetherphone/Core/Maps/MinimapReader.cs
```diff
@@ -160,7 +164,7 @@ private static unsafe uint CurrentAgentMapId()
     }
 
     private static string PlaceName(uint placeNameRowId) =>
-        placeNameRowId != 0 && Plugin.DataManager.GetExcelSheet<PlaceName>().TryGetRow(placeNameRowId, out var name)
+        placeNameRowId != 0 && Plugin.DataManager.GetLocalizedSheet<PlaceName>().TryGetRow(placeNameRowId, out var name)
```
XeldarAlz: Non-blocking: the minimap follows the toggle, but the Maps app (`MapData.PlaceName`, `CurrentLocation`), TravelPlanner, LocationShare and `GameData.TerritoryName` do not. With it on, the same zone shows in two languages. Migrate them or narrow the hint.

src/Aetherphone/Core/Collections/CollectionsCatalogService.cs
```diff
@@ -310,7 +333,7 @@ private LocalUnlocks CollectMounts()
     {
         var ids = new HashSet<int>();
         var total = 0;
-        foreach (var row in dataManager.GetExcelSheet<Mount>())
+        foreach (var row in dataManager.GetLocalizedSheet<Mount>())
```
XeldarAlz: Non-blocking: these seven unlock walks only filter and count by id, so nothing visible changes. Each one loads a second language copy of a large sheet into memory, though. Please revert them to `GetExcelSheet<T>()`.

src/Aetherphone/Core/Hunts/HuntZoneCatalogue.cs
```diff
@@ -101,10 +102,11 @@ private void OnLoaded(Dictionary<string, HuntZoneDefinition> parsed)
     private static Dictionary<string, uint> BuildTerritoryIdLookup()
     {
         var lookup = new Dictionary<string, uint>(StringComparer.OrdinalIgnoreCase);
+        var englishLanguage = GameSheetLanguage.Resolve(SheetLanguageOverride.English) ?? ClientLanguage.English;
```
XeldarAlz: Nit: this always evaluates to `ClientLanguage.English`, so keep the literal. Separately, the `territory.PlaceName.Value` read below hits the same trap this PR fixes in Hunts and Weather. It likely is not English on non-English clients.

src/Aetherphone/Apps/Settings/Pages/LanguagePage.cs
```diff
@@ -46,6 +46,18 @@ public void Draw(in PhoneContext context, Rect body)
             }
 
             card.End();
+
+            var gameDataCard = GroupCard.Begin(theme, 1);
```
XeldarAlz: Nit: this card sits flush under the language list with no header, while the translation block below uses `SettingsSection.Header`. Give it a header or a `SettingsSection.Hint` so it reads as its own setting.

XeldarAlz (changes requested):
CI green (Build + Guards), merges clean, locale lockstep complete. Not run in-game. Four blocking items inline, the rest are non-blocking or nits.

Deldee:
Note, I've reverted the  hunt UI changes, but they will be reintroduced in my next Pr for hunt optimization, since those changes will rely on reading game sheets data

PR #178, opened 2026-09-21
Fix(searchfield): emoji search field x button wasn't clearing the search field

Haydooners:
…aring the search.

## What & Why

Title explains, search field wasnt clearing in the emoji UI when clicking the X button.

## How I tested it

Went into every app I could think of with a chat function and tested the X button.

## Checklist

Quality:


src/Aetherphone/Windows/Components/Fields/SearchField.cs
```diff
@@ -90,8 +90,9 @@ public static void Draw(Rect bar, string imguiId, string hint, ref string text,
             return;
         }
 
-        var hovered = UiInteract.Hover(clearCenter - new Vector2(clearRadius, clearRadius),
-            clearCenter + new Vector2(clearRadius, clearRadius));
+        var clearMin = clearCenter - new Vector2(clearRadius, clearRadius);
+        var clearMax = clearCenter + new Vector2(clearRadius, clearRadius);
+        var hovered = !UiInteract.InputBlocked && UiInteract.HoverWindowOnly(clearMin, clearMax);
```
XeldarAlz: This fixes the emoji picker, but `SearchField` has 32 callers (Messages, Aethergram, Market, Spotlight, GameChatThread, Collections, ...) and `HoverWindowOnly` skips the overlay reservation for all of them. Anywhere a popup or menu reserves an overlay over a search bar's clear button (chat attach menu, mention popup, comment attachment panel, Collections source dropdown), the X now highlights and clears through it.

`HoverWindowOnly` is meant for content that is itself the overlay, which is why the picker's own cells use it. Could `EmojiPicker` opt in instead? For example, an optional `bool overlayOwned = false` on the `PhoneTheme` overload and the core `Draw`, using `!UiInteract.InputBlocked && UiInteract.HoverWindowOnly(...)` only when it's true, and keeping `UiInteract.Hover` as the default.

src/Aetherphone/Windows/Components/Fields/SearchField.cs
```diff
@@ -103,9 +104,9 @@ public static void Draw(Rect bar, string imguiId, string hint, ref string text,
         {
             return;
         }
-
+
```
XeldarAlz: Nit: this blank line picked up trailing spaces.

src/Aetherphone/Windows/Components/Fields/SearchField.cs
```diff
@@ -103,9 +104,9 @@ public static void Draw(Rect bar, string imguiId, string hint, ref string text,
         {
             return;
         }
-
+        
         ImGui.SetMouseCursor(ImGuiMouseCursor.Hand);
-        if (ImGui.IsMouseClicked(ImGuiMouseButton.Left))
+        if (UiInteract.Click(clearMin, clearMax, hovered) || ImGui.IsMouseClicked(ImGuiMouseButton.Left))
```
XeldarAlz: Only one side of this `||` ever fires. `ImGui.IsMouseClicked` clears the text on the press frame, so `hasText` goes false and the button is gone before `UiInteract.Click` reaches its release frame. The `Click` call just leaves a stale tap claim behind.

I'd keep just `if (UiInteract.Click(clearMin, clearMax, hovered))`. It fires on release, plays the tap sound, and matches how the picker's cells and tone swatches work.

XeldarAlz (changes requested):
Thanks for tracking this down, the root cause is exactly right: `EmojiPicker` reserves its whole panel via `UiInteract.HoverOverlay(area)`, so `UiInteract.Hover` is always false inside it and the clear button can never be hovered.

Two changes before merge, details inline: scope the hover change to the emoji picker instead of every search field, and drop the redundant click check.

Also, could you retitle to lowercase conventional style? For example: `fix(emoji): let the picker search clear button work inside its overlay`. The commit headline is truncated too, but squash merge with the corrected title takes care of that.

src/Aetherphone/Windows/Components/Fields/SearchField.cs
Haydooners: Didn't consider that! Added your suggestion and seems to be working as intended.

XeldarAlz (approved):
Both requested changes are in: the hover change is scoped to the emoji picker through overlayOwned, and the clear button fires on the release-frame click only. I pushed the two whitespace fixes to your branch and retitled the PR. Thanks for the fix!

